Issue #325, opened 2015-01-08
Context Module Security Update


acouch:
Context module has a security update: https://www.drupal.org/node/2403351

Since this issue only affects users with Context UI privileges which is only given to the admin user by default. we will not rush to update context. We are also removing Context as the layout editor.

janette:
fixed

PR #355, opened 2015-02-14
325 context upgrade

acouch:
We should be able to reenable the context modules to verify there is not a regression


acouch:
QA site created: http://325_context_upgrade.dkan.qa.nuamsdev.com

acouch:
ref: #325

acouch:
QA site created: http://325_context_upgrade.dkan.qa.nuamsdev.com

PR #366, opened 2015-02-18
Remove drupal-org-dev.make

dafeder:


acouch:
QA site created: http://remote_org_dev_make.dkan.qa.nuamsdev.com

PR #378, opened 2015-02-24
Fixes dataset view for logged in user

acouch:
Adds "auth user" to dataset view.


acouch:
QA site created: http://dataset_view_logged_in.dkan.qa.nuamsdev.com

PR #402, opened 2015-03-16
QA Only

drkloc:
- [x] field_coverage map should work with older Leaflet
- [x] resource preview should work with leaflet 0.7.3


acouch:
QA site created: http://QA_dkan_401.dkan.qa.nuamsdev.com

PR #407, opened 2015-03-18
fixed translation function

takuya310:
Add translation function for 'About', 'Add Dataset', 'Groups'.


nuamsmachine:

nuamsmachine:

takuya310:
Add translation function, again.

dafeder:
Closing - these are on feature exports so will be overwritten every time feature is exported. Supporting good i18n is clearly still a big todo but we should do according to Drupal best practice.

PR #463, opened 2015-05-18
QA only

topicus:
Don't merge this. QA only.


acouch:
QA site created: http://without_panopoly_widget.dkan.qa.nuamsdev.com

PR #496, opened 2015-06-03
Update drupal-org.make

janette:
QA ONLY - do not pull


acouch:
QA site created: http://4-license-facet.dkan.qa.nuamsdev.com

Issue #505, opened 2015-06-17
Set 'contrib' as the default folder in .make files instead of setting each time


frankcarey:
make files now support setting a default subdir folder like so:

```
 Set the default subdirectory for projects.
 defaults[projects][subdir] = contrib
```
PR on it's way

drkloc:
thank you man. merged.

PR #506, opened 2015-06-17
Issue #505: Make files support setting a default subdir, so use it.

frankcarey:
* Added the default to drupal.org.make and removed any unecessary setting of subdir = contrib.


acouch:
QA site created: http://dev-505-set-default-subdir.dkan.qa.nuamsdev.com

PR #615, opened 2015-07-29
Search api upgrade

acouch:
### Acceptance test
 
- [x] DKAN installs correctly
- [ ] Previous versions of DKAN search works correctly after http://cgit.drupalcode.org/search_api_db/tree/search_api_db.install#n117 database update which moves table names with full text table types to a single _text table per server.


nucivicmachine:
QA site created: http://search_api_upgrade.dkan.qa.nuamsdev.com

PR #624, opened 2015-08-05
QA Viz entity charts field reorg

dafeder:
QA only. See NuCivic/visualization_entity_charts#13
NuCivic/internal#744


nucivicmachine:
QA site created: http://internal-774-two-resource-fields.dkan.qa.nuamsdev.com

Issue #654, opened 2015-08-21
Get Features out of revision for 1.9 release


acouch:
On 7.x:
```
 DKAN Sitewide Data Dashboard                         dkan_data_dashboard                                  Enabled   7.x-1.0
 DKAN Data Story                                      dkan_data_story                                      Enabled   7.x-1.0
 DKAN Data Story Storyteller Role                     dkan_data_story_storyteller_role                     Enabled
 DKAN Dataset                                         dkan_dataset                                         Enabled   7.x-1.x
 DKAN Dataset Content Types                           dkan_dataset_content_types                           Enabled   7.x-1.x
 DKAN Dataset Groups                                  dkan_dataset_groups                                  Enabled   7.x-1.x
 DKAN Datastore                                       dkan_datastore                                       Enabled   7.x-1.x
 DKAN Default Content                                 dkan_default_content                                 Enabled
 DKAN Sitewide                                        dkan_sitewide                                        Enabled   7.x-1.x
 DKAN Sitewide Demo Front                             dkan_sitewide_demo_front                             Enabled
 DKAN Sitewide Menu                                   dkan_sitewide_menu                                   Enabled
 DKAN Sitewide Panelizer                              dkan_sitewide_panelizer                              Enabled   7.x-1.0
 DKAN panels                                          dkan_sitewide_panels                                 Enabled
 DKAN Profile Page                                    dkan_sitewide_profile_page                           Enabled
 DKAN Sitewide Roles and Permissions                  dkan_sitewide_roles_perms                            Enabled                         Overridden
 DKAN Sitewide Search                                 dkan_sitewide_search_db                              Enabled
 DKAN Sitewide User                                   dkan_sitewide_user                                   Enabled
 Open Data Schema Map DKAN                            open_data_schema_map_dkan                            Enabled
 Panopoly Images                                      panopoly_images                                      Enabled   7.x-1.21
 Panopoly Widgets                                     panopoly_widgets                                     Enabled   7.x-1.25              Overridden
```

PR #655, opened 2015-08-21
Issue #654 Updates features modules

acouch:
## Acceptance test

- [ ] All features should be out of revision except for ``panopoly_widget``
- [ ] Dataset list should be removed from resource and page content type panel


acouch:
Running into this issue https://www.drupal.org/node/927566

PR #715, opened 2015-10-01
QA ONLY - Removes body field base from dkan_dataset_content_types fea…

drkloc:
…ture

NuCivic/dkan_dataset#164


nucivicmachine:
QA site created: http://removing_field_base_for_body_as_a_feature.dkan.qa.nuamsdev.com

Issue #557, opened 2015-07-10
"Social" block should be last in sidebar, "Data Extent" block should take its place (when enabled)


stefaniefgray:
"Social" block should be last in sidebar, "Data Extent" block should take its place (when enabled)

"Data Extent" is more relevant for users who'd like to use the dataset for GIS; no one uses Google+ and it's pretty easy to paste the URL into Facebook and Twitter without the widget.

Cross-referenced in HHS (https://github.com/NuCivic/healthdata/issues/311) but relevant to DKAN as a whole.

![wisconsinpage](https://cloud.githubusercontent.com/assets/9059428/8629018/82389662-2725-11e5-9607-3754fcf7104c.png)

cienvaras:
PR: https://github.com/NuCivic/dkan/pull/733

janette:
done

PR #733, opened 2015-10-11
Switch positions of 'Social' and 'Data Extent' blocks in Dataset view panel.

cienvaras:
ref #557 

## Acceptance Test

On new installation of DKAN, on a dataset page, 
- [x] "Social" block should be last in the sidebar
- [x] "Data extent" should be just under the group block


nucivicmachine:
QA site created: http://557-move-social-data-extent-blocks.dkan.qa.nuamsdev.com

PR #735, opened 2015-10-13
Try out travisCI build matrix options to see if we can speed up test …

frankcarey:
…times by running in parallel.


nucivicmachine:
QA site created: http://dev-697-parallel-builds.dkan.qa.nuamsdev.com

frankcarey:
We're going with CirlceCI instead of Travis

PR #736, opened 2015-10-13
Try only building 7.x-1.x to avoid building branches twice.

frankcarey:
This should result in any PR branch being run, but only that branch.


nucivicmachine:
QA site created: http://dev-697-only-build-master.dkan.qa.nuamsdev.com

frankcarey:
This will cause only PR branches to be built (which is what we want I think) . This should be ready for a merge.

PR #747, opened 2015-10-23
Use https for urls used in dataset view page

rhabbachi:
Pulled from USDA downstream. No particular ticket to reference.

Domain fixed:
* openstreetmap.org
* assets.okfn.org

Credits goes to @upieper.
### Acceptance:
- [ ] "License" block in dataset view page uses https for the image.
- [ ] "Data Extent" block in dataset view page uses https to pull the tile.


nucivicmachine:
QA site created: http://652-use-https.dkan.qa.nuamsdev.com

PR #756, opened 2015-10-29
Example showing the advantages of using a "locked" make file.

frankcarey:
Running this command in the `/dkan` directory will read from `drupal-org.make.base` (and any included make files) and then generate a `drupal.org.make` file with all the dependencies resolved and everything included (no more includes)

Benefits:
- All versions of projects are pegged to specific versions (exceptions are -dev or git versions)
- You can quickly upgrade all projects to their latest versions for any module you don't peg to a version in `drupal-org.make.base`
- You see the FULL list of dependencies in a single make file. You don't have to go looking in sub-make files for where a module came from.
- Resolving the dependencies and the includes ahead of time speeds up subsequent makes like when testing.
- you can rebuild the locked make file all at once, but selectively add only certain lines if it in git if for some reason you don't actually want to upgrade.
- The base make file can still be pegged to specific versions if you like for a few modules that need it, but they should be documented with comments as to WHY they're being held back.

### DIFF
--- a/drupal-org.make
+++ b/drupal-org.make
@@ -80,8 +80,7 @@ projects[feeds_flatstore_processor][download][url] = "https://github.com/NuCivic
 projects[feeds_flatstore_processor][download][branch] = "master"
 
 projects[schema][subdir] = "contrib"
-projects[schema][download][revision] = "08b02458694d186f8ab3bd0b24fbc738f9271108"
-projects[schema][download][type] = "git"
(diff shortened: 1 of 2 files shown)


nucivicmachine:
QA site created: http://dev-make-file-lock.dkan.qa.nuamsdev.com

kimwdavidson:
Certain elements of this work have been adopted in other issues.

PR #762, opened 2015-11-05
Revert using a 'locked' make file until it's more tested an better do…

frankcarey:
…cumented.


nucivicmachine:
QA site created: http://dev-697-fix-more-tests-rebase.dkan.qa.nuamsdev.com

PR #763, opened 2015-11-05
Completely delete the locked make file. Wasn't called by anything but…

frankcarey:
… may cause confusion.


nucivicmachine:
QA site created: http://dev-697-fix-more-tests-rebase.dkan.qa.nuamsdev.com

PR #764, opened 2015-11-05
Remove make lockfile for now, will re-add once process is finalized

dafeder:


nucivicmachine:
QA site created: http://remove-lockfile.dkan.qa.nuamsdev.com

PR #766, opened 2015-11-11
Enable proboci QA site builds.

dkinzer:
Enables QA site builds.

Ref: NuCivic/pluto#116.

### Acceptance
- [ ] PRs get a QA site on the probo servers on PR and on pushes to the PR branch.

### DIFF
--- a/.probo.yml
+++ b/.probo.yml
@@ -1,0 +1,7 @@
+steps:
+  - name: Site Setup
+    plugin: Drupal
+    makeFile: 'build-dkan.make'
+    profileName: 'dkan'
+    runInstall: true
+    clearCaches: true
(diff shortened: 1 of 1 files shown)


nucivicmachine:
QA site created: http://pluto-116-proboci-builds.dkan.qa.nuamsdev.com

dkinzer:
@tizzo thanks for helping me configure this and also for reconfiguring probo so the drupal makefile worked without any extra steps from us!

acouch:
For some reason this test is failing on clicking around the groups which is unrelated. 

screen shot 2015-11-25 at 10 06 37 am

There is a step from:

screen shot 2015-11-25 at 10 06 43 am

To click "Group" which is not part of the breadcrumb. OG is on a set tag so even on DKAN 7.x-1.x so not sure why that would change https://github.com/NuCivic/dkan_dataset/blob/7.x-1.x/dkan_dataset.make#L124

dkinzer:
Not sure what could be causing these errors since the probo configuration does not affect dkan functionality at all.

PR #767, opened 2015-11-11
Creates test for additional info

acouch:
This test should fail initially.


topicus:
@acouch Could you explain what did you mean with "Initially"? I should check this test passes, right?

PR #775, opened 2015-11-17
Fix problem with long text on field additional info

jacintocapote:
When you try to edit a dataset and the text inside field additional info is very long the styles for this field are broken.


jacintocapote:
Related "CIVIC-159"

PR #796, opened 2015-12-03
Fix upgrade error on dkan_sitewide_roles_perms

dafeder:
ref #796


nucivicmachine:
QA site created: http://sitewide-roles-update-error.dkan.qa.nuamsdev.com

PR #801, opened 2015-12-04
Reverting a commit that broke install

dafeder:


nucivicmachine:
QA site created: http://revert-menu-links-update.dkan.qa.nuamsdev.com

PR #815, opened 2015-12-10
Use nid instead of uuid for JSON API link.

dkinzer:
#### Description
The RDF link for "Other Access" block is already using nid instead of uuid.
nid works and there are cases in which the "uuid" we are allowed to use in the
uuid field does not work for the API.

Ref:  Nucivic/Healthdata#683

#### User Story
As a data consumer of the JSON API I want to retrieve datasets when I click on
the JSON link in a dataset. (this is broken in HHS because many of our a URL as
the uuid for the dataset.)

#### Acceptance Criteria
- [ ] Go to dataset page, in the "Other Access" block click the JSON link.  You
should see a JSON version of the dataset.


nucivicmachine:
QA site created: http://hhs-683-use-nid-for-json-api.dkan.qa.nuamsdev.com

dkinzer:
I think this is still valid.

dkinzer:
I believe this was fixed by a product change.

PR #829, opened 2015-12-18
Add the cancel account perm to authenticated user.

dkinzer:
Ref https://jira.govdelivery.com/browse/CIVIC-663

 ###Acceptance:###
- [ ] List of authenticated user permission should include "cancel account" permission.


dafeder:
I'm merging this but please be aware that the dkan_sitewide_roles_perms module is deprecated. The above ticket referred to permissions in the new dkan_permissions module, which will be included in the next release.

PR #840, opened 2016-01-06
Allow for multiple map features.

dkinzer:
Ref. nucivic/usda-nal#751.
Ref. NuCivic/leaflet_draw_widget#10
### Description

Users sometimes would like to enable adding multiple markers or features to
a leaflet map but doing so would break the extent block. This commit allows
users to enable adding multiple features to a map without breaking the extent
block.
### QA Steps:
- First enable adding multiple features to a map with NuCivic/leaflet_draw_widget#10
- And, set geo spatial field settings for the dataset content type to have unlimited cardinality.: See NuCivic/dkan_dataset#192
- [ ] Edit a dataset and include multiple features in the geo spatial field. After saving data extent block should still show up.


nucivicmachine:
QA site created: http://usda-751-allow-for-multiple-map-features.dkan.qa.nuamsdev.com

dkinzer:
This could still prove useful.

kimwdavidson:
Needs to be refactored to be pulled in. Adding to icebox.

PR #845, opened 2016-01-07
Upgrade colorizer module and stop using patches

dafeder:
Done in part because the bug came back - let's see if this makes a difference.


nucivicmachine:
QA site created: http://workflow-colorizer-upgrade.dkan.qa.nuamsdev.com

PR #880, opened 2016-01-25
Removing ref_field_sync patch that's causing problems

dafeder:


nucivicmachine:
QA site created: http://remove-ref-field-patch.dkan.qa.nuamsdev.com

PR #887, opened 2016-01-28
NuCivic#886 Move theme function from nuboot_radix theme into the facet_icon module

janette:
## Issue 
#886 

A new module was added to dkan_sitewide_search_db, facet_icons. The new module adds icons to the facet blocks. A theme function was added to the theme rather than the module, so sites not using the default theme - nuboot_radix - will break due to the theme function being in the theme rather than the module. This PR adds it to facet_icon. The other PR removes it from nuboot_radix.

## Required Joint PR on Nuboot Radix
https://github.com/NuCivic/nuboot_radix/pull/59

## QA
1. Navigate to the datasets page, confirm that the facet icons are working
2. Switch the default theme
3. Navigate back to the datasets page and confirm that the site does not throw an undefined function error.

[ ] Make sure to revert nuboot_radix to 7.x-1.x before merging.

PR #894, opened 2016-02-02
Move facet_icons module into dkan_sitewide, add update hook.

dafeder:
Check to see if facet icons enabled on both db and solr searches


dafeder:
This got rebased into #887, so may not be necessary to merge.

nucivicmachine:
QA site created: http://facet-icons-move.dkan.qa.nuamsdev.com

PR #896, opened 2016-02-03
Dev overrides make

frankcarey:
Needs cleanup - saving progress on a way to have dkan modules add a temporary overrides.make for developing against various branches of other dkan modules, instead of having to create a special branch in dkan that does all that.


nucivicmachine:
QA site created: http://dev-overrides-make.dkan.qa.nuamsdev.com

dkinzer:
This is pretty old, and we've done a ton of changes since then, so I'm closing (I'll keep the branch around for a while just in case).

If nothing happens for a month after closing feel free to delete the branch.

Issue #912, opened 2016-02-08
Groups and Admin pagers are broken


acouch:
## Description

Group and admin search pagers don't work.

## Steps to Reproduce

Add 11 datasets to a group or change the group pager on a demo site to "1".

Click "next" on ``/group/group-name``

See the following:
<img width="1186" alt="screen shot 2016-02-08 at 9 07 39 am" src="https://cloud.githubusercontent.com/assets/512243/12889250/a96d7458-ce43-11e5-802e-d4ec1d346b8b.png">

## Fix is to add the following code to dkan_sitewide or dkan_dataset

```php
function hook_settings_form_alter(&$form, $form_state, $form_id) {
  switch ($form_id) {
    case 'views_exposed_form':
      // Fixes pager bug.
      if (isset($form['sort_order']) && isset($form['sort_by']['#options']['changed'])) {
        $sort_by = 'changed';
        if(isset($_GET['sort_by'])){
          $sort_by = $_GET['sort_by'];
        }
        else {
          $_GET['sort_by'] = $sort_by;
        }
        $form_state['view']->set_exposed_input($_GET);
        $form_state['input']['sort_by'] = $sort_by;

        $form['sort_order']['#options']['ASC'] = t('Ascending');
        $form['sort_order']['#options']['DESC'] = t('Descending');
      }
      break;
  }
}
```

PR #913, opened 2016-02-08
Fix for #912 Groups and User pager bug

acouch:
ref: #912


nucivicmachine:
QA site created: http://912_groups_pager.dkan.qa.nuamsdev.com

PR #922, opened 2016-02-10
Allow pass extra params to ahoy tests proxied through circle_behat

topicus:
In order to override some behat params we need to allow circle_behat.sh to 
receive behat parameter and pass them to the ahoy test function. In this way
each project can specify things like tags to exclude or include.


frankcarey:
Looks Good - Merging

PR #924, opened 2016-02-10
Add branch parameter to clone the right branch in dkan_init.sh If not branch is set then 7.x.1.x will be cloned.

topicus:


frankcarey:
Looks good, merging

PR #926, opened 2016-02-11
Add ahoy command to download db files from s3 bucket.

dkinzer:
Ref. civic-1394

Adds some ahoy site api commands to work with s3 asset buckets.

None of this is written in stone yet.  But we can decide on the api format now.

Note that most of this is just templates as implementation is still foggy and
depends on further decision making.

Acceptance:
==========
- [ ] As long as aliases have been added correctly, ahoy site asset-db-download
  should download db from s3 bucket.


frankcarey:
Looks good, merging

PR #956, opened 2016-02-25
Add an ahoy command to update the dkanextension

frankcarey:


dkinzer:
This passing, so I'm merging.

PR #965, opened 2016-03-02
Civic 1474 featured topics

janette:
Work in progress


janette:
use https://github.com/NuCivic/dkan/pull/970

PR #988, opened 2016-03-18
Add docker and circle behat config files on update.

dkinzer:
Ref civic-1510.

`ahoy dkan test` requires that the behat.docker.yml and behat.circleci.yml be
present.  But, current sites do not include these files.  If we add these file
during an `ahoy dkan self-update` Then there is always a simple and automatic
way to keep these files present and current in any of the data_starter type
projects.

Acceptance
==========
- [ ] When run from within a site repo `ahoy dkan self-update` will also update
the docker and circleci behat config files which are required in order to run
`ahoy dkan test` from within a site project.


dkinzer:
This is passing, so I'm merging.

PR #1059, opened 2016-04-08
Fix broken update.

dkinzer:
Ref civic-2062


dkinzer:
Solved by https://github.com/NuCivic/dkan/pull/1195

PR #1060, opened 2016-04-08
Change dkan_sitewode_menu.install to apply admin_menu_source variable…

dafeder:
## Description

There is some concern with the way the dkan_sitewide_menu was installing. We've gotten rid of code that creates missing roles, and now it simply adds the setting to any lower-level role.
## Acceptance criteria
- [ ] On a new install, site manager, editor and content creator all have the alternative admin menu command center jawn
- [ ] Same is true for an upgrade.


dafeder:
Went a different way with this..

dafeder:
Not sure if this is still an issue - revisit if possible

PR #1069, opened 2016-04-13
Civic-1955 Add topic icons to the drop down menu

janette:
Issue: CIVIC-1955

## Description
For consistency, the icon from the topic should display in the drop-down menu.

## User story / stories
- [ ] As product team we want a consistent look and functionality for topics

## Acceptance criteria
- [ ] Topic icons should display along with the terms in the Topics dropdown menu.
- [ ] Tests pass


dafeder:
I don't want to rely on the name of the menu link for the logic. Can you grab the title (for the current language, thinking toward future multilingual support) from the vocabulary by loading machine name `dkan_topics`? That will be consistent on any site.

janette:
Since we won't know what the title would be if string overrides is used I'm switching to using the
$variables['element']['#original_link']['link_title'] value

PR #1082, opened 2016-04-21
Makes tests use files from an s3 bucket instead Issue #CIVIC-1736: Upload demo content files to s3

drkloc:
#CIVIC-1736


dharizza:
Close in favor of https://github.com/NuCivic/dkan/pull/1096

Issue #951, opened 2016-02-22
Resource file size often listed as '0'


janette:
## Description
On resource pages there is a block at the bottom for 'Additional Information'
in which is a row that displays the resource file size, sometimes it works and sometimes it does not.
![screenshot_2_22_16__9_51_am](https://cloud.githubusercontent.com/assets/314172/13223458/dedcdb4a-d949-11e5-8db8-0db89de37cb6.png)

## Steps to Reproduce
Since this is an intermittent issue it may be hard to reproduce.

## Acceptance Criteria
• When a user visits a resource page the filesize should list an accurate number or not display at all.
• The file size code should also use `format_size()` to display the string representation for the given byte count

![screenshot_2_22_16__9_50_am-2](https://cloud.githubusercontent.com/assets/314172/13223422/bf6ebb48-d949-11e5-8d04-89afd567ca54.png)

`dkan_sitewide/dkan_sitewide.blocks.inc` line 116

```
// Short term fix.
if($file['filesize'] !==0 && !empty($file['filesize'])) {
  $file_info[] = array(t('filesize'), format_size($file['filesize']));
}
```

janette:
I can not reproduce the issue on vanilla dkan, this seems to be a client specific issue.

PR #1109, opened 2016-05-03
GH-951 add formating to filesize display

janette:
https://github.com/NuCivic/dkan/issues/951

## Description
file sizes are displayed as just a string of numbers, lets format those to be easier to 'read'

## Acceptance criteria
- [ ] Add a large file to a resource.
- [ ] Confirm that the size is displayed in bytes or MB

screenshot_5_3_16__4_35_pm


janette:
: ( 

not sure of the accuracy of the formatter, the file is  573 MB

dafeder:
Unfortunately, both PHP and Drupal continue to define a Kilobyte as 1024 bytes.

https://bugs.php.net/bug.php?id=46534
https://www.drupal.org/node/267883

@janette I think we should go with the standard on our platform for now. If this seems to be a problem for people or standards compliance we can revisit. Worth noting: DCAT requires reporting file sizes in bytes (in XML or JSON feeds).

PR #1111, opened 2016-05-04
Fix blog author links

msolv:
Ref: https://github.com/NuCivic/nuboot_radix/pull/82

Created for QA only.


msolv:
I did the QA here: https://ba15eb9d-e36c-48e4-821f-df693ed795d7--pr-1111.probo.build

And I get a 404 when clicking on the user name when viewing the blog post because the url has a duplicated '/users':

screen shot 2016-05-04 at 4 03 08 pm

PR #1118, opened 2016-05-06
Makes tests use files from an s3 bucket

dharizza:
Issue: #CIVIC-1736, https://jira.govdelivery.com/browse/CIVIC-2446


acouch:
Looks like we need to figure out why the datastore isn't connecting.

dharizza:
Closed in favor of https://github.com/NuCivic/dkan/pull/1096

PR #1139, opened 2016-05-19
Use nuams/web container.

dkinzer:
Ref CIVIC-2392

The new container adds does not have xdebug configured. Some new helper
commands have been added.

`ahoy docker apache start|stop|restart`
`ahoy docker xdebug start|stop`
`ahoy docker log (access|error)`

Acceptance
=========
- [x] Local errors are logged when xdebug does not get in the way.
- [x] `ahoy docker apache restart` restarts apache
- [x] `ahoy docker log` tails the error log.


dkinzer:
Note that I needed the new version of `ahoy docker exec` to implement this so that's included in this PR too.

PR #1160, opened 2016-05-27
Duplicate geojson parsing in the choropleth bundle

topicus:
DON'T MERGE QA ONLY


topicus:
Merged in https://github.com/NuCivic/visualization_entity/pull/44

PR #1167, opened 2016-06-01
civic-2697 fixing ahoy dkan theme commands

asghaier:
Ref CIVIC-2697


janette:
both the setup and watch commands working as expected, thank you @asghaier

Issue #1136, opened 2016-05-18
DKAN Datastore API cannot sort by timestamp


lcshiung:
I am testing the DKAN Datastore API. It is able to sort by all fields in the JSON returned, except the timestamp. When I use the following query:

`http://server_ip/api/action/datastore/search.json?resource_id=6bd1a4d7-93c3-4e47-b77e-8dabdb13fcad&sort=timestamp%20desc&limit=5`

I get the following error:

`["Invalid query provided, double check that the fields and parameters you defined are correct and exist. SQLSTATE[42S22]: Column not found: 1054 Unknown column 't.timestamp_' in 'order clause'"]`

After doing some tracing, I have found that this is because the function feeds_flatstore_processor_safe_name in feeds_flatstore_processor.module appends '_' to timestamp as it is a reserved word:

```
function feeds_flatstore_processor_safe_name($name) {
  $reserved = feeds_flatstore_reserved_words();
  if (in_array($simple, $reserved)) {
    $simple = $simple . '_';
  }

```

Is there any way to resolve this?

On a secondary note, the example used in [DKAN Datastore API documentation](http://docs.getdkan.com/docs/dkan-documentation/dkan-api/datastore-api) may not be updated when it suggests that it is possible to sort by country using "sort[country]=asc". I realise you have to use "sort=country asc".

dafeder:
@lcshiung: Does it work when you sort by `timestamp_`? We could add some code that "unescapes" those reserved words if they're passed in the query. Interested in submitting a PR?

Will definitely check out the documentation and update where necessary.

lcshiung:
I am new to github, and am not very sure how to submit a PR, would it be ok if I suggest how a simple fix can be done? Let me know if you would prefer that I submit a PR.

It looks like the error only affects the "sort" parameter for the API. This is because when sorting, the datastore API attempts to prevent the use of reserved words in parameters (by adding a "_" suffix). This is however unnecessary, since the original intention of reserved words (if I understood the code correctly) were to avoid clashes between resource column names and the reserved words, but only at the table creation stage.

In /profiles/dkan/modules/dkan/dkan_datastore/modules/dkan_datastore_api/dkan_datastore_api.module, comment out the following lines in function dkan_datastore_api_sort_clean($sort):

```
//  foreach ($sort as $key => $value) {
//    $sort[$key] = feeds_flatstore_processor_safe_name($value);
//  }
```

The final function looks like this:

```
[...]

PR #1171, opened 2016-06-03
DKAN Datastore API cannot sort by timestamp

lcshiung:

## Description
I am testing the DKAN Datastore API. It is able to sort by all fields in the JSON returned, except the timestamp. When I use the following query:

`http://server_ip/api/action/datastore/search.json?resource_id=6bd1a4d7-93c3-4e47-b77e-8dabdb13fcad&sort=timestamp%20desc&limit=5`

I get the following error:

`["Invalid query provided, double check that the fields and parameters you defined are correct and exist. SQLSTATE[42S22]: Column not found: 1054 Unknown column 't.timestamp_' in 'order clause'"]`

After doing some tracing, I have found that this is because the function feeds_flatstore_processor_safe_name in feeds_flatstore_processor.module appends '_' to timestamp as it is a reserved word:

```
function feeds_flatstore_processor_safe_name($name) {
  $reserved = feeds_flatstore_reserved_words();
  if (in_array($simple, $reserved)) {
    $simple = $simple . '_';
  }

```
Is there any way to resolve this?

## User story / stories
Users can make queries using the timestamp column in the API.

## Acceptance criteria
In /profiles/dkan/modules/dkan/dkan_datastore/modules/dkan_datastore_api/dkan_datastore_api.module, comment out the following lines in function dkan_datastore_api_sort_clean($sort):

```
//  foreach ($sort as $key => $value) {
//    $sort[$key] = feeds_flatstore_processor_safe_name($value);
//  }
```

The final function looks like this:

```
function dkan_datastore_api_sort_clean($sort) {
  if (strpos($sort, ' asc') !== FALSE) {
    $type = 'asc';
  }
  elseif (strpos($sort, ' desc') !== FALSE) {
    $type = 'desc';
  }
  else {
    return $sort;
  }
  $sort = explode(" $type", $sort);
//  foreach ($sort as $key => $value) {
//    $sort[$key] = feeds_flatstore_processor_safe_name($value);
//  }
  $sort = implode(" $type", $sort);
  return $sort;
}
```

I have tested this to work.

Alternatively, since this change would mean the function "dkan_datastore_api_sort_clean($sort)" serves no purpose, we could entirely remove the function, and also a call to this function in the "dkan_datastore_api_datastore_index" function.

`      //$sort = dkan_datastore_api_sort_clean($sort);
`

## PR dependencies
Please put all the PR dependencies here. That means PR needs to be merged before this one.


dafeder:
Closing, see #1136

PR #1200, opened 2016-06-16
Remove insert a definition list button from bueditor toolbar

janette:
Issue: civic-2961

## Description
Users should only be presented with mark up options that are allowed by the html filter. Definition lists are not allowed, the button to insert a definition list should be removed.

## Acceptance criteria
When logged in as a site manager, editor, or content creator, I should not see the Insert a Definition List button when using the html markdown text format.


janette:
updates moved to https://github.com/NuCivic/dkan/pull/1196

PR #1203, opened 2016-06-17
Update recline test: replace wait time to wait for characters

janette:
Replaces #1201


janette:
@dafeder looks like tests are passing fine without this now, close it if you want.

PR #1207, opened 2016-06-21
#CIVIC-3006: Default content tests fixes

dharizza:
Issue: https://jira.govdelivery.com/browse/CIVIC-3006

## Description
This is a change that we need to include into a patch release because with the actual release we got these problems:
* When running DKAN tests on client sites, some of them are failing because they still relies on default content; here we are trying to fix this as much as possible.
* We are not able to run DKAN tests locally, that fix has already been included into 7.x-1.x but we need to have this back in the release.

## Acceptance criteria
- [x] Tests passes.
- [x] We're able to run DKAN tests locally again.

## PR dependencies
- [x] https://github.com/NuCivic/dkanextension/pull/45
- [x] https://github.com/NuCivic/dkanextension/pull/47


acouch:
I can confirm tests pass locally.

acouch:
I added `@customizable` to scenarios that DKAN sites would likely not have.

PR #1208, opened 2016-06-23
Merge 7.x-1.12.3 changes into 7.x-1.x

dafeder:
Tests need to pass, needs general QA

## PR Dependency
https://github.com/NuCivic/dkan_dataset/pull/238


janette:
@dafeder @acouch 
we will need a deploy step to remove the extra 'groups' menu link, 
-> disable groups page, then enable groups page
maybe something like this
https://www.drupal.org/node/1837948#comment-9163427

pages___dkan

janette:
replaced by #1220

PR #1212, opened 2016-06-23
Remove extra groups link in main menu

janette:
## Description
The menu link for the groups page was removed from dkan_sitewide_panels page manager config but it is stuck in cache. This branch adds a page manager function to fix this and will facilitate updates to the default content in the future.


msolv:
What I've checked:
- There is only one Groups link.
- I've logged in with different users with the different user roles and I've checked that the Command Center links are OK.

Worked as expected.

PR #1219, opened 2016-06-27
Parametrize ahoy dkan remake so it's possible to update specific modules

drkloc:
Issue #CIVIC-2899

### Acceptance Criteria

- [ ] `ahoy dkan remake dkan_dataset' grabs the specified version of the module in drupal.org.make
- [ ] This needs to work for every module


dkinzer:
Ugh, we let this one get stale.. but it looks good... I'll fix the conflict and merge it.

PR #1220, opened 2016-06-27
Fresh merge 7.x-1.12.3 changes into 7.x-1.x

janette:
Tests need to pass, needs general QA

## PR Dependency
https://github.com/NuCivic/dkan_dataset/pull/238


janette:
@dafeder  here's the thing
#1208 fails tests because the Markdowneditor is missing, dkan.profile is using `dkan_install_markdown` rather than `dkan_markdown_setup` so I've created a fresh integration branch, if you merge the PR on dkan_dataset, this should pass tests.

PR #1224, opened 2016-06-29
Upgrade tests to PHP 5.6

acouch:
Issue: CIVIC-3177

## Description
This upgrades CircleCI to 5.6.22.


acouch:
Looks like the memory settings might be different for 5.6? 

``Fatal error: Allowed memory size of 134217728 bytes exhausted `` b/c ``echo "memory_limit = 256M" > ~/.phpenv/versions/5.5.11/etc/conf.d/memory.ini`` no longer works?

PR #1244, opened 2016-07-11
Add hook_update to dkan_ipe for existing dev builds

janette:
## Description
Existing dev builds need the new weight value of the dkan_ipe feature


janette:
no longer needed

PR #1248, opened 2016-07-12
Civic 3330 Fix data story panelizer error

janette:
Issue: civic-3330

## Description
Error displayed when using panelizer > content on data story nodes.
```
Cannot use a scalar value as an array in panels_common_get_allowed_types()
```
Reason: need to select allowed content on panelizer config form.

## Steps to reproduce
1. Create a data story
2. Click the panelizer button
3. click 'Content'
4. You should not see the error above.

## Acceptance criteria
- [ ] No errors displayed when customizing content on a data story page.
- [ ] Tests pass (story tests were disabled prior to this, now they should be enabled and passing)


janette:
you would think I would stop doing this for every PR against the release

PR #1250, opened 2016-07-12
Add roleassign module, update permissions, add tests

janette:
issue: civic 2811
https://github.com/NuCivic/dkan/issues/1180

## Description
As a Site Manager creating a new user, I am not allowed to set the user's role in DKAN 7.x-1.12.

## User story / stories
- [ ] As a site manager I need to assign roles when adding new users.

## Acceptance criteria
- [ ] As a site manager I can assign new users any of the following roles: site manager, editor, content creator.
- [ ] As a site manager, when dkan_workflow is enabled, I can assign new users any of the following roles: Workflow Contributor, Workflow Manager, Workflow Supervisor.
- [ ] Tests pass

## PR Dependency
Adds workflow roles to the available roles for site admins to assign users to
https://github.com/NuCivic/dkan_workflow/pull/45

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,6 +1,7 @@
 7.x-1.12.4 2016-07-12
 --------------------------
 - Update permissions in DKAN Permissions module to allow anonymous users to access RDF endpoints for individual datasets.
+- Add RoleAssign module, update DKAN Permissions to give site managers permission to assign roles.
 
 7.x-1.12.3 2016-06-30
 --------------------------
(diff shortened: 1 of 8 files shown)


modules/dkan/dkan_sitewide/dkan_sitewide.strongarm.inc
```diff
@@ -187,6 +187,18 @@ function dkan_sitewide_strongarm() {
   $strongarm = new stdClass();
   $strongarm->disabled = FALSE; /* Edit this to true to make a default strongarm disabled initially */
   $strongarm->api_version = 1;
+  $strongarm->name = 'roleassign_roles';
+  $strongarm->value = array(
```
dafeder: Two things:

1. I'd rather not hardcode this with rids, we've been moving toward basing settings like this on role names and setting the variable in the db on install. 

2. We need an install hook for dkan_workflow that will add those roles if workflow is enabled

PR #1271, opened 2016-07-25
Xss fix

drkloc:
Issue #CIVIC-3513

+ Sets facets to be rewritable via `hook_facet_items_alter` in order to run `filter_xss` on them
+ Validates content type parameter from the value provided in the url: 'search/type/<content_type>`
+ Validates term ids from the value provided in the url for `search/<taxonomy_field>/<format>-<tid>` when `taxonomy_field` is either:
  + field_tags
  + field_topics
  + field_resources%253Afield_format
+ Validates authors
+ Validates licenses

### Acceptance Criteria

- [ ] `search/type/notvalid` should return a 404 for everything for everything but but valid content types
- [ ] `search/field_tags/notvalid` should return a 404 for everything for everything but but valid term values for tags
- [ ] `search/field_topic/notvalid` should return a 404 for everything but for everything but valid term values for topics
- [ ] `search/field_resources%253Afield_format/notvalid` should return a 404 for everything but valid term values for formats
- [ ] If one of Multiple ocurrances of a facet based on taxonomies is not valid, It should return a 404
- [ ] `search/field_license/notvalid` should return a 404 for everything but valid license occurances
- [ ] any invalid combo of the above options should return a 404

### DIFF
--- a/modules/dkan/dkan_sitewide/modules/dkan_sitewide_search_db/dkan_sitewide_search_db.facetapi_defaults.inc
+++ b/modules/dkan/dkan_sitewide/modules/dkan_sitewide_search_db/dkan_sitewide_search_db.facetapi_defaults.inc
@@ -80,7 +80,32 @@ function dkan_sitewide_search_db_facetapi_default_facet_settings() {
   $facet->settings = array(
     'weight' => 0,
     'widget' => 'facetapi_links',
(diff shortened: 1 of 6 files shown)


acouch:
This looks great. I was not able to inject text into the facets.

Should we add a test to verify that a 404 is produced?

PR #1283, opened 2016-07-31
Add update hook for 12.3+ patch upgrade.

dkinzer:
REF CIVIC-3573
 
 The 12.6 patch upgrades for out of the box sites has not been working as
 expected (tests are failing) because the new Markdown and role-assign features
 are not enabled or configured correctly.

## Acceptance Criteria 
 - [x] DKAN tests pass for Site upgrades with this patch combined with the patch
   to skip customizable tests.


dkan.profile
```diff
@@ -411,3 +411,109 @@ function dkan_set_roleassign_roles(&$context) {
   }
   variable_set('roleassign_roles', $roleassign_roles);
 }
+
+/**
+ * Configures BUEditor and markdown test format.
+ */
+function dkan_bueditor_markdown_install() {
```
dkinzer: @acouch this function doesn't make sense inside the profile.  It only gets called by sites on updates not on first time installation.

acouch: Ideally it would get called upon install as well as updates. As we add these functions I would think they would live in the dkan.profile.

dkinzer:
@dafeder @acouch this can be merged now.

dkinzer:
Looks like we might actually still need this PR.. need to redo.

dkinzer:
Merged release-1-12 and resolved conflict.

PR #1290, opened 2016-08-01
Civic 2686 harvested datasets

topicus:


topicus:
Test are not passing here but the source of truth is the dkan_harvest. In that repo test are passing.

PR #1295, opened 2016-08-03
Fix tag collision when testing sites.

dkinzer:
REF CIVIC-3578

The tags are not available at time of testing and causing test to fail.
Also we need to change tags to be less likely to collide.

AC
==
- [ ] sandiego oob with dkan 7.x-1.12.6 passes tests.


dkinzer:
Meant to PR to be against release-1-12 branch: https://github.com/NuCivic/dkan/pull/1296

PR #1303, opened 2016-08-05
Ensure administrator role has all permissions

acouch:
## Description
The ``administrator`` role should always have all permissions. This ensures that new permissions are not added on a module enable without updating the ``administrator`` role. 

## Acceptance criteria
- [ ] Tests still pass.


acouch:
I think this is unnecessary. adminrole already sets off hooks. I think it was a features_master issue with one client. If we want this we should push upstream to adminrole.

PR #1305, opened 2016-08-05
Fix topics page

acouch:
Issue: CIVIC-3773

## Description
Topics broke with the XSS fix. This restores the topics page.

## User story / stories
- [x] A a site visitor when I click on Topics I should see the Topics search and topics extended.

## Acceptance criteria
- [x] Tests fail then pass


acouch:
Don't pull this.

PR #1323, opened 2016-08-17
Panels and panelizer updates for security vulnerabilities

dafeder:
Make sure tests pass


dafeder:
Failing because of incompatibility between panels access changes and exported panels. See https://www.drupal.org/node/2785915

janette:
@dafeder I think we should remove the IPE for the 'datasets' page, i.e. search and revise the permissions, looks like they got off track
people___dkan

dafeder:
Yes - removing IPE seems like a separate ticket though. Let's just get panels working again first (and update any permissions that are wrong).

PR #1339, opened 2016-08-26
On dev branch, test static beforecreate function

dafeder:
Test new branch of dkanextention and making beforecreate and aftercreate functions static. 

This branches off of the behat-update branch, which branches off of 7.x-1.x.

It cherry-picks a commit from the static-beforecreate branch, which is a branch off of the _dcat_ branch, not 7.x-1.x.

If we merge this from 7.x-1.x, should simply rebase 7.x-1.x into dcat rather than merging the static-beforecreate branch into dcat.

### DIFF
--- a/test/composer.json
+++ b/test/composer.json
@@ -2,7 +2,7 @@
     "require": {
         "behat/behat": "3.1.*@dev",
         "devinci/devinci-behat-extension": "dev-master",
-        "nucivic/dkanextension": "master-dev"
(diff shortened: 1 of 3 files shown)


dkinzer:
@dafeder I've read in the following link that whether to make these static or not depend on whether they are invoked before a feature instance is created or not.  Once it is created then the call happens from the instance: http://docs.behat.org/en/v2.5/guides/3.hooks.html

PR #1341, opened 2016-08-29
Add harvest source to content type facet block on search page

janette:
Issue: civic 3888

## Description
Add harvest source to the content types facet on search page

## Acceptance criteria
- [ ] Create a harvest source and view the search page, you should see Harvest Source with an icon in the content types facet block.

dkan


janette:
wrong base

PR #1350, opened 2016-09-04
Refactor `ahoy dkan test`

dkinzer:
Allow for extra params in local environment.
Switch bash with ruby.

REF CIVIC-2728

AC
==
- [x] dkan tests pass (i.e. this does not break dkan test invocation on circle)
- [ ] allows user to locally run something like `ahoy dkan test feature/xyz --name='blah blah'`


dkinzer:
We wouldn't need this at all if ahoy didn't strip away quotes.

dkinzer:
closing in favor of #1351

PR #1367, opened 2016-09-14
Fix odsm validation

janette:
Issue: civic-4101
The data.json validation is failing because of errors on the '@type' property.

**This is where the change happened**
https://github.com/NuCivic/open_data_schema_map_dkan/commit/0b6fc89bf1e5bb836ff70012cb682fae7ca3b305


janette:
@jacintocapote any idea if this will break something in dcat?

dafeder:
Merging for now, there will be another round of DCAT QA before 1.13 release.

Issue #1372, opened 2016-09-16
Version 7.x-1.12.10 - PHP Parse Error: Unexpected '[' (PHP 5.3.x)


lcshiung:
Hi,

  When trying to update DKAN to version 7.x-1.12.10, we encountered a PHP error on line 445 of /profiles/dkan/dkan.profile.

  After looking at the file, it is likely because of the use of '[]' for array assignment, which is only supported in PHP 5.4 and above. From the [DKAN documentation](http://docs.getdkan.com/dkan-documentation/dkan-developers-guide/installing-dkan), it says that PHP 5.3.x and above is supported. Is it the policy to still support PHP 5.3.x?

dafeder:
@lcshiung you're right. Out testing environment, as stated in the docs, is 5.5, so some things like this can slip through. But we avoid breaking 5.3 compatibility when possible. You could try patching from #1373 for now.

lcshiung:
Thanks. Btw, I also had to amend the following lines (line 482, 483) of dkan.profile from:
`     'html' => ['default' => $eid, 'alternative' => 0],
      'plain_text' => ['plain_text' => 0, 'alternative' => 0]`

to:

`     'html' => array('default' => $eid, 'alternative' => 0),
      'plain_text' => array('plain_text' => 0, 'alternative' => 0)`

before it finally works with PHP 5.3.x.

dafeder:
Thanks, will fix this in the distro for next release as well!

Sent by MailWise<http://www.mail-wise.com/installation/4> – Your emails, with style.:)


-------- Original Message --------

janette:
Fixed! https://github.com/NuCivic/dkan/pull/1736

PR #1373, opened 2016-09-16
Remove PHP > 5.3 assignment syntax from dkan.profile #1372

dafeder:
dkan.profile breaks on PHP <5.4 because of [] array assignment syntax. We should avoid breaking PHP 5.3 compatibility when possible. All this should need is for tests to pass.


dkinzer:
@dafeder thanks for that.

PR #1382, opened 2016-10-04
Add colors to local test results

janette:
## Description
Easier to see where errors are on local tests if the results are in color.


dafeder:
@acouch in your wheelhouse, can you review and merge?

PR #1393, opened 2016-10-10
Move all fields to hidden on view modes are using panelizer and remov…

jacintocapote:
Issue:https://jira.govdelivery.com/browse/CIVIC-4093

## Description
The default view modes of panelized content types should have all fields marked as hidden. 

## User story / stories
- [ ] As product team we need all fields marked as hidden.

## Acceptance criteria
- [ ] When you access to report status page you shouldn't see the message "This view mode is being controlled via Panelizer. For performance reasons, it is recommended to move all fields to 'hidden'. Fields not marked as hidden will be prepared for output but will not actually output, thus needlessly increasing render and page load time."

…ed dkan_data_dashboard.info.orig #CIVIC-4093

### DIFF
--- a/modules/dkan/dkan_topics/dkan_topics.features.field_instance.inc
+++ b/modules/dkan/dkan_topics/dkan_topics.features.field_instance.inc
@@ -19,9 +19,8 @@ function dkan_topics_field_default_field_instances() {
     'display' => array(
       'default' => array(
         'label' => 'hidden',
-        'module' => 'dkan_topics',
         'settings' => array(),
-        'type' => 'dkan_topic_formatter',
+        'type' => 'hidden',
         'weight' => 0,
       ),
       'search_result' => array(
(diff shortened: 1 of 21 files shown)


modules/dkan/dkan_topics/dkan_topics.features.field_instance.inc
```diff
@@ -113,9 +112,8 @@ function dkan_topics_field_default_field_instances() {
     'display' => array(
       'default' => array(
         'label' => 'hidden',
-        'module' => 'dkan_topics',
         'settings' => array(),
-        'type' => 'dkan_topic_formatter',
+        'type' => 'hidden',
```
dafeder: This leads to spacing being to great between the topic and the rest of the story. Before it looked like this:



now it's like this



Please provide a fix in the theme/css

dafeder:
Also @jacintocapote please add a changelog message

PR #1394, opened 2016-10-11
Update ahoy tools with and behat config.

dkinzer:
REF https://jira.govdelivery.com/browse/CIVIC-4257
# Description

Allows for the use of behat suites for better and more control of test configurations.
# QA Steps
- [x] `ahoy dkan test` should still work as expected
- [x] circle dkan tests should run as expected


acouch:
Once @msolv approves @dkinzer can merge.

PR #1398, opened 2016-10-11
Remove javascript tags

topicus:


dkinzer:
@topicus any chance you have the bandwidth to reproduce this for release-1-12 branch as well.  It would greatly improve the dkan_starter site testing time as well.

test/features/dataset.all.feature.orig
```diff
@@ -0,0 +1,258 @@
+@api
```
dafeder: @topicus looks like you need to configure your mergetool to stop backing up: `git config --global mergetool.keepBackup false`

topicus:
@dkinzer I'll work on that.

dkinzer:
@topicus thanks!!!

PR #1407, opened 2016-10-13
Update Chaos Tools to fix deprecated PHP4 style constructors in PHP7

colans:
## Description

With PHP 7, we can see several of the following messages immediately after a site installation completes:

> Deprecated function: Methods with the same name as their class will not be constructors in a future version of PHP; ctools_context has a deprecated constructor in require_once() (line 127 of /path/to/drupal/profiles/dkan/modules/contrib/ctools/ctools.module).

This has already been fixed upstream in [Deprecating PHP4 style constructors](https://www.drupal.org/node/2528736) and other related issues.  The latest stable release, [1.10](https://www.drupal.org/project/ctools/releases/7.x-1.10), includes all of these.

I've confirmed that updating the module from 1.9 (the current version) to 1.10 fixes these issues, and that's the only change I've made to the makefile.


dafeder:
Thanks @colans I'm trying to figure out why the tests aren't passing on this branch and expect to include this in our next point release (1.13).

dafeder:
Superseded by #1456

PR #1408, opened 2016-10-13
Remove dkan_dataset and dkan_migrate_base

janette:
not sure how this snuck back in


dafeder:
These tests keep timing out but we know no code actually changed, so merging.

PR #1415, opened 2016-10-14
Pin cli image to tag.

dharizza:
Issue: https://jira.govdelivery.com/browse/CIVIC-4369

## Description
Pin cli image to a tag instead of latest. The idea is to control the upgrade process to ahoy v2 without breaking developers environments.


dkinzer:
tag in hub is actually named 2016-10-16, need to redo.

PR #1424, opened 2016-10-17
Pin cli container.

dkinzer:
REF CIVIC-4369.

Merge downstream #1423


dkinzer:
Since this is just a correction, and tests are passing I'v merged so as not to be a blocker.

PR #1425, opened 2016-10-17
Remove more javascript tags

topicus:


test/features/dataset.author.feature
```diff
@@ -207,22 +199,22 @@ Feature: Dataset Features
     Then I should see "Dataset 07 has been updated"
     And I should see "Groups were updated on 1 resource(s)"
 
-  @noworkflow @javascript
+  @noworkflow
   Scenario: Add group to a dataset with resources
     Given I am logged in as "Katie"
     And I am on "Dataset 08" page
     When I click "Edit"
-    And I fill in the chosen field "edit_og_group_ref_und_chosen" with "Group 02"
```
dafeder: I'm a little concerned that all the work we did getting chosen stuff to work in selenium will be lost and not available if we need it. I'm going to try to add something to the readme to refer back to these commits.

PR #1430, opened 2016-10-19
Harvest fails with 'dkan_dataset_metadata_source' module enabled

rhabbachi:
https://jira.govdelivery.com/browse/CIVIC-4411
## Description

Harvesting sources while having dkan_dataset_metadata_source on fails with the following error

```
reset() expects parameter 1 to be array, string given                                                                                                                                                                                                                      [error]
File /var/www/dkan/modules/contrib/entity/includes/entity.wrapper.inc, line 1102
```
## How to reproduce
1. Make sure dkan_dataset_metadat_source is enabled.
2. Create Harvest Source node.
3. Harvest the previously created Harvest Source node.
## QA Tests
- [ ] Applying the reproduce steps should result in a successful harvest.
## PR dependencies

N/A


rhabbachi:
WIP. Need to add Unit tests.

PR #1439, opened 2016-10-19
Remove pathauto computation from migrations and fixtures

topicus:


dkan.profile
```diff
@@ -501,7 +502,9 @@ edit-menu-description
     ))
     ->condition('eid', '5')
     ->execute();
-  drupal_flush_all_caches();
+}
+
+function dkan_post_install() {
```
dafeder: @topicus can you explain what you're doing here? I don't understand this change or the idea behind the dkan_post_install() function

modules/dkan/dkan_fixtures/modules/dkan_default_content/data/data_dashboard_show?id=59c96dfb-d64e-4025-b766-d9f1c7faaa47.json
```diff
@@ -1,5 +1,6 @@
 {
     "result": {
+        "__path": "state-workforce-and-retirement-statistics",
```
dafeder: Why the `__` prefix?

dkan.profile
topicus: 1. Drupal flush caches is not needed. 
2. Given we've been doing things in the last step like clear cache or enable aggregation, etc. I think it's worth no name this operation properly instead of doing that in another install operation. In other words: name last install step explicitly to avoid things like adding multiple clear cache at different operations and forget them later.

PR #1454, opened 2016-10-24
Set html format for harvested resource in the body field

topicus:
### Acceptance criteria
- [ ] Create a harvest source using http://s3.amazonaws.com/dkan-default-content-files/files/data_harvest.json
- [ ] Go to /dataset/us-national-foreclosure-statistics-january-2012-harvest/resource/30330a6c-8f49-499d-b522
- [ ] Description should not display html tags


modules/dkan/dkan_harvest/dkan_harvest.migrate.inc
```diff
@@ -1017,6 +1017,7 @@ class HarvestMigration extends MigrateDKAN {
     $resource_emw = entity_metadata_wrapper('node', $resource_node);
 
     $resource_emw->title = $res->title;
+    $resource_emw->body->format->set('html');
```
dafeder: Yup, that would do it. Will merge as soon as tests pass

PR #1460, opened 2016-10-25
Fix dkan_workflow_permissions override and role assign js

janette:
Issue: CIVIC-4442

## Description

- Dkan Workflow Permissions overridden OOB
- dkan_workflow_permissions.js does not work for site managers
- Add tests for role pairing javascript

## How to reproduce

1.  Enable dkan_workflow
2. View feature overrides for dkan_workflow_permissions
3. log in as a site manager
4. Go to People > Create User
5. Select one of the workflow roles, confirm it does not also select a standard paired role

## QA Tests

- [x] Confirm dkan_workflow_permissions is not overridden when enabled
- [ ] Confirm that when logged in as a site manager, assigning a workflow role to a user automatically assigns the paired standard role.


modules/dkan/dkan_workflow/modules/dkan_workflow_permissions/dkan_workflow_permissions_sm.js
```diff
@@ -0,0 +1,48 @@
+(function ($) {
```
dafeder: We don't need a whole second JS file just to use a different field ID. Use the Drupal settings object to pass the correct input ID to the script.

janette:
@dafeder ok I can scratch off Learn Drupal settings object off my bucket list. Also did you see that attaching the js with `$form['account']['roles']['#attached']['js'][]` doesn't work for site managers so I am using `$form['#attached']['js'][]` for site managers, not sure if that is an issue but otherwise the js does not get added.

dafeder:
I did see that @janette, must have to do with the stupid way OG reference fields work

janette:
@dafeder this one is ready for review

PR #1463, opened 2016-10-25
Remove Federal Extras fields after uninstalling module

jacintocapote:
CIVIC-4484

## Description

ODFE fields not removed when Open Data Federal Extras module is disabled and uninstalled. 

## How to reproduce

1.  After enabling, then disabling and uninstalling Open Data Federal Extras module, the fields remain on the dataset form and two of the fields are required, and have no options.

## QA Tests

- [ ] Enable federal extra module. Then disable and uninstall.
- [ ] Go to dataset node form and you shouldn't see any field related with federal extra.


modules/dkan/open_data_federal_extras/open_data_federal_extras.install
```diff
@@ -45,3 +45,22 @@ function open_data_federal_extras_update_7002() {
           SET field_odfe_bureau_code_value = REPLACE(field_odfe_bureau_code_value, '-', ':')";
   $result = db_query($sql);
 }
+
+/**
+ * Delete fields when uninstall module.
+ */
+function open_data_federal_extras_uninstall() {
```
dafeder: @jacintocapote this works great, except that the admin menu still shows those fields as existing after the uninstall. Can you add a menu cache rebuild to the uninstall hook and then we should be golden?

modules/dkan/open_data_federal_extras/open_data_federal_extras.install
```diff
@@ -45,3 +45,24 @@ function open_data_federal_extras_update_7002() {
...
+    'field_odfe_system_of_records',
+    'field_odfe_investment_uii'
+  );
+  foreach( $fields as $field) {
+    if (field_info_field($field)) { 
+      field_delete_field($field); 
+    };
+  }
+
+  menu_cache_clear_all();
```
dafeder: I still see the fields in the admin menu but not worth another round. Merging

PR #1468, opened 2016-10-26
Add compound filters

topicus:
This PR adds the ability to use nested keys in filters, excludes, overrides y defaults. E.g. publisher.name. Syntaxis to specify nested values is: key.nested_key.other_nested_key
### Acceptance criteria
- [x] Add a new harvest source
- [x] Test filters, excludes, overrides and defaults.
- [x] Test filters, excludes, overrides and defaults using nested values.
- [x] Test should pass


janette:
@topicus I need more detailed instructions on how to test this please.

topicus:
Let me add some background: 
**Filters:** include datasets matching with the criteria.
**Excludes:** exclude datasets matching with the criteria.
**Overrides:** override certain key with a desired value.
**Default:** create a default either a key is missing or value it's empty.

In all the cases you need to specify a key and a value. Depending on the case is the behavior you will get. In other words, keys and values (you can use more than one) are parameters to use and filter, exclude, override and default are **like** functions using those params to transform a harvest source.

This PR allow you to play with nested keys. For example:  exclude datasets with the publisher name "nucivic". Currently there is no way to do this because name is a nested key of publisher (` "publisher": { "name": "nucivic", "organization": "nucivic"}`) 

This PR address that by letting you reference nested keys using a dot notation. E.g. `publisher.name`.

PR #1469, opened 2016-10-27
Handle null resources.

dkinzer:
REF CIVIC-4483

Description
===
Fixes `ahoy utils mysql-prune` php wrong argument passed warning


acouch:
I think this should be against the release branch https://github.com/NuCivic/dkan/pull/1480

dkinzer:
@acouch issue is that ahoy tools that we use in jenkins are often from 7.x-1.x, I guess we can start switching that to using release-1-12 branch instead?

PR #1475, opened 2016-10-28
Move templates out of config

acouch:
## Description

We have a templates folder in config that should be in assets.

This is for dkan starter. PR here: https://github.com/NuCivic/dkan_starter/pull/12


acouch:
Meant to do this against the release.

PR #1478, opened 2016-10-28
Fix probo admin password.

dkinzer:
# Description

Fixes password not set in probo site.


dkinzer:
Move this over to dkan_starter.

PR #1479, opened 2016-10-28
Handle null resources.

acouch:
See: https://github.com/NuCivic/dkan/pull/1469


acouch:
Wrong branch https://github.com/NuCivic/dkan/pull/1480

PR #1482, opened 2016-10-28
Improved 'Preview' page on DKAN Harvest

msolv:
Issue: https://jira.govdelivery.com/browse/CIVIC-4682 and https://jira.govdelivery.com/browse/CIVIC-4678
## Description

The following improvements were made on the 'Preview' page:
- The cache of the source is performed through Batch API so the user sees some progress while caching big files.
- The cache is only performed if the cache directory does not exist.
- Everytime the source is Edited the cache directory is emptied so a cache will be performed.
- The number of datasets on cache is displayed on top of the table.

## QA Tests
- [x] Go and create a Harvest Source.
- [x] Confirm that after clicking on 'Save' you see a progress bar and then you are redirected to the preview page.
- [x] Confirm that if you reload the page you don't see a progress bar again.
- [x] Go and edit the Source adding an Override for example.
- [x] Confirm that you see the progress bar again.
- [x] Confirm that you see the number of datasets on the top of the table on the Preview page.


dafeder:
Very clever solution!

modules/dkan/dkan_harvest/dkan_harvest.module
```diff
@@ -223,41 +223,76 @@ function dkan_harvest_page_tab_access($node) {
...
+    }
   }
 
   return $output;
 }
 
+/**
+ *
+ */
+function dkan_harvest_cache_source($node, &$context) {
```
dafeder: This seems very redundant with dkan_harvest_cache_sources() - want to consolidate the code.

PR #1488, opened 2016-11-02
Replace notes with body in harvest #CIVIC-4572

topicus:


modules/dkan/dkan_harvest/dkan_harvest.features.field_instance.inc
```diff
@@ -10,6 +10,55 @@
 function dkan_harvest_field_default_field_instances() {
   $field_instances = array();
 
+  // Exported field_instance: 'node-harvest_source-body'.
```
dafeder: @topicus can we reduce the size of this? Also let's put the field at the bottom of the form, so people don't miss the filters.

modules/dkan/dkan_harvest/README.md
```diff
@@ -43,7 +43,7 @@ Project Open Data (as well as most metadata APIs) includes many fields that are
 To access the name property for filtering or overriding, you can set `publisher.name` in the first text box and the value you want to use in the second one.
 
 
-![Add Harvest Source](images/node_add_harvest_source.png)
+![Add Harvest Source](https://cloud.githubusercontent.com/assets/381224/19936998/53e3012c-a0fe-11e6-8253-16c5270eaedf.png)
```
dafeder: Let's crop this to show just the filter fields, and move it up above the paragraph that starts "Project open data..."

dafeder: Also, the screenshot of the harvest source page should include a description now

PR #1491, opened 2016-11-04
Fix problem with update from v11 with resources without groups and parent dataset with group

jacintocapote:
Issue: https://jira.govdelivery.com/browse/CIVIC-3567

## Description

For updates from dkan v11 we have a little problem with groups associated to resources. On v11 you can create a resource and assign a group different from a datase. On this version groups aren't sync between dataset and resource. But you should have the sames groups (dataset parent and resource). This behaviour is right on v12 but if you update from v11 and in your previous version you have resources and datasets without same groups we have a problem.

## How to reproduce

1.  Install fresh dkan v11. Create 1 dataset and assign to a group. Create a resource associate to this dataset without group.
2. Update to v12. And the resource doesn't have group but the dataset yes.

## QA Tests

- [ ] Install fresh dkan v11. Create 1 dataset and assign to a group. Create a resource associate to this dataset without group.
- [ ] Update to v12 and execute update.php
- [ ] The dataset and resource should have the same group.

…ate_civic_3567 #CIVIC-3567

### DIFF
--- a/.ahoy/.scripts/behat-parse-params.rb
+++ b/.ahoy/.scripts/behat-parse-params.rb
@@ -15,7 +15,7 @@ def behat_join_params args
       "--" + key_value[0].strip + "=" + "'" + "#{key_value[1]}".strip + "'"
     else
       if [
-        "colors", "no-color", "end", "suite", "format",
+        "colors", "no-colors", "end", "suite", "format",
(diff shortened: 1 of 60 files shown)


topicus:
@jacintocapote I could not make it work. Group remain without group. Can you verify?

PR #1515, opened 2016-11-10
Make search test more generic.

dkinzer:
REF https://jira.govdelivery.com/browse/CIVIC-5123

Description
===
This test fails under certain site scenarios because it's not generic enough.


dkinzer:
Don't think this is solving anything.

dkinzer:
Weird, had to rebuild this 3 times before it passed.

PR #1521, opened 2016-11-10
Re-export command center menu

janette:
Issue: 
civic-4423
civic-4879

## Description

This will fix the 1.13 upgrade process for the command center menu.


msolv:
Tested on an upgrade. Working as expected.

PR #1525, opened 2016-11-11
Fixed default content.

msolv:
Issue: CIVIC-4507

## Description

Some datasets were still including links to Stefanie's demo site. Those links were updated so now they are relative links.

## QA Tests

- [ ] Check the description on all default datasets and confirm that there are no links pointing to Stefanie's site.
- [ ] Check that tests pass.


janette:
@msolv the related content links are not picking up the base url

msolv:
@janette We've decided to not include the references to the Data Stories or Data Dashboards as related content. I've added those back to the description section.

PR #1535, opened 2016-11-16
Upgrade Drupal core to 7.52

rhabbachi:

## Description
https://www.drupal.org/SA-CORE-2016-005

Check if the upgrade breaks tests.


dafeder:
Core and module upgrades should happen in patch releases. Closing in favor of #1536

PR #1539, opened 2016-11-17
Release branch with tests from Dev

janette:
Issue: Dev integration catchup

## Do not merge

We need tests from dev but make files from release, this will be the dkan 'core' to test the integration branches of outside repos


janette:
junk

PR #1542, opened 2016-11-18
Creation migration after creating harvest source

marciuz:
CIVIC-5133

## Description

Please check the issue https://github.com/NuCivic/dkan/issues/1540

## User story

When an harvest source is created, the migration don't exists.
With this hook integration is created on the fly.

## How to reproduce

1. Create a new node of type harvest source
2. Run drush ms to see the existent migration


dafeder:
We'll change the base branch on this back to 7.x-1.x and try to add it to the 1.14 release once the 1.13.4 release is out and we have integrated those changes back into 7.x-1.x. Apologies for the delay @marciuz - this fell off our radar for a bit!

dkinzer:
This works beautifully, but it's failing our linting so I will merge into a branch that I can commit to in order to fix the linting issue.

PR #1546, opened 2016-11-18
Fix Format search facet does not display selected terms

rhabbachi:
Issue: CIVIC-4771

## Description
The *Format* search facet does not display selected terms as expected, the block keeps the 'ctools-toggle-collapsed' class which hides the active content

## QA
1. Go to the search page
2. Select a Format term
3. Format facet item should not be collapsed (see screenshot).

image

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,6 +1,7 @@
 7.x-1.12.x
 -------------------------
 - Update ctools to 1.11
+- Fix "format" facet collapsed even when selected on the search page.
 
 7.x-1.12.11 2016-10-20
 --------------------------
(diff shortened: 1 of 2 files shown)


rhabbachi:
Tests failing during the build phase with 
```
 * [@self.22] Unable to clone feeds_flatstore_processor from             [error]
```

This does not look right to me. Re-ruing without cache.

dafeder:
Still seeing the issue on the license facet

rhabbachi:
@dafeder Good catch!

Should be fixed in the latest commits. Seems like using the facet's field alias should do the job more cleanly without the need for regex parsing.

image

PR #1570, opened 2016-11-25
Hide "Back to dataset" button for datasetless resources [CIVIC-4448]

rhabbachi:
Issue: CIVIC-4448

## Description
When the "Back to dataset" button is clicked on a resource with no associated dataset the URL changes to /node/XXX/dataset and the view looks empty.

## How to reproduce

* Create a resource without associated dataset.
* Go to the view page.
* Click on "Back to dataset".
* The URL will change and the view will be empty (only the buttons will be displayed).

## QA Tests
- "Back to dataste" button should not be available for resources without datasets.


msolv:
Tests were passing before I updated the CHANGELOG. Merging this without waiting for tests to pass again.

Issue #1578, opened 2016-11-30
Include Release Notes in Change Log


acouch:
People without the ability to read/write can't see the upcoming release notes: https://github.com/NuCivic/dkan/releases/tag/untagged-e521d895a974d1c2369e

This includes DKAN adopters or those evaluating DKAN.

I'd propose we move that to ``release.md`` so non-devs can see it.

TODO:

- [ ] Discuss this strategy
- [ ] Decide if we should just use the changelog.txt for this purpose and how the two files differ
- [ ] Move previous releases to `releases.md` file

dafeder:
Release notes are now in the documentation site: http://docs.getdkan.com/en/latest/releases/notes/index.html

PR #1579, opened 2016-11-30
Create releases.md

acouch:
Creates a releases file. See: #1578


dafeder:
Doing this in the /docs folder now

PR #1580, opened 2016-11-30
Update subtheme creation ahoy command CIVIC-4409

janette:
Issue: CIVIC-4409

## Description

The ahoy command for creating a subtheme was not working
```ahoy dkan theme new-from-kit [theme name]```

## Acceptance Steps
1. set up local build
2. run the command above (use a one word theme name or use an underscore)
3. confirm that the new subtheme was created and is enabled as the default theme


rhabbachi:
Working as expected. Merging.

PR #1582, opened 2016-11-30
Remove content padding on data extent and social blocks Civic 3946

janette:
## Description

Global block content padding is not needed on the data extent block or the  social links block. 

## QA Tests

- [x] View a dataset that has a data extent block, confirm that the block content extends to the edge of the block.
- [x] Confirm that the social links extend to the edge of the block

crime_data_for_the_ten_most_populous_cities_in_the_u_s____dkan


msolv:

PR #1599, opened 2016-12-07
Updates datastore cron workflow

acouch:
## Description

This fixes a couple of bugs with running the datastore via cron.

The ``dkan_datastore_queue_import_worker()`` was calling a method ``updateByFile()`` that copied the file attached to the resource even if it already existed. The method was also being passed the ``$url`` instead of the ``$uri`` which meant that once cron was run the actual file was replaced with a URL.

I broke up the ``updateByFile()`` method to ``updateByFileURI()``, ``updateFile()``, and ``updateByCLI()``.

``dkan_datastore_queue_import_worker()`` now just calls ``updateByCLI()`` since the cron job has no reason to care about the file being attached. If files need to be attached then they can use the ``updateFile()` or ``updateByFileURI()`` methods. 

## How to reproduce

1. Tests wouldn't pass on DKAN Starter or locally. 
2. Running ``dkan_datastore_queue_import_worker()`` with a resource uuid and fid locally erases the file on a resource and replaces it with a URL:

screen shot 2016-12-06 at 3 38 27 pm

## Acceptance Criteria

- [x] Tests pass
- [x] ~~Either write tests for or remove ``updateByFile()`` and ``updateByFileURI()`` methods~~ Create separate ticket for datastore class tests

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -59,7 +59,8 @@
  - Upgrade search_api to 1.20
 DKAN Datastore:
  - Added DKAN Datastore module into DKAN Core.
- - Greatly expand the datastore API to support aggregation functions, better joins, multiple queries. See dkan_datastore_api's README file
+ - Greatly expand the datastore API to support aggregation functions, better joins, multiple queries. See dkan_datastore_api's README file.
+ - Fixed bugs when running the datastore via cron.
 DKAN Harvest:
  - Added DKAN Harvest module into DKAN Core.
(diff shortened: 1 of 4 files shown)


acouch:
Lets handle the PHPUnit tests in a separate ticket: https://github.com/NuCivic/dkan/issues/1601

msolv:
Tests were passing before updating the changelog. Merging this without waiting for tests to pass again.

PR #1614, opened 2016-12-13
Fix location warning when creating a remote resource

topicus:
## Description
When we create a resource using a remote file we are getting:

```
Notice: Undefined index: Location in /Users/aaron/docker-share/dkan/dkan/modules/dkan/dkan_dataset/includes/getRemoteFileInfo.php on line 210
string(59) "Hospital_Inpatient_Discharges_by_DRG__Northwest__FY2011.csv"
```

## QA Steps
- [ ] Create a resource using a remote file http://s3.amazonaws.com/dkan-default-content-files/files/Polling_Places_Madison_0.csv (don't press select)
- [ ] Save
- [ ] No errors should be displayed


acouch:
I can confirm this works with ``https://data.wa.gov/api/views/mu24-67ke/rows.csv?accessType=DOWNLOAD``.

PR #1624, opened 2016-12-16
[CIVIC-3876] XML endpoints labeled as XSL

rhabbachi:
Issue: CIVIC-3876

## Description

Harvest source with XML resources are returning an error.

## How to reproduce

1. Create a harvest using http://data.medicare.gov/data.json.
2. Add a filter:title | Healthcare Associated Infections - Hospital
3. Harvest the source.

*Expected*: The dataset is harvested without error. All 4 resources is there.
*Actual*: The harvested dataset will be missing a resource. The harvest will return an error `Resource remote URL ([Redacted]) extension () not allowed`.

## QA Tests

- Running the *How to reproduce* steps should work as the expected result.


dafeder:
Having some memory issues when trying to reproduce original bug, @rhabbachi to investigate.

rhabbachi:
Running the harvest from drush with the (newly added --instrument option) shows that almost all the time harvesting is spent on the "HarvestMigration::complete" method.

image

Didding some more to find out the root of the issue.

PR #1625, opened 2016-12-17
[CIVIC-5112] Extension detection mismatch lead to error during Harvest

rhabbachi:
Issue: CIVIC-5112

## Description
This was reported by @marciuz afteris facing issue when harvesting xls resources with `application/vnd.ms-excel` mimetype with the following error:

{code:none}
Resource remote url ([Redacted])   [error]
extension (xlb) not allowed
{code}

The remote file mimetype is {{application/vnd.ms-excel}}, which can have mulitple extensions pre the default Drupal mimetype library {{xlb}}, {{xlt}}, and {{xls}}. Faced with this choice the Harvest will return the first extension {{xlb}} as the detected extension which is not part of the allowed extension list and will generate the error.

## Steps to Reproduce
1. Using a data.json, set a resource url to xls file with 
2. Harvest the source.

*Expected*: The harvest should pass with no error. The harvested dataset should have a resource with the {{testdoc.xls}} remote file showed as an exel file. 
*Actual*: Same error is returned after the harvest. No resource with exel file.

## Acceptance Criteria
1. When detecting the remote file extension and faced with multiple possibilities. Use the extension from the URL as a tie breaker. 

h2. Test Updates
1. Add a test for the GetRemoteFileInfo class with {{application/vnd.ms-excel}} file.

## Changelog
N/A

### DIFF
--- a/modules/dkan/dkan_dataset/includes/getRemoteFileInfo.php
+++ b/modules/dkan/dkan_dataset/includes/getRemoteFileInfo.php
@@ -1,12 +1,5 @@
 <?php
 
-/**
- * @file
- * Class to get content type and name of remote file.
- *
- * Socrata shim copied from data.gov.
- */
-
 namespace dkanDataset;
 
 /**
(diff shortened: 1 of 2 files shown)


marciuz:
Hi Riadh,

IMHO can be better if you declare :
`private $info;`
at the beginning of the class. The idea to (re)use the $this->info is very good, but without declaration the code is less readable (and maybe isn't strictly correct? I'm not sure about it).

PR #1642, opened 2016-12-26
Remove eva module

topicus:
Issue: CIVIC-5066

## Description

Since it's not longer necessary this PR remove EVA module from dkan.

## QA Tests

- [x] Log in as administrator
- [x] User profile search should work as before


acouch:
I think we need to identify the old views that are using EVA and disable them on an update hook.

* ``user_profile_search``
* ``groups_search``
* ``og_count``

topicus:
@acouch I'm not sure because:
```
groups_search
og_count
```
Are not in 1.12 and:
```
user_profile_search
```
Can be handled as an upgrade step by reverting the feature. 

@dafeder What do you think? I'm not sure about this last point.

topicus:
@acouch Can you check my comment and confirm this is ok?

acouch:
@topicus if we don't disable old views that rely on EVA sites will break. Perhaps it would just be best to disable all views using EVA.

topicus:
@acouch This two:
```
groups_search
og_count
```
Were removed in a previous version. I've checked  also that following the upgrade steps (https://github.com/NuCivic/dkan/releases/tag/untagged-cbbe94e4b64eb58ca682) from release-1-12 to civic-5066-eva swap this `user_profile_search` to a regular view.

Then only difference between eva views and the rest it's the param in the new_display function. Instead of using 'entity_view' (EVA) we use 'panel_pane' or block. 

Since upgrade steps includes feature revert and all the eva views were in code an update function it's not needed.

acouch:
@topicus if anyone overrides these in the UI then they are no longer drawn from code.

topicus:
Humm, if a person overrides anything but the view id after upgrade steps should work without any issue because it's a setting in the view. Upgrade steps includes feature reverting so the view should reflect the code state after revert.

Then only way it could fail it's when a user change the view id but in that particular case an update hook will fail as well.

acouch:
> Upgrade steps includes feature reverting so the view should reflect the code state after revert.

Where are the upgrade steps? https://github.com/NuCivic/dkan/releases/tag/untagged-cbbe94e4b64eb58ca682

I don't think we revert views on our own sites: https://github.com/NuCivic/dkan_starter/blob/master/config/settings.custom.php#L57

Which features need to be reverted?

topicus:
This are the upgrade steps. They are in the `Upgrading to the latest version` section.  Those include reverting views (ahoy drush fra).

```
ahoy drush rr
ahoy drush fr dkan_dataset_content_types -y
ahoy drush fr dkan_permissions -y
If ODFE is enabled then: ahoy drush fr open_data_federal_extras -y
If ODFE is enabled then: rm -rf sites/all/modules/open_data_federal_extras
ahoy drush fr open_data_schema_map_dkan -y
ahoy drush updatedb -y
ahoy drush en dkan_ipe -y
ahoy drush en dkan_harvest_dashboard -y
ahoy drush en menu_admin_per_menu -y
ahoy drush fra -y
ahoy drush rr
```

acouch:
👍

PR #1644, opened 2016-12-27
[CIVIC-5161] Harvest will not process Resources URL that does not have a scheme

rhabbachi:
Issue: CIVIC-5161

## Description
Sometimes Harvest Sources will provide URL resources that does not provide a scheme. Currently those URLs will fail the initial checks and throw an error.

A workaround would be to append a default scheme (http) to the provided URL if parse_url fails to find one.

Also The harevst test grew too big and monolithec. I separated them into 2 test phpunit files to make adding tests easier.

## How to reproduce
1.  Create a new Harvest Source using the data.json file provided `dkan_harvest_datajson_test_schemeless_resource.json`

*Expected*: Harvest parses the URL and add the appropriate resource node to the parent dataset.

*Actual*: Harvest will return an error and fails to import the resource.

## QA Tests
- [x] Reproducing the issue should import the schemeless URL as a resource. 

## Dependent PR
- [x] This PR needs https://github.com/NuCivic/dkan/pull/1647 to be merged first.


test/phpunit/dkan_harvest/DatajsonHarvestMigrationScenariosTest.php
```diff
@@ -495,7 +521,6 @@ public function testResourceRedirect() {
    * drops some taxonomies which will make following tests fail.
    */
   public function testHarvestError() {
-    return;
```
rhabbachi: This was a left over from a previous debugging session. Dropped the return statement to re-enable the test.

PR #1667, opened 2017-01-10
Fix menu source for dkan workflow roles

topicus:
Issue: civic-5267

## Description
Admin menu source weren't properly configured for dkan_workflow roles. So we needed to add function to configure the source menu for those roles during the install. We also needed to create hook_update to fix this in the current instances.

## QA tests new installations
- [x] Check all the workflow roles point to the command menu center in `/admin/config/administration/admin_menu/source`

## QA test old installations
- [x] Make dkan using release-1-12
- [x] Enable dkan_workflow
- [x] Update dkan_worflow with branch civic-5267-visualization-menu-missing
- [x] Run `drush updatedb`
- [x] Check all the workflow roles point to the command menu center in `/admin/config/administration/admin_menu/source`


msolv:

PR #1676, opened 2017-01-16
When is enabled dkan workflow, content harvested is showed as draft status

jacintocapote:
Issue: CIVIC-5254

## Description

When is enabled dkan workflow, content harvested is showed as draft status. 

## How to reproduce

1.  Enable dkan workflow.
2. Create a node harvest resouce and import content.
3. Go to admin/workbench/drafts-active and you will see content as draft status.

## QA Tests

- [x] After enable dkan workflow and harvest some content you don't see the content harvested on admin/workbench/drafts-active


msolv:
Tests were passing before updating the CHANGELOG. Merging this without waiting for tests to pass again.

PR #1696, opened 2017-02-02
Disable dkan dataset api and enable open data schema #CIVIC-5550

jacintocapote:
Issue: https://jira.govdelivery.com/browse/CIVIC-5550

## Description

Module dkan dataset api is obsolote we need to disable and enable open data schema map.

## How to reproduce

1.  Enable dkan dataset api and open data schema map. We have conflicts between both modules.

## QA Steps

- [ ] Enable dkan dataset api.
- [ ] Execute updates and should be disable and enable open data schema map.

### DIFF
--- a/dkan.install
+++ b/dkan.install
@@ -237,15 +237,6 @@ function dkan_update_7009() {
 function dkan_update_7010() {
   if (module_exists('dkan_dataset_api')) {
     module_disable(array('dkan_dataset_api'));
-
-    //Clear endpoint cache.
     module_enable(array('open_data_schema_map'));
-    //Clear cache after disable dkan_dataset_api
-    $apis = array('data_json_1_1', 'data_json');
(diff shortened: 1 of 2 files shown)


dkan.install
```diff
@@ -230,3 +230,22 @@ function dkan_update_7009() {
...
 }
+
+/**
+ * Disable deprecated dkan_dataset_api and enable open_data_schema_map.
+ */
+function dkan_update_7010() {
+  if (module_exists('dkan_dataset_api')) {
+    module_disable(array('dkan_dataset_api'));
+
+    //Clear endpoint cache.
```
dafeder: It looks like the update hook creates the odsm filecache when the module is enabled… i don’t think we want to do that

If someone doesn’t have cron set up to clear the cache regularly they will never see their endpoints updated

PR #1713, opened 2017-02-13
Point README links to new docs site

dafeder:
Just points the docs links in the readme to the new DKAN documentation site. There should be no broken links in the README now.


msolv:
This is only a change on the READMY. Merging without waiting for tests to pass.

PR #1726, opened 2017-02-21
Generate link in the search result correctly even if node doesn't have an alias

blauerberg:
Issue: CIVIC-5768

## Description

Link in the search result won't generate correctly if node doesn't have an alias.

## How to reproduce

1. disable alias pattern of dataset in `admin/config/search/path/patterns`.
2. remove existing aliases in `admin/config/search/path/delete_bulk`.
3. (Or create new dataset without an alias)
4. search the dataset.
5. in search result, title of dataset will be point to the root path of the site.


rhabbachi:
Thanks @blauerberg. We are tracking your contribution in our internal Jira board. Hopfully will be back to you with some review feedback really soon!

dafeder:
Thanks @blauerberg!

PR #1729, opened 2017-02-22
Update views to 3.15

dafeder:
Responds to https://www.drupal.org/node/2854980

No functionaltiy changes expected, just make sure tests still pass


dafeder:
Cloising in favor of #1781

Issue #1732, opened 2017-02-23
Error: Call to undefined function features_get_default() on GUI install


deirdrehall:
I am trying to install dkan-drops-7-master (the fully built version) under xampp in Windows 7 Professional. Every attempt, I get to DKAN final setup tasks and get the error:

An AJAX HTTP error occurred. HTTP Result Code: 500 Debugging information follows. Path: http://localhost/dkan/install.php?profile=dkan&locale=en&id=2&op=do StatusText: Service unavailable (with message) ResponseText: Error: Call to undefined function features_get_default() in dkan_build_menu_links() (line 270 of C:\xampp\htdocs\dkan\profiles\dkan\dkan.profile).

Is there a way to fix this and proceed?

dafeder:
Hi @deirdrehall. This has been a very tough bug to squash and has been around for a while. I can tell you that in almost all cases the site does build just fine, and if you navigate to your new DKAN site you'll find everything as it should be. Also, if you install with Drush rather than the web interface I would not expect to get any errors. I will keep this open though and take another pass at fixing.

Let us know how it goes.

deirdrehall:
If I nagivate to the site, I get a growing number of errors such as: 
Notice: Undefined offset: 1 in nuboot_radix_preprocess() (line 44 of C:\xampp\htdocs\dkan\profiles\dkan\themes\nuboot_radix\includes\panel.inc).
Notice: Undefined offset: 1 in nuboot_radix_preprocess() (line 44 of C:\xampp\htdocs\dkan\profiles\dkan\themes\nuboot_radix\includes\panel.inc).
Notice: Undefined offset: 1 in nuboot_radix_preprocess() (line 44 of C:\xampp\htdocs\dkan\profiles\dkan\themes\nuboot_radix\includes\panel.inc).
Notice: Undefined offset: 1 in nuboot_radix_preprocess() (line 44 of C:\xampp\htdocs\dkan\profiles\dkan\themes\nuboot_radix\includes\panel.inc).
Notice: Undefined offset: 1 in nuboot_radix_preprocess() (line 44 of C:\xampp\htdocs\dkan\profiles\dkan\themes\nuboot_radix\includes\panel.inc).
Notice: Undefined offset: 1 in nuboot_radix_preprocess() (line 44 of C:\xampp\htdocs\dkan\profiles\dkan\themes\nuboot_radix\includes\panel.inc).
[...]

PR #1735, opened 2017-02-23
Increase PHP version #

acouch:
## Description

Docs say DKAN can use PHP 5.3 which doesn't work. See #1732 .


dafeder:
We're trying to maintain backward compatibility. We'll strip that array syntax out of dkan.profile, I thought it had all been eliminated at this point.

PR #1750, opened 2017-02-27
Update the storytelling documentation CIVIC-5583

janette:
Issue: CIVIC-5583

## Description

Improve storytelling section in the documentation


janette:
Not sure if comments belong in this section or if it needs to be in the docs at all

PR #1753, opened 2017-02-28
Apply fix for datastore replaces NA values as zeros.

dkinzer:
REF CIVIC-5354
REF nucivic/feeds_flatstore_processor#9

QA Steps
==
See nucivic/feeds_flatstore_processor#9


dkinzer:
PR for QA purposes only.

PR #1761, opened 2017-03-02
Add @no-main-menu tag to tests that require the main menu CIVIC-5746

janette:
Issue: CIVIC-5746

## Description

Clients that disable the main menu will fail two tests that require the main menu links, adding a tag so we can skip these tests when we need to.


janette:
The test for viewing a published resource does not need to involve the main menu, I'm going to re-write that test to just go straight to the dataset page.

PR #1762, opened 2017-03-02
Fix hook updates CIVIC-5882

janette:
Issue: CIVIC-5882

## Description

Update hooks were mixed up during release integration causing some newer hooks not to run on updates to 1.13

## Reminders
- [x] CHANGELOG updated.


dkinzer:
Tested patch on a site and it works.

PR #1766, opened 2017-03-03
Change order of steps to pass on client site CIVIC-5746

janette:
Issue:  CIVIC-5746

## Description

A client site is unable to pass the dataset.all.feature test (Scenario: Order datasets by "Date changed" by oldest first.) without this change in order


drkloc:
I tested the feature on local. It's Working.

test/features/dataset.all.feature
```diff
@@ -83,8 +83,8 @@ Feature: Dataset Features
       | Dataset 2 years ago   |  Yes       | Test        | -2 year      |
       | Dataset 1 year ago    |  Yes       | Test        | -1 year      |
       | Dataset 3 years ago   |  Yes       | Test        | -3 year      |
-    And I search for "Dataset"
     And I am on "Datasets Search" page
```
dafeder: I think technically this should be "when" instead of "and" (although apparently it runs the same way with either one)

PR #1767, opened 2017-03-03
Test datastore_api empty value handling.

dkinzer:
REF CIVIC-5903
REF CIVIC-5364

Description
===
PHPUnit tests should address the NULL value case fixed in the parent ticket,
including API response. Add some rows to the sample CSV that address the use
cases this was done to address.


drkloc:
Run this locally. Test is capturing the usecase.

PR #1785, opened 2017-03-09
Fix UI install

### LINKED ISSUE #1732: Error: Call to undefined function features_get_default() on GUI install

acouch:
CIVIC-6070

## Description

This fixes #1732 by ensuring the ``features.export.inc`` file is available during install.


dafeder:
@acouch thanks. Should the conditional be needed? `require_once()` should take care of that.

acouch:
I couldn't get ``module_load_include()`` to work. I'm not sure why. It seems safer to use that. I'm not sure if there is a context where ``DRUPAL_ROOT`` wouldn't work. Maybe it would be better to see if that file exists and if not throw an error or note that the function didn't perform.

dafeder:
But what i mean is, just at a PHP level require_once shouldn't let you load a file that's already loaded, right?

acouch:
I think ``require_once()`` only loads the file once and doesn't throw an error if you have already loaded it: https://repl.it/GOqQ/6 http://stackoverflow.com/questions/2418473/difference-between-require-include-and-require-once

acouch:
@dafeder are you waiting on anything for this? All UI installs are broken until this gets in.

acouch:
Bump?

dkinzer:
@acouch I tested this locally, you need to be using ` module_load_include('inc', 'features', 'features.export');`

acouch:
@dkinzer So what does that mean? Can we pull this in already?

acouch:
I tested this on Pantheon and the full install works without timeouts with this patch.

dkinzer:
@acouch Your code has an error on it: 

`module_load_include('inc', 'features', 'export')` always returns false because it looks for a file called `export.inc` in the features module path which of course does not exist.

What is needed is to look for the file `features.export.inc` by using  `module_load_include('inc', 'features', 'features.export')` or `module_load_include('export.inc', 'features', 'features')`.

Please review #1818 to see if that also fixes your issue.

Thanks.

acouch:
Argh. That looks right. Closing this issue.

PR #1786, opened 2017-03-09
Add og admin role to editors when adding user to a group CIVIC-5962

janette:
Issue: CIVIC-5962

## Description

Editors are not able to edit datasets created by other users when inside a group.

## How to reproduce

1. Add an editor to a group.
2. Log in as that editor user.
2. Navigate to a dataset that belongs to that group that was created by another user.
3. Notice that you are unable to edit the dataset.

## QA Steps

1. Add an **editor** to a group.
2. You should see a message that the user was given the administrator role in the group.
3. Log in as that editor user.
4. Navigate to a dataset that belongs to that group that was created by another user.
5. Confirm that you are able to edit the dataset.
6. log in as a site manager
7. remove the editor role from the editor user
8. You should see a message that the og admin role was also removed.
9. Go to the group members page, confirm that the user no longer has the og admin role.
10. Now edit the user and add back the editor role
11. Confirm you see a message that the user was also assigned the og admin role for the group they were still members of.

## Reminders
- [x] CHANGELOG updated.
- [x] Update release notes stating that existing editor users will NOT be retroactively given the role, and these users' memberships will need to be manually edited.


janette:
Also addressing minor problems with the group content node:
1. Removing horizontal tab
2. Fix margin on group block

committee_on_international_affairs___7_x-1_x

committee_on_international_affairs___dkan

msolv:
Tests were passing before the changelog was updated. Merging this without waiting for tests to pass again.

PR #1803, opened 2017-03-15
Switch access to featured groups sort order view from role to permission

janette:
Release glitch

## Description

Client sites may have various role ids for the site manager role. The featured groups sort order view has access control set to the role, which will break if the role id is not 4. Need to switch back to a permission.

## Acceptance Critera
Tests pass on client site


msolv:

PR #1810, opened 2017-03-21
Do not throw exception when there is nothing to index.

dkinzer:
REF CIVIC-5646

Throwing an exception when there are no textfields to query is useless and
breaks the preview.

QA Steps
===
- [ ] Load a CSV resource that only contains numerical info
- [ ] Import resource into the datastore
- [ ] Verify that preview still works


modules/dkan/dkan_datastore/modules/dkan_datastore_api/dkan_datastore_api.module
```diff
@@ -622,9 +622,6 @@ function dkan_datastore_api_add_index_query(&$data_select, $table, $query) {
       throw new Exception('Text search currently only works with MySQL, PostgresSQL or SQL Server datastore.');
     }
   }
-  else {
-    throw new Exception('No text search compatible fields on this datastore.');
-  }
```
dafeder: Just noting that this allows searches that will not return good results because numbers are not indexed for searching. I think the trade-off is acceptable because search is kind of broken either way in the case of fully numeric tables, and at least this way the preview works. Perhaps in the future we can handle the searching on such tables better.

PR #1813, opened 2017-03-23
Update the groups page view to sort alphabetically CIVIC-5990

janette:
Issue: CIVIC-5990

## Description

Currently the Groups page sorts groups by the date when they were last added, which may be appropriate for a 'recently added groups' block but not for a page that lists all groups on the website. 
The current order in which groups are displayed is counter-intuitive. It does not tell site visitors why groups are appearing in the order in which they appear, and there are no options for filtering or changing the sort order. This makes it appear that groups are being listed in an entirely random order.

## How to reproduce

1.  Click **Groups** in the main menu, confirm that the groups are **not** sorted alphabetically

## QA Steps

- [ ] Click **Groups** in the main menu, confirm that the groups **are** sorted alphabetically

## Reminders
- [x] CHANGELOG updated.


dafeder:
We should be sorting the front page groups view the same way as the main page.

PR #1818, opened 2017-03-27
Fix module_load_include statement.

dkinzer:
REF #1785


acouch:
Confirmed this works on Pantheon. @dkinzer = :100:

PR #1821, opened 2017-03-28
dkan_dataset module is trying to load dkan-flaticon.css when this file is now moved to nuboot_radix

oriole9g:
Issue: 1819


dafeder:
Looks good @oriole9g!

PR #1826, opened 2017-03-29
QA upgrade of nvd3 library CIVIC-6068

janette:
Issue: CIVIC-6068

## Description

QA only, testing charts with updated nvd3 library


topicus:
Closing this because is not longer needed for QA.

PR #1829, opened 2017-03-29
Update t() arguments naming scheme

rhabbachi:
## Description
Update argument variable names to match template string.

This updates the argument variable names to match the t() template naming. The argument are missing the "@" sign which is generating messaging like this:

> Key **@license** was automatically generated because doesn't exists in the dataset **@Nursing Home Compare**. Make sure you want to modify the key **@license**

Instead of the expected message:

> Key **license** was automatically generated because doesn't exists in the dataset **Nursing Home Compare**. Make sure you want to modify the key **license**


msolv:

PR #1835, opened 2017-04-03
CIVIC-6137: Fix UX so it is clear that Public Access Field is required

sheldonrampton:
Issue: https://jira.govdelivery.com/browse/CIVIC-6137

## Description

Users who try to edit existing datasets sometimes see a confusing message that says, "Rights field is required." In fact, there is no "Rights" field, since this field has been renamed "Public Access Level."

## How to reproduce

1. See Jira ticket

## QA Steps

1. See Jira ticket.

## Reminders
- [ ] CHANGELOG updated.


dafeder:
There is a bug here, but this isn't quite it. There is a second field called "rights" that appears when the public access level is set to "private", this field is supposed to be required only if visible. However, if you have "none" selected for access level, "Rights" is both hidden AND required. The fix would be to make Rights not required when "none" is selected.

PR #1857, opened 2017-04-14
Update chosen configuration to exclude specific select fields.

dkinzer:
Description
===

REF CIVIC-5765

Trying to update the Moderation state property for moderable content types (by
default dataset and resources) is not possible because the Moderation combo box
is squashed.

QA Steps
==
- [x] The moderation state on datasets or resources should show the current moderation state of the content when dkan_workflow is enabled.


janette:
@dkinzer can you figure out how to remove the use of chosen on the order fields in the dataset form (resources & related content)

PR #1859, opened 2017-04-17
Apply jQuery update correctly on upgrade

dafeder:
CIVIC-6221

This fixes DKAN sites that did not get their jQuery versions set to 1.10 as expected after upgrading to 1.13. It also removes the `field_modified_source_date` field, which was not removed as expected on some upgrades. 

## QA Tests

1. Build DKAN from 1.12.
2. Set jquery version to 1.7.
3. Update to this branch (`113-update-fixes`) HEAD
4. Check jQuery version; should now be 1.10
5. Check to make sure `field_modified_source_date` is not there


msolv:

PR #1860, opened 2017-04-18
Fixed spelling

mohnjatthews:
Changed `requet a session cookie` to `request a session cookie`.


mohnjatthews:
I have no idea why this change isn't passing. I changed one word in the documentation, so no logic has changed.

rhabbachi:
Good catch @mohnjatthews!

PR #1863, opened 2017-04-18
Update restws to 2.7

dafeder:
CIVIC-6226

Tests should pass and basic QA should reveal no issues. This should be a very low-impact change and we should try to remove restws completely in the next point release.


msolv:
PHP Unit tests failed because the demo site was down and it's required for tests to run properly. Merging this without waiting for tests to pass again.

PR #1874, opened 2017-04-21
Remove drush command from hook_update CIVIC-6234

janette:
Issue: CIVIC-6234

## Description
if the hook update fails to apply on an upgrade, the status page will indicate there are pending updates, if you try to run the update through the UI it will fail with this error:

`Call to undefined function drush_delete_dir()` Unable to run drush on acquia

## QA steps
1. Adjust the db to think it has not run dkan_update_7013()
```drush sql-query "UPDATE system SET schema_version = 7012 WHERE name = 'dkan';"```
2. Go to the update UI `update.php`
3. Confirm the update completes without error


msolv:
Tests were passing before the conflicts on the CHANGELOG were solved. Merging this without waiting for tests to pass again.

PR #1881, opened 2017-04-26
Fix yaml syntaxt to make it standard

topicus:
## Description

Yaml syntax present within behat configuration files are not respecting
the behat spec so if you try to use ruby to parse it, parser complains 
about the percentage sign and other characters starting scalar values.

## QA Steps

- [x] Tests should pass.


msolv:
Tests passed before the CHANGELOG was updated. Merging this without waiting for tests to pass again.

PR #1901, opened 2017-05-10
Upgrade to media 2.1

dafeder:
Security update https://www.drupal.org/node/2877316


msolv:
Tests were passing before the changelog was updated. Merging this without waiting for tests to pass again.

PR #1902, opened 2017-05-10
Use states to hide/show groups on dataset form #CIVIC-5087

jacintocapote:
Issue: https://jira.govdelivery.com/browse/CIVIC-5087

## Description

We need to allow select groups when a resource is created without dataset related.

## How to reproduce

1.  Try to create a resource without dataset related and you can't select a group.

## QA Tests

- [ ] Select none value on resource create page and you can select a group to associate with it.


dafeder:
Need to change base back to 7.x-1.x, after 1.13.4 integration.

PR #1909, opened 2017-05-17
Make test login optional for client sites

topicus:
Some sites won't provide a public login button so login test step should not be required.

I've added the customizable tag so we can skip that.


dafeder:
Duplicate of #1908

PR #1910, opened 2017-05-17
Make fast import queue run during cron.

dharizza:
Issue: CIVIC-6363

## Description

When we try to import the datastore for a node using fast imports and the option for `queue_filesize_threshold`, the website displays a message that says, `File was succesfully enqueued to be imported and will be available in the datastore in a few minutes.` However, the import never runs. Even if we run the cron manually to force the import we get no success. Using drush queue-run also doesn't work. Here's the drush command and the response:

```
drush @cadepttech.test queue-run dkan_datastore_queue
Processed 0 items from the dkan_datastore_queue queue in 0 sec.
```

The issue is the callback function is not being run on crun, so here we're setting the option for skipping the queue on cron to false, that way the issue gets fixed.

## QA Steps

* Go to admin/dkan/datastore and select the option `Use fast import for files with a weight over:`, then add a size in the textbox "File size threshold".
* Go to a resource page in which the file is bigger than the size set in the step before and click on "Manage Datastore".
* Make sure the option "Use fast import" is checked and click on "Import".
* You should see a message that reads `File was succesfully enqueued to be imported and will be available in the datastore in a few minutes.`
* After the next cron run (or if you execute it manually), the resource should be fully imported into the datastore.

## Reminders
- [x] CHANGELOG updated.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,5 +1,6 @@
 7.x-1.14.x
 ----------
+ - #1910 Make fast import queue run during cron.
  - #1888 Added validation on axis tick settings on visualization_entity.
  - #1725 Fix `ahoy dkan uli` output login link.
  - #1881 Fix yaml syntaxt to make it standard. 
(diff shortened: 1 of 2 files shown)


dafeder:
Closing in favor of #1935

PR #1912, opened 2017-05-18
Adds ahoy dkan unittests command

dkinzer:
Issue: CIVIC-4954

## Description

Adds ahoy command to run phpunit tests


msolv:

PR #1914, opened 2017-05-19
Civic 6337 relax linting

dkinzer:
Issue: https://jira.govdelivery.com/browse/CIVIC-6337

## Description

Relax the coding standard so that PRs do not fail.


dafeder:
We've decided not to do this for now.

PR #1921, opened 2017-05-24
[CIVIC-5980] Remove description label for Harvest Source node view

rhabbachi:
Issue: CIVIC-5980

## Description

There is a 'Description:' label on the Harvest Source view page. It is unnecessary and should be removed, especially since we don't use it on any other content type.

## QA Steps

1. Create Harvest Source Node with a description.
2. Go the "View" tab for the Harvest Source.

The description should not have a label.

## Merge process

- None needed.

## Reminders
- [x] CHANGELOG updated.


msolv:

PR #1922, opened 2017-05-24
Fixing link styling.

mohnjatthews:
There was a space between the link text and the url, breaking the markdown styling.


dafeder:
The bigger problem is actually that the link is formatted for MarkDown but this is an RST file. Fixed.

PR #1924, opened 2017-05-24
Fix dkan_update_7012

janette:
## Description

Fix dkan_update_7012 so that if a site is still using a deprecated module, the data is not stripped from the db.


msolv:

PR #1925, opened 2017-05-25
Use download to measure file size, prevent memory crash [CIVIC-6300]

dafeder:
~CIVIC-6300~
CIVIC-5551

When harvesting a source with large resource that do not provide HTTP content-length headers, the remote stream wrapper will try to load the entire file into memory to get the size. This can easily crash the harvest.

This PR points to a new forked version of remote_stream_wrapper (which has been abandoned for 7.x). It modifies that module and has it download the file to the disk before measuring. On the product side we will fork remote_stream_wrapper in order to maintain these changes.

An additional fix in this PR is to recline module, to prevent if from rendering previews when being invoked by the search indexer, as this was exhausting the memory and in most cases not adding useful data to the search index.

## QA Tests

The source https://chhs.data.ca.gov/data.json should be harvestable from the command line.

To test just a single dataset with large resources you can try this truncated data.json

https://s3.amazonaws.com/dkan-default-content-files/files/big-resource-data.json

## Merge process

- [x] NuCivic/remote_stream_wrapper#1
- [x] NuCivic/recline#92
- [x] Re-point makefile to correct branches and merge this one
- [ ] NuCivic/jenkins_release_n_update_builds#23

## Reminders
- [x] CHANGELOG updated.


jvitello:
screen shot 2017-06-12 at 5 34 59 pm

Harvest completed as expected. Going to double check my suspicions that the conflicting identifier was causing the error.

PR #1929, opened 2017-05-31
Update Bureau codes

dafeder:
CIVIC-3772

This updates the Bureau code allowed values list in the Open Data Federal Extras module to use the updated version by @batemapf at seanherron/OMB-Agency-Bureau-and-Treasury-Codes#5. The format of the CSV changed slightly so that is reflected in the list_to_array script.

## QA Steps
* The allowed values in the program codes field reflect the updated list


msolv:

PR #1936, opened 2017-06-07
Fix slow remote_file datastore imports.

### LINKED ISSUE #2093: Remote file resources import very slowly into datastore
## Description

When trying to import a remote file resource into the datastore it takes a very long time.  This is an unexpected behavior and is likely fixable by first downloading the remote file locally before attempting to import it.

This is partly the cause of CIVIC-5292 and CIVIC-5293
 
##  Steps to Reproduce

* Add the following remote file resource to a dataset https://gisdata.nd.gov/Metadata/CSV/NDHUB.Roads_County.csv
* Attempt to import the resource into the datastore
* Observe how slowly the import takes place
 
## Acceptance Criteria

* The datastore import of a remote file should be equivalent or very near in speed to the import if the resource where uploaded.
* There is a user defined limit for the size of a remote file to be imported
* Default for size limit of remote file import is set to a variable (Site Managers should edit the variable on the /admin/dkan/datastore page even when Fast Import module is disabled)
  * Help text that should say that remote files are only supported up to X. If the user needs a larger file, it should be uploaded to their DKAN site.
  * Show an error if a user tries to import a remote file that is too large
* Any temporary file is deleted after import operation

## Documentation

update RTD with changes 

## Test Updates
Unit test could be added to verify that remote files are handled slightly differently than non remote files.
 
## Documentation Updates
If not possible to increase the speed of these types of import it should be noted in the UI so that it does not come a surprise to the user.

dkinzer:
REF CIVIC-5317
Fixes #2093 

Makes remote_file temporarily upload_file fields during a datastore import.
This has the value of bi-passing the remote-stream wrapper which is very slow
for datastore imports and can take up to 3 hours for 15 megabyte file.

Description
===
When trying to import a remote file resource into the datastore it takes a very long time. This is an unexpected behavior and is likely fixable by first downloading the remote file locally before attempting to import it.
This is partly the cause of CIVIC-5292 and CIVIC-5293

Steps to Reproduce
===
* Add the following remote file resource to a dataset https://gisdata.nd.gov/Metadata/CSV/NDHUB.Roads_County.csv
* Attempt to import the resource into the datastore
* Observe how slowly the import takes place


msolv:
Closes in favor of: https://github.com/NuCivic/dkan/pull/1940.

dkinzer:
@msolv this issue should not get conflated with the harvest stream-wrapper issue.  It's not related.

PR #1938, opened 2017-06-07
Allow skipping of features test via config.

dkinzer:
REF CIVIC-6480

Description
===
We want to run dkan tests on client sites but some client sites will have extensive customizations (HHS) that cause tests to fail in ways we can't adjust for so we need to be able to skip the whole file. Using tags to skip the feature causes a new error because there are no scenarios to run.
Acceptance Criteria

A. Find an argument in behat that avoids the No specifications found at path(s) error when using tags
OR
B. There is a section in the config.yml file to list features to skip when running tests. Something like the following

```
circle:
  test_dirs:
    - config/tests/features
    - dkan/test/features
  skip_tags:
    - customizable 
    - fixme
  skip_features:
    - dataset.admin.feature
    - dataset.author.feature
    - dataset.editor.feature
```

### DIFF
--- a/.ahoy/.scripts/behat-parse-params.rb
+++ b/.ahoy/.scripts/behat-parse-params.rb
@@ -3,7 +3,6 @@
 #
 require "base64"
 require "./dkan/.ahoy/.scripts/config.rb"
-require "pp"
 
 def behat_join_params args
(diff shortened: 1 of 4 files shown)


janette:
@dafeder this is working perfectly on HHS

PR #1940, opened 2017-06-08
Added changes from PR#1936

msolv:
Issue: CIVIC-5317

Makes remote_file temporarily upload_file fields during a datastore import.
This has the value of bi-passing the remote-stream wrapper which is very slow
for datastore imports and can take up to 3 hours for 15 megabyte file.

Description
===
When trying to import a remote file resource into the datastore it takes a very long time. This is an unexpected behavior and is likely fixable by first downloading the remote file locally before attempting to import it.
This is partly the cause of CIVIC-5292 and CIVIC-5293

Steps to Reproduce
===
* Add the following remote file resource to a dataset https://gisdata.nd.gov/Metadata/CSV/NDHUB.Roads_County.csv
* Attempt to import the resource into the datastore
* Observe how slowly the import takes place

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,5 +1,6 @@
 7.x-1.13.4
 ----------
+ - #1940 Fix slow remote_file datastore imports.
  - #1925 Fix issues with remote_stream_wrapper and recline causing Harvest and Search API to fail on certain larger resources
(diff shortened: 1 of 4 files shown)


dkinzer:
This is not part of the remote-stream fork fix it should not be comflated with that.. solves completely different things.  Reclosing this and opening the other origional issue.

PR #1942, opened 2017-06-09
Restore orphan count and delete drush commands.

dkinzer:
CIVIC-6467

Description
===
The `drush orphaned-resources-count` and `drush orphaned-resources-delete`
commands are still required because the harvest can still generate orphaned
resources.

This PR restores them.


dafeder:
This fixes a regression in which these commands were lost. However, these commands are to be used with caution because they don't adequately differentiate between "standalone" resources (resource nodes that were intentionally created without a resource) and "orphan" resources (resources left over from past harvests that no longer serve any purpose on the website.

So the delete command especially should be used with caution until this has been addressed.

PR #1943, opened 2017-06-09
Update tests to be more compatible with client sites CIVIC-6492

janette:
Issue: CIVIC-6492

## Description

- Adjust DKAN tests that need minor tweaks to work on client sites
- Update the @disablecaptcha function to apply to user forms
- Remove permissions file from dkan_harvest_dashboard (covered by dkan_permissions) if dkan_harvest is reverted it removes dashboard permissions from the site manager role.
- Update group members list view access to authenticated users and update corresponding tests
- Document behat tags on test/README.md
- Start the unique tag for every scenario process

## QA Steps

- [ ] Tests pass

## Reminders
- [ ] There is test for the issue.
- [ ] CHANGELOG updated.
- [ ] Coding standards checked.
- [ ] Review docs.getdkan.com (or in /docs) to see if it still covers the scope of the PR and update if needed.

PR #1944, opened 2017-06-12
Fix php error for dots in URL. Add new `ahoy dkan server` command

dkinzer:
Description
==
CIRCLE CI Uses the php server to server dkan and there are some issues that
cannot be reproduced locally unless we use this server.  This command enables
pages to be served via the PHP server in order to test locally some issues that
be be caused by the PHP server.

Updates circle.yml to use this new command.

Acceptance
==
- [ ] `ahoy dkan server` serves dkan.

### DIFF
--- a/.ahoy/.scripts/.ht.router.php
+++ b/.ahoy/.scripts/.ht.router.php
@@ -1,0 +1,45 @@
+<?php
+/**
+ * @file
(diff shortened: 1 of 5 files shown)


dafeder:
These changes are included in #1943; closing

PR #1955, opened 2017-06-19
Docker Local Environment Updates

kducharm:
- Eliminated use of docker-machine, does not require virtualbox any longer and uses Docker installation direct from docker.com for Mac OS and Linux
- Made proxy container start/stop/destroy as part of all containers, exposed ports, added /etc/hosts entry via scripts
- Fixed issues with install regarding database params for Docker
- Fixed uli in case of docker usage
- Fixes `ahoy dkan uli` not working if https is enabled.


.ahoy/site/.ahoy.yml
```diff
@@ -5,6 +5,18 @@ commands:
     usage: Run drush via ahoy (helpful for abstraction)
     cmd: ahoy cmd-proxy drush --root=docroot "{{args}}"
 
+  uli:
```
dkinzer: We need to move this back to `ahoy/dkan.ahoy.yml` because dkan_starter uses it but dkan_starter has it's own ahoy.yml OR we need to duplicate this in dkan_starter's own .ahoy.yml file.

dkinzer:
Looks good.  Merging.  Holding off on docs merge because I think it's  not accurate currently

PR #1965, opened 2017-06-21
Update Drupal core to 7.56

msolv:
Issue: None

## Description

Drupal 7.56 has been released. This PR was created in order to update the Drupal core on DKAN.

## QA steps

- [x] Automatic tests should pass.


webflo:
@msolv Hi, sorry i am not familiar with the dkan dev workflow. Why is this change not in 7.x-1.x? Thanks!

msolv:
Hi @webflo! We are going to release 1.13.4 very soon so all the upgrades are being done there. What we usually do is to upgrade both contrib modules and Drupal core on the next planned release. On this case, the next one is a patch release and we work on them using separate branches (1.13.4 is the current one). All changes that were made on those separate branches get merged into 7.x-1.x as soon as each patch release is out.

PR #1973, opened 2017-06-23
Add topics to Harvested datasets on migration #1970 

### LINKED ISSUE #1970: Add topics to Harvested datasets on migration
There is no direct match between the the Project Open Data schema and "topics" as they are defined in DKAN. Topics are a DKAN-specific data structure, not something that is federated. Project Open Data does have something similar called a "theme" field, but we harvest that as a separate "POD theme" field in DKAN, so even if we used the harvester's "overrides" functionality, the override would get harvested to the POD theme field and not to DKAN's Topics field.

The best approach would be to add a "Topics" field to the Harvest Source content type in DKAN. When creating a harvest source, site managers would use its Topics field to specify the topics that they want added when the harvest runs. 

## User story

As a *site manager*
I want to *add topics to a harvest source*
so that *the topics I add will get added to its datasets when the harvest runs*.

## Acceptance criteria

* A "Topics" field exists for the Harvest Source content type, using the same taxonomy as the Topics field (field_topic) on the Dataset content type.
* Site managers can add one or multiple topics to a Harvest Source using the node edit form.
* When a harvest runs, the topics specified for that Harvest Source get added to all of the datasets that it harvests.

kducharm:
CIVIC-6523

Fixes #1970 

## Description
Adds field_dkan_harvest_topic to harvest source content type, alters migration to set the dataset's topic field to topics set in the harvest source.

## User story

As a *site manager*
I want to *add topics to a harvest source*
so that *the topics I add will get added to its datasets when the harvest runs*.

## Acceptance criteria

* A "Topics" field exists for the Harvest Source content type, using the same taxonomy as the Topics field (field_topic) on the Dataset content type.
* Site managers can add one or multiple topics to a Harvest Source using the node edit form.
* When a harvest runs, the topics specified for that Harvest Source get added to all of the datasets that it harvests.

## QA Steps
- [ ] Run behat test `feature/dkan_harvest.feature`
- [ ] Create a new harvest source, set the topics. Perform the harvest, edit one of the datasets and verify the topics match.

## Reminders
- [ ] CHANGELOG updated.


dangur:
Created https://github.com/NuCivic/dkan/pull/2017 which pulls from NuCivic repo so automated tests will run.

PR #1978, opened 2017-06-27
Add description values on datasets in workflow.feature tests

janette:
Issue: CIVIC-6479

## Description
When running dkan tests on a custom build:
If body fields are set to `plain_text` then we get the error 
```Insert value list does not match column list: 1136 Column count doesn't match value count at row 1 (PDOException)```

Since RawDKANEnityContext.php treats **text** and **text_formatted** differently we need to specify a value in the dataset set ups to avoid this problem on customized sites. 

## Reminders
- [ ] CHANGELOG updated.


janette:
closing in favor of a better fix in https://github.com/NuCivic/dkan/pull/1980

PR #1662, opened 2017-01-09
Allow select groups for resources without datasets #CIVIC-5087

jacintocapote:
Issue: https://jira.govdelivery.com/browse/CIVIC-5087

## Description

We need to allow select groups when a resource is created without dataset related.

## How to reproduce

1.  Try to create a resource without dataset related and you can't select a group.

## QA Tests

- [ ] Select none value on resource create page and you can select a group to associate with it.

PR #1979, opened 2017-06-28
Allow users to change Groups on standalone resources

msolv:
Issue: CIVIC-5087

## Description

Users should be able to add or remove groups on standalone _resources_ (_resources_ that do not form part of a _dataset_ node). 

Also, the groups field had some style errors so this PR includes some CSS fixes for that.

## QA Steps

- [ ] Create a standalone resource.
- [ ] Confirm that you are able to assign the resource to a Group.
- [ ] Edit a resource that is part of a dataset. Confirm that you do not see the groups field.

## Reminders
- [x] CHANGELOG updated.


dafeder:
This replaces #1662

modules/dkan/dkan_dataset/dkan_dataset.forms.inc
```diff
@@ -342,9 +342,28 @@ function dkan_dataset_form_alter(&$form, &$form_state, $form_id) {
     $form['field_link_api'][$field_link_api_langcode][0]['#title'] = '';
     $form['field_link_remote_file'][$field_link_remote_file_langcode][0]['#title'] = '';
 
-    // Resources inherit groups from parent dataset.
-    // Hide field so users cannot manually modify that information.
-    $form['og_group_ref']['#access'] = FALSE;
+    $groups = dkan_dataset_get_group_options();
```
dafeder: It looks to me like this is not working correctly. It appears to now shows the group field on all resources

msolv: That line being removed is OK. On line 355 we are hiding or showing the field using #states so that's the reason for this change here.

msolv:
@dafeder When this PR was created the 'Datasets' group on the resource form was a common select field but now the field changed and it's an autocomplete one. We are using #state in order to display or hide the field and I think that the conditions do not work anymore now that the field type changed. This will need a new round of work.

fmizzell:
A cleaner, rebased version of this code exists here: https://github.com/NuCivic/dkan/pull/2184

PR #1984, opened 2017-06-28
Apply patch for DatabaseSchemaObjectExistsException.

dkinzer:
REF: https://www.drupal.org/node/1551132

## Description
This patch is needed when moving modules that create tables from a client site into dkan core, when you build locally the modules will run the install/enable functions and some of the tables will already exist in the sanitized database.
`Table [table-name] already exists and has data. Can't automatically recreate it.`


janette:
This is needed when moving functionality from a client site into dkan core, when you build locally some of the tables will already exist.

dafeder:
@janette can you update the description of this PR with more specific explanation of where this is needed? I'm not sure what "move functionality into core" means in this context.

janette:
@dafeder more detail than my comment? or just move the comment to the description?

janette:
Originally we used this patch on  a client site but ended up using a script to truncate tables on `.ahoy/site/.scripts/upgrades/upgrade_1_13.sh`. Concluding we won't need this  on the product.

PR #1989, opened 2017-06-29
Add reconnect step to ahoy dkan drupal-rebuild.

dkinzer:
Every once in a while the mysql docker container IP changes and this brakes the
connection because settings.php is still pointing to the old IP.  This change
resets the mysql IP in order to avoid this issue.


dafeder:
Test failure was a fluke. Merging!

PR #1996, opened 2017-07-05
Change 'Datasets' field on Resources form to use autocomplete

msolv:
Issue: CIVIC-6574

## Description

The dataset field on the resources form is currently listing all the available options which on sites with a big amount of datasets can be an issue. We've modified the field to use autocompletion, the same way we are doing now the dataset form to add resources.

This PR includes:
- Changes on 'dkan_dataset.forms.inc' and 'dkan_dataset_content_types.features.field_instance.inc' to implement the fix.
- The 'iFillInTheResourcesFieldWith' custom step was used to fill in the autocomplete 'Resources' field on the Datasets form. We needed the same kind of step in order to fill in the 'Datasets' field so the custom step was renamed to 'iFillInTheAutocompleteFieldWith' and moved out from 'DatasetContext' to 'DKANContext'.
- Several Behat tests were fixed in order to use the new step name and/or to support the new autocomplete field.
- This PR also includes several code style fixes on Contexts that are part of DKANExtension.

## QA Steps

- [ ] Go to create a resource.
- [ ] Confirm that you are able to assign it to any dataset using the autocomplete form.

## Reminders
- [x] CHANGELOG updated.


msolv:
This PR was redo here: https://github.com/NuCivic/dkan/pull/2035 using 1.13.6 as the base branch. The code style fixes on custom Contexts on DKAN Extension are not needed anymore so they were removed.

PR #2000, opened 2017-07-07
Fixes 1.13.4 upgraded issues encountered in devy upgrade.

dkinzer:
Description:
====

These changes are to fix bugs discovered in the latest rounds of site upgrades.

1. Fix dkan_bueditor_markdown_install() uses variable before it is initialized.
2. Fix dkan_update_7016 assumes bueditor id is '5' (not always true).
3. Fix dkan_topics_field_formatter_view() does not check if term exists causing drupal to throw.
4. Fix dkan_default_content_disable() call to processRollback() on null.
5. Fix harvest tests do not clean up.


dkinzer:
closing in favor of #2003

PR #2012, opened 2017-07-18
Add custom filelds and test tags.

dkinzer:
* Allow skipping of some tests.
* Add custom fields.


dafeder:
This was re-rolled in #2018

PR #2013, opened 2017-07-18
Hide toogle buttons inside preview embeds

topicus:
Ref: CIVIC-6188
## Description
When grid previews are embedded, they are missing the pagination and row count. Find out if there's some way to restore them but maintain the clean look we get on map and graph previews. (See attached image)
We may not be able to do this if it will take a ton of time. Whoever takes this on should think about how we could achieve this and check in with Dan or Mariano on the potential solution.
Tested on DKAN 1.13.3

## Steps to Reproduce
turn on preview embeds and embed the grid into the description of a dataset
compare the embed to the standard preview

## Acceptance Criteria
- [x] pagination, search, and filters show up in Grid Preview embeds- 
- [x] Grid, Graph, Map toggle does not show up in Grid Preview embeds
- [ ] none of the tools or toggles show up in Graph or Map Preview embeds


dkinzer:
This is not meeting the third acceptance criteria.

* Graph or Map preview embeds still have tools.

dkinzer:
I cannot get this branch to pass and I cannot reproduce the issue locally.  I'm going to redo the branch off of 1.13.8 and create a PR against that branch. 

Closing in favor of #2102

PR #2015, opened 2017-07-19
Add BeforeSuite file shiv.

dkinzer:
REF CIVIC-6610

When stage file proxy is enabled we need to make sure some required files are present in public://
By adding a BeforeSuite step that copies these files from a site to the local
environment, we fix errors that arise from files not being present in the
testing environment.

QA Steps
==
On a local version of devy with this patch
- [ ] Enable the stage_file_proxy module
- [ ] Verify that config.yml default:stage_file_proxy_origin is set to devy prod or set.
- [ ] Verify that configy.yml default:stage_file_proxy_files points to default_log.png or set it
- [ ] Remove default_log.png from public:// files directory
- [ ] Remove all files like dkan-topic.* from public:// files directory
- [ ] Verify that the following test does not fail: dkan/test/features/theme.admin.feature:16
- [ ] Verify that the following test does not fail: dkan
- [ ] dkan/test/features/topics.feature:48

### DIFF
--- a/test/dkanextension/src/Drupal/DKANExtension/Context/RawDKANContext.php
+++ b/test/dkanextension/src/Drupal/DKANExtension/Context/RawDKANContext.php
@@ -68,6 +68,43 @@ class RawDKANContext extends RawDrupalContext implements DKANAwareInterface {
   }
 
   /**
+   * @BeforeSuite
+   */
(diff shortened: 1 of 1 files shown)


dafeder:
This was re-rolled in #2018

PR #2016, opened 2017-07-19
Docker Local Environment Updates (#1955)

dkinzer:
Backports #1955 into 1.13.5

REF CIVIC-661

* Docker fix for mysql args

* Changes from docker-machine to docker, fixed uli

* Added 'ahoy docker hostfile' to add to /etc/hosts

* Make backwards compatible with docker-machine.

* Make backwards compatible with docker-machine.

Missed this one.

* Add dkan-uli.rb script.

Will handle function for `ahoy dkan uli`.

* Move uli command back to dkan.ahoy.yml

* Update CHANGELOG.

* Remove .ahoy.yml uli command.

### DIFF
--- a/.ahoy/.scripts/config.rb
+++ b/.ahoy/.scripts/config.rb
@@ -1,11 +1,28 @@
 require 'yaml'
 
+# Deep merge only available in ruby >= 3.0
+# TODO: replace defaults with custom deep_merge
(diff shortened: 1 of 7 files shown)


dafeder:
This was re-rolled in #2018

PR #2014, opened 2017-07-19
Reconcile PR1943 with the 1.13.5 release

janette:
patch for getting upstream changes into 1.13.5 release

REF #1943

PR #2018, opened 2017-07-19
Integrates several 1.13.5 patches.

dkinzer:
REF #2012 #2014 #2015 #2016


dafeder:
Once DKAN starter patch is created, we should change the base branch and merge this into `7.x-1.x`

PR #2039, opened 2017-07-31
Document terminus deployment

dafeder:
Terminus is Pantheon's recommended way to deploy distros now. Section added to docs with example Terminus commands.


jvitello:
@dafeder I tested the RTD build and it looks good.

I double checked the links, and the instructions look clear. The tests aren't passing, but I'm assuming you wanted to merge anyway since it's a small change to the documentation.

All clear to merge, but I'll let you do that.

PR #2049, opened 2017-08-09
Fix dkanextension does not autopopulate as expected.

dkinzer:
Fix bug where autopopulation fails either because:

(1) The field is made to be required via form_api_alter
(2) The field is set but is empty.
(3) The field label is not properly mapped in behat.yml

REF CIVIC-6552

QA Steps
===
- [ ] Continue to pass.

### DIFF
--- a/test/behat.yml
+++ b/test/behat.yml
@@ -90,6 +90,7 @@ default:
               status: status
               published: status
               body: body[und][0][value]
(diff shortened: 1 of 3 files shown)


janette:
Tests pass after turning on POD validation and requiring a group on datasets

PR #2050, opened 2017-08-09
Modify test scenarios

janette:
- Use smaller file for testing datastore
- Update workflow_19 to work when group is required

## Description
Testing the datastore functionality often fails on timing, taking too long to import or delete, so switching to a smaller file.

Reworking workflow_19 to work when datasets require a group


janette:
closing in favor of #2054

PR #2060, opened 2017-08-16
dkanextension and test changes required for 1.13.7 site upgrade

dkinzer:
REF CIVIC-6539

POD context does not currently load all available license that are added via
hook_license_subscribe.

This change fixes that issue.

Also, fixes issue with DatasetContext where fields  in the dataset node form are not found.

Also, adds a way for site implementers to skip fields to be tested for.

Also updates dataset_res_api tests to use description field.

Also updates feature files with scenario tags.

Also update WorkflowContext so that it does not fail for some site instances.

Also changes remote files to be a smaller version.

QA steps
===
- [ ] Add a new license via hook_license_subscribe.
- [ ] pod.feature should still pass.


janette:
RE: test/features/user.editor.feature

1.13.7 `Then I should see "Show items where"`
7.x-1.x `Then I should see "Operations"`

PR #2074, opened 2017-08-30
Fix iframe allowed in dataset results summary.

dkinzer:
REF CIVIC-6504

* If an iframe is the first thing in a description, the iframe will be displayed
on search results page.
* If summary is longer than 250 words summary can get cut off mid word.
* Ellipses at end of summary or not possible.

QA Steps
==
- [x] Add iframe to dataset verify that it does not show up in result summary.
- [x] Add a long body text to a dataset, verify that search results does not get cut off in middle of word by default.
- [x] Verify that by default ellipses are added to end of long auto summaries.

### DIFF
--- a/modules/dkan/dkan_dataset/dkan_dataset.module
+++ b/modules/dkan/dkan_dataset/dkan_dataset.module
@@ -1233,11 +1233,19 @@ function dkan_dataset_entity_info_alter(&$entity_info) {
  */
 function dkan_dataset_preprocess_node(&$vars) {
   if ($vars['view_mode'] == 'search_result') {
-    if (!empty($vars['body'][0]['summary'])) {
-      $body_summary = $vars['body'][0]['summary'];
+    if (isset($vars['body'][0])) {
+      if (!empty($vars['body'][0]['summary'])) {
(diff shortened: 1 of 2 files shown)


modules/dkan/dkan_dataset/dkan_dataset.module
```diff
@@ -1233,11 +1233,32 @@ function dkan_dataset_entity_info_alter(&$entity_info) {
  */
 function dkan_dataset_preprocess_node(&$vars) {
   if ($vars['view_mode'] == 'search_result') {
+    if (!empty($vars['body'][0]['summary'])) {
+      $body_summary = $vars['body'][0]['summary'];
+    }
+    else {
```
kducharm: Got `Notice: Undefined offset: 0 in dkan_dataset_preprocess_node() (line 1240 of /src/dkan/modules/dkan/dkan_dataset/dkan_dataset.module).` - may need a isset check on the body[0] in general.

dkinzer: Good catch. I'll make the code change and resubmit.

Issue #2026, opened 2017-07-24
split harvester batch process into more manageable chunks


jvitello:
## Description

The harvester GUI uses a batch process, but it completes it in one step from 0% to 100%. We have been experiencing errors when running the harvester via the GUI that seem to be related to the batch being one big process (time outs and server unavailable errors).

We think it will be a better user experience for incremental progress to be displayed. We think some of the other tickets in the backlog may also be resolved with this change.

The harvester also fails after long harvester sessions. The hope is that this will resolve those issues as well.

## Steps to Reproduce

- add https://data.chhs.ca.gov/data.json as a harvest source (it's a big source at 262 datasets)
- go to the source page, preview and harvest now
- also test migrations via the dkan harvest dashboard 

## Dev Notes

We need to figure out how much the batch should be divided. Maybe every 20 datasets per chunk.

## Acceptance Criteria

- batch process is broken into smaller chunks
- after each chunk is processed, the batch progress bar is updated for the user

jvitello:
@dangur Were you able to start this issue?

dangur:
@jvitello not significantly

jvitello:
@dangur If no one has really started, we will probably take this one back. It's becoming a higher priority.

dangur:
That is fine. Please reassign to someone who can work on it soon.

PR #2075, opened 2017-08-30
Modified harvest to process migrations in chunks

msolv:
Issue: CIVIC-6494
Connects #2026 

## Description

This PR includes all the changes needed to modify DKAN Harvest to process migrations in chunks. Harvests are performed through the migration module using a batch process. This improvement splits the migration so the user can see frequent feedback on the progress bar.
The improvement includes:
- The addition of a new Drupal variable called 'dkan_harvest_batch_items_per_chunk' used to save the value for the number of items that should be processed on each chunk. It defaults to 'DKAN_HARVEST_BATCH_ITEMS_PER_CHUNK'
- The migration process was split so a batch operation called 'dkan_harvest_migrate_source' is called as many times is needed, once per chunk.
- The default logging was disabled on the DKAN Harvest migration class. This change was needed because as default the migrate module adds one log record to the log table every time the processImport() method is called and we only wanted 1 event log per harvest source migration.
- The logging aspect was improved in order to log only 1 event log per harvest source migration.
- Bug fix: If the user tried to delete a Harvest Source with no datasets from the Harvest Dashboard and error was displayed. This PR includes a fix for that.
- NOTE: Drush commands do not use a batch in order to process the migrations but they need to be tested anyway because the code was refactored.

## Manual QA

Follows the suggested list of things to check for this improvement. All the processes listed here should be tried and checked to confirm that they work as expected. This improvement should also be tested using different 'DKAN_HARVEST_BATCH_ITEMS_PER_CHUNK' values:
- A value lower than the total number of datasets on the source.
- A chunk value that's a divisor of the total number of datasets in the source (Ex: 2 if the number of datasets is 10).
- A chunk value that's not a divisor of the total number of datasets (Ex. 3 if the number of datasets is 10).

Also, on all cases you should check:
- The progress bar being displayed as expected.
- The messages displayed to the user.
- The events section.
- The errors section.
- The number of datasets associated with the Harvest Source.

To QA:
- [x] Create a harvest source and harvest it using the "Harvest Now" button.
- [x] Re-harvest the same harvest source.
- [x] Create a harvest source with a source file with errors (There is an example JSON file on the 'data' section of the PHPUnit tests) and harvest it.
- [x] Create a harvest source and harvest it. Then update one of the datasets on the source file (ex.: the title) and harvest the source again.
- [x] Create a harvest source and harvest it. Then remove one of the datasets from the source file and harvest the source again.
- [x] Add the dataset back on the source file and harvest the source again.
- [x] Create one harvest source and go to the Harvest Dashboard. Select the source and apply a "Cache" operation. Then try the "Migrate" operation.
- [x] Create one harvest source and go to the Harvest Dashboard. Select the source and apply a "Harvest (Cache + Migrate)" operation.
- [x] Create 2 harvest sources and go to the Harvest Dashboard. Select both sources and apply a "Cache" operation to both at the same time. Then try the "Migrate" operation.
- [x] Create 2 harvest sources and go to the Harvest Dashboard. Select both sources and apply a "Harvest (Cache + Migrate)" operation to both at the same time. 
- [x] Create a harvest source. Cache it using drush: `ahoy drush dkan-harvest-cache <source_machine_name>`
- [x] Migrate the harvest source using drush: `ahoy drush dkan-harvest-migrate <source_machine_name>`
- [x] Create a harvest source. Harvest the harvest source using drush: `ahoy drush dkan-harvest <source_machine_name>`
- [ ] Create 2 harvest sources. Cache them using drush: `ahoy drush dkan-harvest-cache`.
- [ ] Migrate both harvest sources using drush: `ahoy drush dkan-harvest-migrate`.
- [ ] Create 2 harvest sources. Harvest them using drush: `ahoy drush dkan-harvest`.

## PR Dependencies

None.


dafeder:
#2026

dafeder:
Fixes #2026

PR #2079, opened 2017-08-31
NO MERGE: QA for CIVIC-6683.

dkinzer:
REF nucivic/visualization_entity#77
REF nucivic/recline.view.nvd3.js#47


janette:
marvelous!

PR #2095, opened 2017-09-05
Datastore api field exclude alter hook

kducharm:
## Description
Adds a hook_datastore_api_field_exclude_alter so that field exclusions can be modified for API queries.

## User story
As an admin, I would like to be able to not exclude some of the internal fields used in the datastore API query so they can be used as a primary key for the data.

### DIFF
--- a/modules/dkan/dkan_datastore/modules/dkan_datastore_api/dkan_datastore_api.api.php
+++ b/modules/dkan/dkan_datastore/modules/dkan_datastore_api/dkan_datastore_api.api.php
@@ -1,0 +1,25 @@
+<?php
+
+/**
+ * @file
+ * Hooks provided by DKAN Datastore API module.
+ */
+
+/**
+ * @addtogroup hooks
+ * @{
+ */
+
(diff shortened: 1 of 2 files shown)


modules/dkan/dkan_datastore/modules/dkan_datastore_api/dkan_datastore_api.api.php
```diff
@@ -0,0 +1,25 @@
...
+ * @{
+ */
+
+/**
+ * Allows fields exclusions to be altered for API queries.
+ *
+ * @param array $excludes
+ *   List of fields to exclude.
+ */
+function hook_dkan_datastore_api_field_excluded_alter(&$excludes) {
```
dafeder: Linter rejected: `Type hint "array" missing for $excludes`

dafeder:
Only error after my type hint fix was a circleCI timing issue, still waiting for it to pass but going to merge so we can move forward with 1.13.8 release.

PR #2102, opened 2017-09-07
Hide toogle buttons inside preview embeds

dkinzer:
Ref: CIVIC-6188 

When grid previews are embedded, they are missing the pagination and row count. Find out if there's some way to restore them but maintain the clean look we get on map and graph previews. (See attached image)
We may not be able to do this if it will take a ton of time. Whoever takes this on should think about how we could achieve this and check in with Dan or Mariano on the potential solution.
Tested on DKAN 1.13.3

turn on preview embeds and embed the grid into the description of a dataset
compare the embed to the standard preview

- [x] pagination, search, and filters show up in Grid Preview embeds- 
- [x] Grid, Graph, Map toggle does not show up in Grid Preview embeds
- [ ] ~~none of the tools or toggles show up in Graph or Map Preview embeds~~


dkinzer:
@dafeder this is green now and in my opinion can be merged even though last AC is not applying.   The tooling is useful for all graph cases and trying to implement the last AC is less useful and also much harder since it cannot be done simply via a css change.

janette:
needs a small css update on nuboot_radix/scss/components/recline.scss

```
@media all and (max-width:991px) {
  .recline-query-editor {
    border: none;
    margin: .5em 0;
    padding: 0;
  }
  // Remove margin if embedded.
  #recline-embeded .recline-query-editor {
    margin: 0;
  }
}
```

PR #2109, opened 2017-09-12
Update open_data_schema_map_dkan to use language field token rather t…

janette:
Issue: hhs-997

## Description

POD validation errors for the language field. It looks like {{false}} is being added as a value to the data.json when there is no defined language. Language is optional for POD.

## How to reproduce

1. create a harvest source from `http://datafiles.samhsa.gov/data.json`
2. view `admin/config/services/odsm/validate/pod`
3. confirm you see errors on lanugage

## QA Steps
1. create a harvest source from `http://datafiles.samhsa.gov/data.json`
2. view `admin/config/services/odsm/validate/pod`
3. confirm there are no errors

## Reminders
- [ ] CHANGELOG updated.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,5 +1,6 @@
 7.x-1.14.x
 ----------
+ - #2109 Fix ODSM language token to solve validation errors.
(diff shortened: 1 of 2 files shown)


dkinzer:
I've tested this on a site instance and it works beautifully.

PR #2122, opened 2017-09-22
Drop datastore on import.

dkinzer:
Description
===

DKAN sometimes repeatedly imports the same data file (a CSV file) into its
datastore (a MySQL database table) without first truncating the rows from the
previous import. This causes the database table to balloon in size, eventually
consuming an inordinate amount of disk space and creating performance problems.

This commit/change fixes this problem by always dropping the datastore table
prior to importing.  This will happen regardless of whether the import is
initiated via admin page or CLI.

Also, we remove the 'delete' option since that is now redundant.

QA Steps
===
I was not actually able to duplicate the issue locally, so I'm not sure how we can test this for sure...
- [ ] Tests pass


modules/dkan/dkan_datastore/includes/DkanDatastore.inc
```diff
@@ -341,18 +280,49 @@ class DkanDatastore extends Datastore implements DatastoreFormInterface {
    * Drop form submision.
    */
   public function dropFormSubmit(&$form_state) {
+    $this->dropDataStore();
+    drupal_set_message(t('Datastore dropped!'));
```
dafeder: Can we include the resource name in the message?

dkinzer: OK, updated the title... btw above was just a refactor so "Datastore dropped!" was the original message for drops.

dkinzer:
close in favor of #2161

PR #2146, opened 2017-09-27
WIP - Decoupled Datastore

fmizzell:
Connects #2179


modules/dkan/dkan_datastore/dkan_datastore.module
```diff
@@ -500,7 +500,7 @@ function dkan_datastore_go($id = NULL, $class = NULL) {
     $class = variable_get('dkan_datastore_class', 'DkanDatastore');
   }
 
-  return class_exists($class) ? Datastore::instance($class, $id) :
+  return class_exists($class) ? DkanDatastore::instance($id, $class) :
```
nerdstein: Is there a reason we wouldn't keep the earlier params to preserve some consistency?

modules/dkan/dkan_datastore/dkan_datastore.pages.inc
```diff
@@ -122,8 +170,11 @@ function dkan_datastore_import_tab_form($form, &$form_state, $node) {
  * Submit function for import tab.
  */
 function dkan_datastore_import_tab_form_submit($form, &$form_state) {
+  module_load_include("php", "dkan_datastore", "src/Form/Import");
```
nerdstein: Just wondering, but for everything noted as "dkan_datastore" - would this be potentially replaced with a plugin system and some configuration to define the selected data store?

kimwdavidson:
Need tests passing and fast importer working with new architecture

modules/dkan/dkan_datastore/src/Helper/Node.php
```diff
@@ -0,0 +1,35 @@
...
+      return FALSE;
+    }
+  }
+
+  /**
+   * Loads node from uuuid.
+   */
+  public static function getNodeFromUuid($uuid) {
+    static $nodes = array();
+    if (!isset($node[$uuid])) {
```
kducharm: Should this be $nodes, and can you not init $nodes to array() or it would get wiped out?

modules/dkan/dkan_datastore/src/DatastoreInterface.php
```diff
@@ -0,0 +1,53 @@
...
+namespace Dkan\Datastore;
+
+interface DatastoreInterface {
+
+  public function __construct($uuid);
+
+  /**
+   * Return the node (resource) associated with this datastore.
+   */
+  public function getNode();
```
kducharm: What do you think about making this more generic - getResource(), perhaps we have option to not store resources in nodes? Implementation of interface can decide how to get the resource... but would probably need a Resource class w/methods to do that

Also - dkan_datastore_resource_load() exists (can't find anything using it though), if you had a Resource class - a loadResourceById() method could do similar to ^^ and check if it's an integer or not so it either loads by NID or UUID. I'm not a big fan of assumed multiple ID formats though.

modules/dkan/dkan_datastore/src/Helper/Node.php
```diff
@@ -0,0 +1,35 @@
+<?php
+namespace Dkan\Datastore\Helper;
+
+class Node {
+  /**
+   * Gets nid using uuid.
+   */
+  public static function getNid($uuid) {
```
kducharm: `entity_uuid_load()` function exists if you want to use that instead, but that does run hooks on the entity load.

modules/dkan/dkan_datastore/src/DatastoreInterface.php
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * Progress of the import (0 - 1).
+   */
+  public function getImportProgress();
+
+  /**
+   * A string with datastore status information.
+   */
+  public function getStatusMessage();
```
kducharm: I think some other getStatus() interface + methods should exist with predefined vars/structure, maybe start with Feed's fields (progress_importing, progress_clearing, imported, count, next) but more generically named so we can augment in the future for other engines

modules/dkan/dkan_datastore/src/Helper/Node.php
fmizzell: you are correct, it should be nodes. the static keyword makes the $nodes variable persistent across calls to this function in the current process, so after it is initialized, the first time, it will be available to future calls of the function without getting overridden.

modules/dkan/dkan_datastore/src/DatastoreInterface.php
fmizzell: I agree with decoupling this from nodes and only being attached to the more generic concept of "resource" but as you mentioned, that would necessitate some extra abstraction that I did not want to tackle until I knew I could decoupled this thing. I can definitely change the name though and add a @todo for further decoupling later on.

modules/dkan/dkan_datastore/src/DatastoreInterface.php
fmizzell: I also like this idea, I will add this to a @todo. This abstraction might be necessary anyways once we start tackling queued imports (I am not doing that yet :) )

fmizzell:
Ok, I was hoping to get away with removing a lot of things that were not being used from the UI, but some of those things are being used by drush commands and I did not realize that until running the unit tests. So, we will have to definetely clean up the unit tests and make sure we are still testing what was intended. Also, there are a few leaky places in the API, but I hope that the unit tests will catch those also. In short, the last leg of work here is to get the unit tests passing and at that point we will be in a good place to merge.

I also would like to add @todos to capture all of the many helpful comments/feedback given so far.

PR #2148, opened 2017-09-28
Ahoy/Docker/Proxy/Docker-sync/XDebug Updates

kducharm:
## Description
- Improvements to Ahoy Docker commands.
- Converts docker-compose files to version 2.
- Separates docker-compose files into common, proxy/no-proxy, nosync/docker-sync. 
- Introduces environment variables AHOY_DOCKER_SYNC, AHOY_WEB_DOMAIN, and AHOY_USE_PROXY. Documented in `ahoy docker env`.
- Adds Traefik proxy support w/http and https when enabled via environment variable, dynamic host subdomains based on project dir name to support multiple simultaneous DKAN instances.
- Adds ability to use docker-sync, specifically for Docker For Mac usage to increase file performance 10x+ (https://github.com/EugenMayer/docker-sync).
- Fixes `ahoy docker xdebug start` for Linux/MacOS.


kimwdavidson:
Gerardo reviewing, but we also need to figure out how to push this without risking breaking all the builds.

fmizzell:
The code looks straightforward. @dafeder could you double check to make sure that we are not affecting any other external tools that use docker stuff. Obviously the CI environments still work and run the tests, but we don't know what other things depend on these configurations.

Issue #2153, opened 2017-09-28
Harvest overrides and defaults only pick up first config values


janette:
## Description
Unable to override or set defaults on multiple fields when harvesting.

## Steps to Reproduce
1. Create a harvest source `node/add/harvest-source`
2. Add two override settings
3. Add two default settings
4. Harvest and confirm that only the first override and first default are working as expected.

![test_resource___dkan](https://user-images.githubusercontent.com/314172/30989046-f0b568a2-a461-11e7-8129-42533b4cfd2d.png)

## Acceptance Criteria
- [ ] Multiple overrides can be applied to a harvest source.
- [ ] There is a message to the user for each override value set.
- [ ] Multiple defaults can be applied to a harvest source.
- [ ] There is a message to the user for each default value set.
- [ ] Tests would be nice
- [ ] Changelog.txt updated
- [ ] Documentation updated/improved

PR #2155, opened 2017-09-29
Adjust harvest functions to pick up multiple override or default values

janette:
Connects #2153 

## Description
When creating a harvest source, only the first override or default value is used, multiple value configurations are ignored.

## QA steps
- [x] Create a harvest source from the demo site http://demo.getdkan.com/.
- [x] Add two overrides: theme = something, publisher.name = Harvy
- [x] Add two defaults: author = test, describedBy: http://whatevs.com
- [x] Confirm you see a message for both override values.
- [x] Confirm you see a message for both default values.
- [x] Harvest and confirm the values are as expected.

## Reminders
- [ ] CHANGELOG updated.


janette:
@dafeder Defaults still need work

PR #2156, opened 2017-09-29
Simplify widget tests to eliminate repeated selenium problems on tests

dafeder:
The repeated testing of panels IPE "widgets" -- really just making sure Panels IPE works, which is a third party module with its own tests -- is constantly making our builds fail due to unpredictable quirks in how selenium/Chrome behave. Simply checking to make sure the IPE comes up with the correct options is a decent stopgap until we figure out a better testing strategy for this.


janette:
Getting error on @group_all_6
`SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '234-3-216' for key 'PRIMARY' (PDOException)`
This test completes successfully locally

Issue #2151, opened 2017-09-28
Print (Spike) Investigate whether images can NOT be cut off when printing


alexiscott:

alexiscott:
Description
===
This relates to 2142, but it is more specific.

iris-i:
Ok, this may have been the longest QA process in the history of QA's but I could not see this fix working as indicated. I tested it in dkan locally, then I tested it in my local ga_gbpw sandbox, but could not get the print styles to prevent the pie charts from being cut off on top. 
@alexiscott could you add some extra QA steps on how you tested this in Georgia or how wlse Ic an see this fix working? My testing steps may be incorrect.

I have attached PDF files of my local georgia dashboard with and without the fix in place. They are identical.
[local-georgia--with-css-fix.pdf](https://github.com/NuCivic/dkan/files/1345807/local-georgia--with-css-fix.pdf)
[local-georgia-without-css-fix.pdf](https://github.com/NuCivic/dkan/files/1345806/local-georgia-without-css-fix.pdf)

alexiscott:
Hi Iris,
Thanks and sorry; I know it's tricky to test because the problem is not reproducible on dkan local.
What I was doing was to copy the compiled (you need to run gulp) output of print.css into the browser source window, and then pull up the print preview window.

I also thing that there was a problem with the CSS, so I have changed it to this:

@media print {
  svg {
     -webkit-column-break-inside: avoid;
      page-break-inside: avoid;
      break-inside: avoid;
    }
}

Testing on https://gbpw.georgia.gov/physician-workforce-data

iris-i:
Ah I see, thanks Alex. I'll give this another go.

iris-i:
Awesome! This fix works as intended. Images no longer get cut off when printing. 
🎉 
[pdf-fix.pdf](https://github.com/NuCivic/dkan/files/1354397/pdf-fix.pdf)

PR #2157, opened 2017-09-29
Print do not cut off images

alexiscott:
Connects #2151 

## Description
 Adding print styling so that images, table rows, and charts work with page breaks.


iris-i:
@alexiscott this works as designed. Could you resolve the merge conflicts so we can merge it in? Thanks!

alexiscott:
Attempted merge resolution by rebasing out css.map files which should in principal not need to have been committed.

PR #2161, opened 2017-09-30
Truncate datastore table on re-import.

dkinzer:
Description
===
The fast dkan_datastore_fast_import module does not truncate data when doing multiple
imports of the same data.  This has the effect of gradually ballooning the size
of datastore DB tables under certain circumstances.

One particular scenario where this happens is with queued datastore_fast_import
imports. These imports are allowed 600 seconds and in cases where they do not
they will simply stop and then set to restart. 

Of course anything that prevents the queue from resolving can cause it to
reimport until it does resolve.

The regular dkan_datastore module does not exhibit this problem because feeds
automatically handles this problem correctly.

This PR makes the following changes to help with this issue:

* It adds unit tests that specifically test queuing multiple imports of the same data.
* It changes the dkan_datastore_fast_import queues so that they can be triggered
  by a regular cron_job call (this was previously disabled)
* It adds a truncate pre step to imports to resolve the issue.

This PR does not attempt to refactor the code as in #2122 as this refactoring is now
being worked on more comprehensively in another PR (#2146).

QA Steps
==
- [ ] Verify that the new phpunit tests meet the criteria of testing a regression of this issue.

### DIFF
--- a/.ahoy/docker-compose.yml
+++ b/.ahoy/docker-compose.yml
@@ -30,7 +30,7 @@ web:
 # DB node
 db:
   hostname: db
-  image: blinkreaction/mysql:5.5
+  image: blinkreaction/mysql:5.6
   ports:
     - "3306"
   environment:
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,5 +1,6 @@
 7.x-1.14
 --------
+ - #2161 Fix Datastore Fast Import does not truncate old data on multiple imports.
  - #2075 DKAN Harvest: Modified batch process to split harvest migrations in chunks.
  - #2145 Front page hero image no longer requests empty URL if variable empty.
  - #2082 Add UI for generating ODSM file caches.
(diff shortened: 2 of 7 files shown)


fmizzell:
I code reviewed (very simple straight forward code) and the started trying to replicate the issue. 
I did not succeed at replicating the issue. I managed to get the import stuck plenty of times by leaving while the batch was running (I also never managed to trigger the queuing mechanism offered by the fast importer from the UI), but the import continued successfully in the background without issue, and the UI would not allow me to queue things again while the import is going. Anyways, we need better, more specific, steps on how to reproduce this issue before we can say that it has been fixed.

dkinzer:
@fmizzell it's difficult to reproduce this issue, that is why I created the unit test so that you can see the failure:

```
There was 1 failure:

1) DkanDatastoreFastImportQueueTest::testMultipleQueueImportSingleDatastore
Failed asserting that 2969 matches expected 8907.
``` 

https://circleci.com/gh/NuCivic/dkan/8422

To reproduce one would have to create a scenario that invokes a queue and breaks while importing (but I think that is easier said than done).

The phpunit test is conceptually the same idea.  So I think it shows it fixes the issue.

Issue #2177, opened 2017-10-12
Move DKAN repos to GetDKAN GitHub org


grugnog:
Github will redirect SSH/HTTP, but it would be good to try and make the change as comprehensively as possible, so we don't need to keep coming back.

Non-exhaustive list of prep-tasks:
- [x] Prepare PRs for DKAN to change repo URL in docs, CI, etc
- [x] Prepare PRs for other repos that reference DKAN (e.g. getdkan.org repo)
- [x] Identify places in Jenkins and other config/scripts that reference the repo, list the files/URLS below and prep (but don't apply) config changes as needed
- [ ] Identify integrations (CircleCI, ProboCI, Slack, Waffle etc) that will need updating/fixing below
- [x] Set up GitHub team on GetDKAN with access to repo

Move repository!

Follow up tasks:
- [x] Update/fix CircleCI
- [ ] Update/fix ProboCI
- [x] Update/fix Slack integration(s)
- [x] Update/fix Waffle integration
- [ ] Apply Jenkins config changes

dafeder:
Waffle seems to have updated itself :)

PR #2188, opened 2017-10-19
Move to GetDkan org

dafeder:
Connects #2177

Preparing references to NuCivic org in Github to change to GetDKAN.


janette:
Tests failed on known glitch (unexpected alert open), merging it.

PR #2184, opened 2017-10-18
Allow users to change Groups on standalone resources

fmizzell:
Issue: CIVIC-5087

## Description

Users should be able to add or remove groups on standalone _resources_ (_resources_ that do not form part of a _dataset_ node). 

Also, the groups field had some style errors so this PR includes some CSS fixes for that.

## QA Steps

- [x] Create a standalone resource.
- [x] Confirm that you are able to assign the resource to a Group.
- [x] Edit a resource that is part of a dataset. Confirm that you do not see the groups field.

## Reminders
- [x] There is test for the issue.
- [x] CHANGELOG updated.
- [x] Coding standards checked.
- [ ] Review docs.getdkan.com (or in /docs) to see if it still covers the scope of the PR and update if needed.

PR #2208, opened 2017-10-31
Fix access to groups field

janette:
Follow up fix for #2184 

When creating a dataset the user will see the groups field on 'step 2', this should be hidden.

## QA Steps

- [x] Create a standalone resource.
- [x] Confirm that you are able to assign the resource to a Group.
- [x] Edit a resource that is part of a dataset. Confirm that you do not see the groups field.
- [ ] Create a dataset, on step 2 _Add a Resource_, confirm you do not see the groups field

## Reminders

- [x] CHANGELOG updated.


janette:
Tests passed before fix to changelog conflict, merging.

PR #2215, opened 2017-11-03
Add new module Token Tweaks.

eknapier:
* Update CHANGELOG.txt

* Update dkan.info and drupal-org.make

## User story

As a user  I want to be able to edit an ODSM API endpoint without experiencing memory issues or timeouts. 

## How to reproduce

To reproduce, edit any ODSM API endpoint and observe memory exhaustion issues, or extremely long load time. See this separate PR for more details on another user's experience: https://github.com/GetDKAN/open_data_schema_map/pull/102

## QA Steps

- [x] edit any ODSM API endpoint and confirm no memory issues persist and that page loads quickly.
- [x] visit admin/config/system/tokens to ensure UI for setting token recursion level is present and module installed correctly. (a lower recursion level can be set here if a lower setting is needed to prevent memory issues when editing any ODSM API endpoint.


kducharm:
@eknapier I couldn't get past memory errors on Probo.ci with any of the token levels, I attempted to see if a .probo.yml setting would try to bump up the memory_limit to 512m but it didn't work. Were you able to get this to work in a 256m limit or only in sandbox, which I think is 512m?

PR #2226, opened 2017-11-08
Allows the services API to be able to consume datasets with the same …

### LINKED ISSUE #2196: The services API can not consume what it puts out
If you retrieve a dataset with the API, and then try to save it as a new dataset, the API will not be able to save it. Most of the errors are related to reference fields. 

In particular the dataset reference field in a resource is validated by the API with an entity_reference_autocomplete validator which expects the title of a node instead of a node id. Other fields may have the same problem.

fmizzell:
Fixes #2196


modules/dkan/dkan_datastore/modules/dkan_datastore_api/dkan_datastore_api.module
```diff
@@ -1123,3 +1123,180 @@ function dkan_datastore_api_debug($data_select) {
...
+    $options = _options_get_options($field, $field_instance, $properties, 'node', $node);
+  }
+
+  return $options;
+}
+
+/**
+ * Check field values against field options.
+ */
+function _dkan_datastore_api_field_validation_check_field_values_against_options($field_name, $field_values, $options) {
```
dafeder: Does not check if cardinality is correct - could have submitted multiple values for a single value field. What would happen in that case?

modules/dkan/dkan_datastore/modules/dkan_datastore_api/dkan_datastore_api.module
```diff
@@ -1123,3 +1123,180 @@ function dkan_datastore_api_debug($data_select) {
...
+
+      $nid = $old_node->nid;
+      $response = array('nid' => $nid);
+      if ($uri = services_resource_uri(array('node', $nid))) {
+        $response['uri'] = $uri;
+      }
+      return $response;
+    }
+    else {
+      return _node_resource_update($nid, $request);
```
dafeder: What function is this? `_node_resource_update()`

modules/dkan/dkan_datastore/modules/dkan_datastore_api/dkan_datastore_api.module
```diff
@@ -1123,3 +1123,180 @@ function dkan_datastore_api_debug($data_select) {
...
+  if (isset($resources['node']['operations']['update'])) {
+    $resources['node']['operations']['update']['callback'] =
+      "_dkan_datastore_api_resource_update";
+  }
+}
+
+/**
+ * Services operation callback: node creation.
+ */
+function _dkan_datastore_api_resource_create($request) {
```
dafeder: Does not check access control

modules/dkan/dkan_datastore/modules/dkan_datastore_api/dkan_datastore_api.module
```diff
@@ -1123,3 +1123,180 @@ function dkan_datastore_api_debug($data_select) {
...
+    return services_error("Type not set for node");
+  }
+
+  return $response;
+}
+
+/**
+ * Services operation callback: node update.
+ */
+function _dkan_datastore_api_resource_update($nid, $request) {
```
dafeder: No access control? Also does this create a new revision?

modules/dkan/dkan_datastore/modules/dkan_datastore_api/dkan_datastore_api.module
fmizzell: I believe the services module handles authentication and access, but we can check.

modules/dkan/dkan_datastore/modules/dkan_datastore_api/dkan_datastore_api.module
fmizzell: The code is working with the entity api directly so it will do whatever it does. I believe if no vid is set it will create a new version. I am not sure what the behavior is if a vid is set.

modules/dkan/dkan_datastore/modules/dkan_datastore_api/dkan_datastore_api.module
fmizzell: This function is the original callback for all node related service calls, since we are handling only a few node types we want to pass the rest of the calls to services.

modules/dkan/dkan_datastore/modules/dkan_datastore_api/dkan_datastore_api.module
fmizzell: Good question. My guess is that it will still work, but I think that checking for cardinality is a good check.

janette:
Lets incorporate the test changes from https://github.com/GetDKAN/dkan/pull/2375

modules/dkan/dkan_dataset/modules/dkan_dataset_rest_api/dkan_dataset_rest_api.module
```diff
@@ -35,3 +35,184 @@ function dkan_dataset_rest_api_services_request_postprocess_alter($controller, $
...
+ * Services operation callback: node update.
+ */
+function _dkan_dataset_rest_api_resource_update($nid, $request) {
+  $node = node_load($nid);
+
+  if (isset($node->type)) {
+    if ($node->type == "dataset" || $node->type == "resource") {
+
+      if (empty($node->nid)) {
+        return services_error(t('Node @nid not found', array('@nid' => $node->nid)), 404);
```
dafeder: Seems like a strange error to throw if the nid is empty - it's just always going to read "Node  not found"

modules/dkan/dkan_dataset/modules/dkan_dataset_rest_api/dkan_dataset_rest_api.module
```diff
@@ -35,3 +35,184 @@ function dkan_dataset_rest_api_services_request_postprocess_alter($controller, $
     }
   }
 }
+
+/**
+ * Implements hook_services_resources_alter().
```
dafeder: I think we should include some kind of comments explaining how and why we're changing the service callback

fmizzell:
Got into rebase hell. I will create a new clean PR

PR #2229, opened 2017-11-08
Strengthening Datastore Behat Tests (#2183)

### LINKED ISSUE #2183: Behat Test Revivification
Several weaknesses in the current DKAN Behat suite should be addressed to ensure the most streamlined workflow, and to avoid false flag failures. Some specific strategies we've identified that would help towards this end:

- Adding additional cleanup methods to address problem areas in testing suite (i.e., adding call to clear cache in a post scenario hook targeting a particular behat tag. )
- Remove unnecessary, superfluous lines from Behat tests that do not explicitly test the scenario description. (i.e., see several instances of "And I wait for "No imported items."")
- Consider replacing final "THEN" steps reliant upon visible text strings with lines mapping to API level methods, to add clarity to Gherkin statements and for less fragility. Unless the scenario description clearly states that a purpose of the test is to verify the presence of a specific text string, then an API method may be preferred for testing the final outcome. 
- Consider deprecating some UI tests in favor of API / "service level" tests where such an approach is logical. For example, perhaps it is useful to have 1 UI test to fill out a certain node form, but multiple tests tracing the same UI steps is ultimately redundant, even if each is testing different outcomes. In these instances some UI tests could be deprecated in favor of tests that may not need a browser at all to verify, but rather just a few calls to DKAN/Drupal.

One important goal of this work: Allowing for all tests to run (in any order) with all DKAN modules enabled, including the DKAN Workflow module. This module is disabled currently because of Behat test failures and it is an ultimate goal for us to re-enable the DKAN Workflow module so that all tests in the full suite can run in any order or combination without unexpected test failures.

eknapier:
* Fixed bug with dkan_datastore_records() not returning FALSE

    * Added new Behat step to DatastoreContext.php for calling dkan_datastore_records()

    * Modified Behat tests to use this new step in favor of searching for text strings to determine status of datastore

This PR addresses some issues raised in #2183. The modifications to resource behat features will hopefully address the recurring random failures with these tests.

No QA Steps other than to confirm that tests are passing in CI.

## User story

Fixes #2183

### DIFF
--- a/modules/dkan/dkan_datastore/dkan_datastore.module
+++ b/modules/dkan/dkan_datastore/dkan_datastore.module
@@ -391,7 +391,8 @@ function dkan_datastore_records($nid) {
         $query = db_select($table, 't')
           ->fields('t', array('timestamp'))
           ->range(0, 1);
-        if ($result = $query->execute()) {
+        $result = $query->execute();
+        if (!empty($result->rowCount())) {
           return TRUE;
         }
       }
--- a/test/dkanextension/src/Drupal/DKANExtension/Context/DatastoreContext.php
+++ b/test/dkanextension/src/Drupal/DKANExtension/Context/DatastoreContext.php
@@ -105,6 +105,28 @@ class DatastoreContext extends RawDKANContext {
   }
 
   /**
+   * @Then :title should have no datastore records
+   */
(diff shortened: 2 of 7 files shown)


janette:
YES!

PR #2237, opened 2017-11-14
Migration of CircleCI from V1 to V2

kducharm:
Migrates CircleCI from V1 to V2, using docker containers to run automated tests

## QA Steps
- [ ] Verify all behat tests are running and passing, and that all features are accounted for in parallel tasks
- [ ] Verify unit tests are passing
- [ ] Verify linting is passing


kducharm:
Some behat tests are not passing, but I don't believe that is a result of the migration.

Pythagory:
This looks good to me. I agree with the removal of that "LOCAL DATA" test, it seemed like a strange thing to test in Behat and really environment-dependent.

.circleci/config.yml
```diff
@@ -0,0 +1,140 @@
...
+      # Lint only on 1st parallel instance. Deploy: blocks until parallelism is done, so doing it this way.
+      - run:
+          name: Run Lint
+          command: |
+            if [ "$CIRCLE_NODE_INDEX" -eq "0" ]; then
+              ahoy dkan lint
+            fi
+      # Unit Tests only on 1st parallel instance. Deploy: blocks until parallelism is done, so doing it this way.
+      - run:
+          name: Run Unit Tests
```
dafeder: Sorry to nitpick but can we call these PHPUnit Tests? They're not really "unit tests" - they're functional tests in PHPUnit.

CHANGELOG.txt
```diff
@@ -1,6 +1,7 @@
 7.x-1.14.1
 ----------
  - #2103 Fixed field mapping defaults for open data schema APIs
+ - Migration of CircleCI from V1 to V2
```
dafeder: Add the PR number (we will automate this some day)

.circleci/config.yml
kducharm: Done - ahoy dkan unittest should probably be changed, but not today :)

CHANGELOG.txt
kducharm: Done

PR #2249, opened 2017-11-18
Change timezone handling to none on temporal coverage field

janette:
Temporal coverage dates display in the sites timezone. This leads to differences between harvested temporal coverage values and the data.json provided by DKAN.

## QA
- Confirm that the temporal coverage dates listed in data.json are using UTC and not adjusted to site timezone adjustments.

## Reminders

- [ ] CHANGELOG updated.


kducharm:
@dafeder I had some time so I CR and QA'd this w/Janette, needed some manual steps to create a harvest and re-harvest after changing timezones. Confirmed data.json didn't change after deleting dataset, changing timezone, and re-harvesting and that data.json was not cached.

PR #2265, opened 2017-12-01
Fix hero background image setup

janette:
Previous fix to the hero panel did not take into account older sites may be using the older front page template. https://github.com/GetDKAN/dkan/pull/2145

Also adjusting the margin/padding of the hero pane so content is vertically centered.

## QA Steps

This line fixes the image path variable for sites using the older panel template https://github.com/GetDKAN/dkan/pull/2265/files#diff-3c8bfae081ff80bac6eae071d046a96fL11
Everything else is css padding/margin fixes to keep the hero content vertically centered.

- [x] Add a background image at `admin/appearance/settings/nuboot_radix` and save
- [x] Confirm the image appears in the front page hero region.
- [ ] Change the layout to use the older custom front layout option. 
- [ ] Confirm the hero image loads as the background image.

## Reminders

- [ ] CHANGELOG updated.


janette:
Note: VA and ND will need this as a patch for the 1.14 upgrade

kimwdavidson:
QA this on ND or VA

PR #2268, opened 2017-12-04
Refactor dataset.author.feature and resource.author.feature

janette:
The backgrounds used on these two files are unnecessarily large, many scenarios do not need any of the set up content and the time it takes to create and tear down all of this test content takes longer than it takes to run the test. This causes SQLSTATE[23000]: Integrity constraint violation: duplicate entry errors on CircleCi

**dataset.author.feature background**
- 9 users
- 4 groups
- 2 tags
- 9 datasets
- 6 resources

**resource.author.feature background**
- 7 users
- 3 groups
- 2 tags
- 6 datasets
- 15 resources!!

Rather than creating all of this content, lets just create the minimum amount per scenario rather than putting everything in the background.

### DIFF
--- a/test/features/resource.author.feature
+++ b/test/features/resource.author.feature
@@ -1,65 +1,17 @@
-# time:5m47.66s
+# time:4m30.30s
 @api @disablecaptcha
 Feature: Resource
 
   Background:
     Given pages:
-      | name          | url                |
-      | Content       | /node/add          |
-      | User          | /user              |
+      | name          | url           |
+      | Content       | /node/add     |
+      | User          | /user         |
     Given users:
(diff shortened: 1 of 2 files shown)


test/features/resource.author.feature
```diff
@@ -133,228 +103,359 @@ Feature: Resource
 
   @resource_author_07 @noworkflow
   Scenario: Add a resource with no datasets to a dataset with no resource
+    Given groups:
```
fmizzell: @janette Do we need to create a group for this test?

janette: @fmizzell this allows the test to pass on sites that require a group for datasets

Issue #2027, opened 2017-07-24
Harvester should use file format as resource title when no title is provided


jvitello:
## User story
As a Site Manager, when I harvest data from a source that does not provide titles for resources/distributions, I should see the format of the file used as the title, rather than the URL, so that it is easier to read.

## Original Request

Datasets being harvested are being titled with relative link titles, which do not accurately reflect their content, request research into possible alternatives.

## Description

We think the client is referencing the behavior where if no title exists for a distribution in a data.json harvest, the url is used as the title in DKAN. We'd like to use the filename instead, or use the file type as the title if no file name can be isolated.

POD schema for distributions:
https://project-open-data.cio.gov/v1.1/schema/#distribution

## Steps to Reproduce

- create a resource with the attached data.json
- harvest using the resource download link
- review the resource titles

## Acceptance Criteria

- If a data.json distribution does not include a title, use the file name as the title
- If a title can't be determined use the file format as the title "{{file type}} File"

## Tests

We should add php unit tests to make sure all of this is happening correctly.

kimwdavidson:
Blocked by PHP unit tests

fmizzell:
The issues with the unit tests will be dealt in #2301

PR #2284, opened 2017-12-15
Change the default title for resources to use the format rather than url

janette:
Connects #2027

Harvested resources that do not have title values are using the url as the title. Switching to use the format value instead.

## User story

When I harvest data from a socrata source, I don't want to see long urls used as the resource titles, very confusing.

## How to reproduce

1. Create a harvest source `node/add/harvest-source` with https://data.medicare.gov/data.json
2. Filter by `keyword` = `doctor`
3. Harvest, notice that the resources have titles like **https://data.medicare.gov/api/views/hanv-ru8h/rows.csv?accessType=DOWNLOAD**

## QA Steps

1. Create a harvest source `node/add/harvest-source` with https://data.medicare.gov/data.json
2. Filter by `keyword` = `doctor`
3. Harvest, notice that the resources have titles that use the format value as the title

## Deployment Steps

Existing harvests will only update if the hash changes, you can force an update with `drush dkan-hm [source_machine_name] --skiphash`

## Reminders

- [ ] CHANGELOG updated.


janette:
@dafeder @eknapier 'All checks have passed' is not entirely true, under phpunit tests is
```
OK, but incomplete, skipped, or risky tests!
Tests: 88, Assertions: 214, Skipped: 7, Incomplete: 1
```

dafeder:
Your test isn't being skipped though @janette , so not your problem :)

Will merge after branch updates and tests re-run

PR #2294, opened 2017-12-21
Point to 1499-remove-iframe-preview branch of recline

dafeder:
QA only. GetDKAN/recline#98


dafeder:
Let's try this on 1.15

PR #2298, opened 2017-12-27
Modifies curl settings so url headers and info can be properly acquir…

fmizzell:
When trying to harvest some sources, some of the resources are incorrectly typed because we are not getting proper information from the urls. This PR fixes that issue.

An example harvestable source that currently miscategorizes resources is "https://data.chhs.ca.gov/data.json"


janette:
@fmizzell this phpunit test fails with these changes `dkan/test/phpunit/dkan_dataset/getRemoteFileInfoTest.php`

janette:
@fmizzell if this is merged into develop then it will be released in 1.15, is that right?

janette:
This is a bug fix so changing the base to 7.x-1.x

Issue #2250, opened 2017-11-20
Task: Implement new branching strategy


kducharm:
### Most Important
- [x] Update zip and tar.gz for1.14.1
- [x] Release 1.14.2
- [x] Create a script that adds version information to the appropriate .info files (Gerardo has a script that removes the version info. Modifying that to add info should be simple)
- [x] Create a packaging script. For each release we should create a packaged version of DKAN with updates to the CHANGELOG, and version information
- [ ] Identify other necessary changes to point release workflows and possible automation
- [ ] Update DKAN docs to explain all of this
- [x] Create scripts that modifies the CHANGELOG from git information.

### Good to Have
- [ ] Revise/create a new script for patch releases (1. Merge develop into the production branch 2. Creates an integration branch from production 3. Creates a PR from the integration branch to be merged into develop)
- [ ] Enhance the package manager to not only add information but to create a fully functional DKAN with drupal (just like dkan_starter).  
- [ ] Connect all this scripting with more automation, either on a github hook, or through jenkins

### Done
- [x] Create a new `develop` branch on DKAN. This will now be the default branch in the repo.
- [x] Write ticket for point release automation 
- [x] Create 2.x releases for recline, open_data_schema_map, visualization_entity
- [x] PR/release of DKAN switching to fixed versions of EVERYTHING, no more branches in makefiles

kimwdavidson:
Dan to rewrite this as implementing new branching strategy

fmizzell:
I do not have the right permission to set default branches and other setting on the repo \cc @dafeder

dafeder:
Try now

fmizzell:
Update: The release script can now freeze the develop branch, and the release method can merge that branch into "master" (7.x-1.x) and develop. The only thing left is to create the tag on "master". There are quite a few extra things happening in that release method that I would like to look at and discuss with @dafeder before calling it done.

acouch:
Closing this ticket as it has not been updated recently and is not on our near-term roadmap. Please re-open if you would like to discuss further or are actively working on it.

PR #2341, opened 2018-01-23
Modifies curl settings so url headers and info can be properly acquired with curl

janette:
re-roll of #2298
Connects nucivic/cadepttech#85

When trying to harvest some sources, some of the resources are incorrectly typed because we are not getting proper information from the urls. This PR fixes that issue.

An example harvestable source that currently miscategorizes resources is "https://data.chhs.ca.gov/data.json"
## QA Steps

- [ ] harvest from the url above
- [ ] confirm the resource files are entered into the remote file field and not the api field


janette:
@fmizzell 
test harvest source on probo https://ba15eb9d-e36c-48e4-821f-df693ed795d7--pr-2341.probo.build/harvest_source/test

kimwdavidson:
Waiting on improvements in #2250

fmizzell:
@janette got some test failures after updating the branch.

PR #2358, opened 2018-02-07
Add try/catch on datastore.inc to avoid cron backlog

janette:
Connects https://github.com/NuCivic/dkan_management/issues/40

## QA Steps

- [ ]

### DIFF
--- a/modules/dkan/dkan_datastore/includes/Datastore.inc
+++ b/modules/dkan/dkan_datastore/includes/Datastore.inc
@@ -40,7 +40,14 @@ abstract class Datastore {
     if (!isset($node[$uuid])) {
       $nid = $this->getNid($uuid);
       if (!$nid) {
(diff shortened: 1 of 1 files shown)


modules/dkan/dkan_datastore/includes/Datastore.inc
```diff
@@ -40,7 +40,14 @@ abstract class Datastore {
     if (!isset($node[$uuid])) {
       $nid = $this->getNid($uuid);
       if (!$nid) {
-        throw new Exception(t("uuid !uuid not found.", array('!uuid' => $uuid)));
+        try {
+          if (!$nid) {
+            throw new Exception(t("uuid !uuid not found.", array('!uuid' => $uuid)));
```
fmizzell: @janette If we are throwing an exception and immediately catching it, I think we can remove the exception and simply add the watchdog_statement. Thoughts?

PR #2405, opened 2018-02-26
Adjust group.editor.feature so we only create content when needed

janette:
Remove unnecessary creation of resources on scenarios that don't need them. Lets only create resources for the scenarios that need them.

circleci


fmizzell:
@janette The last time I was dealing with constraint violations was due to a silent failure before the failing test. Did you check on that? It would be odd on a resource due to the primary id being an nid which is an auto-increment, but still worth looking into.

janette:
Yes this is side stepping the real issue but does get us to green. I found this article which seemed worth looking into when we have some devops time https://knpuniversity.com/screencast/behat/clear-data-symfony-extension#clearing-the-database-before-each-scenario

janette:
I think this is still an improvement as we avoid creating an unnecessary resource on 12 scenarios.

PR #2409, opened 2018-02-28
 Allows the services API to be able to consume datasets with the same…

### LINKED ISSUE #2196: The services API can not consume what it puts out

fmizzell:
Fixes #2196


dafeder:
@fmizzell can you provide some acceptance tests for this? I did try to create a dataset and got back "406 Not Acceptanble: Node type is required". Here's a full CURL command to reproduce what I did:

https://gist.github.com/dafeder/55e8566778252dc43055f2339762b545

PR #2412, opened 2018-03-01
Installation instructions for ahoy

dafeder:
There is still a lot to be done to clean up the documentation of installation and dev practices, but this is a start looking toward open data day and all the participation we're expecting.


dafeder:
@janette I agree, but I'm going to ask if you can merge anyway and save this for another fix/PR. For Data Day tomorrow, the important thing is to have a single page we can point people to to get their environments set up. I'll try to remove the redundant info in another PR soon, or submit a PR yourself if you get to it first.

PR #2419, opened 2018-03-07
Decouple search from panels, other features

dafeder:
Currently, the different pieces of the default DKAN search page - the datasets search index, the db search server, the main search view and the search (datasets) page definition including panel configuration, are all in either dkan_sitewide_search_db or dkan_sitewide_panels. This PR separates the index definition, view and panel into a separate module (dkan_dataset_search) which can be disabled and completely replaced.

Before releasing we need to thoroughly test and figure out what upgrade path if any is needed for sites currently using the datasets index.

## Acceptance Criteria
- Nothing breaks on sites using dataset index after upgrade
- Dataset index can be disabled, and a new index enabled without losing functionality


nucivicmachine:
Closing in favor of #2733

Issue #2414, opened 2018-03-07
Discovery: file storage via S3


erogray:
Some clients need to store large amounts of files, which is impractical using current storage scheme. 
https://docs.acquia.com/acquia-cloud/manage/files/about
https://docs.acquia.com/article/using-external-storage-files

We should look into decoupling files storage, such that large files can be kept on S3. 

https://www.drupal.org/project/s3fs
https://www.drupal.org/project/s3fs_cors
https://www.drupal.org/project/amazons3

fmizzell:
@erogray Given that #2428 provides a lowly-integrated yet workable solution to use S3 as a storage mechanism to keep large files, do we still need this open?

erogray:
Closing in favor of https://github.com/GetDKAN/dkan/issues/2179 and https://github.com/GetDKAN/dkan/pull/2428 - Discovery seems complete.

PR #2428, opened 2018-03-19
Reliable datastore

fmizzell:


erogray:
connects https://github.com/GetDKAN/dkan/issues/2414

Issue #2429, opened 2018-03-19
Collections functionality not appearing


kimwdavidson:
And See All link appears on every node

ethanteague:
Placing here as a reminder: we need to move Collections facet from dkan_dataset_collections.info to dkan_sci_search.info.

acouch:
Closing this ticket as it has not been updated recently and is not on our near-term roadmap. Please re-open if you would like to discuss further or are actively working on it.

PR #2432, opened 2018-03-22
Dataset collections syncing with USDA - no new functionality

ethanteague:
Connects #2429 
## QA
- [ ] Add a dataset parent collection item to a dataset via the collection field, and make sure you save the node not as a draft, but as Published.
- [ ] Confirm the parent collection item appears in the child collection table
- [ ] Follow the link, confirm it appears on parent, and see all link takes you to search page with parent item selected as collection facet with only collection results being returned.


dafeder:
Closing, now lives in https://github.com/GetDKAN/dkan_dataset_collections

PR #2437, opened 2018-03-29
Patch paragraphs to validate for required fields.

dafeder:
Connects isdapps/adc-code#1798

Required fields on paragraphs (for extended metadata) do not validate as required when submitting a node form. This patch fixes that.

## QA Tests

* Create a paragraph bundle with a required field
* Create a dataset and attach your new paragraph bundle
* Try to submit the node form with your required paragraph field blank


kimwdavidson:
@dafeder Is this for code review?

Issue #2451, opened 2018-04-10
Home link path doubled when site installed in directory


leeomara:
## Description

When DKAN is installed in a sub directory of a webroot (e.g. `example.com/data`), the home link in the header has the path doubled (e.g. `/data/data`)

I am using DKAN 7.x-1.14.5.

## Steps to Reproduce

- Place DKAN in webroot subdirectory (mine is in `data`)
- Update `$base_url` in `settings.php` to match (e.g. `http://127.0.0:8888/data`)
- Visit the home page (or any other page)
- Notice the URL of the home link has the path doubled (e.g. `/data/data`)

## Acceptance Criteria

The home link should be the site root.

## Test Updates

None.

## Documenation Updates

None.

kimwdavidson:
@leeomara Would you mind rebasing your branch against the 7.x-1.x branch? The tests were broken you first pushed this. Thanks!

erogray:
Thanks @leeomara !

PR #2452, opened 2018-04-10
Fix home link path-doubling when site in directory

leeomara:
Do not wrap $front_page in url()

connects https://github.com/GetDKAN/dkan/issues/2451

This change fixes an issue (link above) that we ran into when DKAN was installed in a subdirectory of the webroot.


leeomara:
Hi @janette, for sure next time I'll use a different branch name. Sorry for not seeing the notice sooner.

Cheers!

PR #2453, opened 2018-04-11
Update linting script

dafeder:
This updates the linting script in ahoy to use ruby instead of bash, and be smarter about getting the file list from git.

## QA Tests

* The PR initially has a broken version of dkan_datastore.module, which should break the linting test. Once confirmed, revert that change and merge.


dafeder:
Closing in favor of #2454

PR #2458, opened 2018-04-14
Fix the chrome issue

fmizzell:


janette:
**@workflow_23**
Lets add a @fixme tag here as this is an issue with the repeated enable/disable of dkan_workflow, there is a permission being added to anon users on enable that seems to not happen for tests but if you build a site, enable dkan_workfow, everything is working as expected. We need a better test process.

PR #2464, opened 2018-04-18
Move functions out of .install

janette:
We need to call the `_dkan_sitewide_menu_setup_admin_menu_source()` when implementing custom permissions as it will also include menu links to additional admin pages. Currently the function is in the .install file, so `function_exists('_dkan_sitewide_menu_setup_admin_menu_source')` returns false, we can use `module_load_install()` or move functions out of the install file.
connects https://github.com/NuCivic/ga_reports/issues/448


janette:
@kimwdavidson it is in the code review lane and assigned to Aaron, is there something missing to indicate it is ready?

acouch:
Looks good to me.

PR #2470, opened 2018-04-23
Get harvest tests passing

dafeder:
Fixing some broken tests post-circle container changes. Despite valiant efforts the have to just be disabled for now. The harvest feature should pass.


dafeder:
@acouch @janette looks like the harvest tests are no longer breaking.

PR #2489, opened 2018-05-03
Allows the user to change the map-tiles url used for the 'Data Extent…

### LINKED ISSUE #2421: Data extent block is not using the same tile layer as the map visualization
We noticed that the data extent block is using the https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png tile layer instead of the stamen tiles (https://stamen-tiles-{s}.a.ssl.fastly.net/terrain/{z}/{x}/{y}.png) that are used widely in DKAN. For consistency purposes, it would be great if this was corrected, or added as configuration.

fmizzell:
Fixes #2421

### DIFF
--- a/docs/components/dataset/datasetfeatures.rst
+++ b/docs/components/dataset/datasetfeatures.rst
@@ -41,6 +41,8 @@ DKAN provides a widget so that a spatial area can be drawn if desired:
 
 .. figure:: ../../images/spacial-geographical.png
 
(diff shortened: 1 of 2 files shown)


chrodos:
@fmizzell I know I am late on that, but there should also be a variable for the attribution (e.g., $map_tile_url_attribution). If a user changes the URL, the attribution text should also change.

PR #2492, opened 2018-05-04
Moving the file_resup module to version 1.5.

### LINKED ISSUE #2491: Security update required for file_resup
Looks like the file_resup folks have pushed a security fix for a problem similar to SA-CORE-2018-004 into yesterday's 7.x-1.5 update.

That's a highly critical issue. What will it take to get this fix into the dkan distro?

https://www.drupal.org/project/file_resup/releases/7.x-1.5

> This release contains a fix similar to SA-CORE-2018-004.

fmizzell:
Fixes #2491

### DIFF
--- a/drupal-org.make
+++ b/drupal-org.make
@@ -146,7 +146,7 @@ projects:
     download:
       type: git
       url: https://git.drupal.org/project/file_resup.git
-      revision: 6cf030c2c139374c75527221775965ff0242df15
+      revision: fd5aad6bd26ca84303bb07c3f757cd24a5cb5c01
   filefield_sources:
     version: '1.11'
   font_icon_select:


elstudio:
Thanks for getting to this so quickly!

One question, though -- 

Is there a good reason to specify a specific git commit hash in the make file rather than simply version 1.5?

Using the plain version number makes Drupal's updates screen much clearer -- that way we all know this is the plain-vanilla file_resup-7.x-1.5 from HEAD rather than some special version.

(Had we been on v1.4 previously, I would have submitted a pull request and spared you a little bit of work.)

fmizzell:
@elstudio That is a good question. I will ask around, and if there isn't a particular reason, we will do exactly what you suggest. Thanks.

PR #2496, opened 2018-05-09
Control datastore via REST API

dafeder:
This PR adds the ability to configure, import, and drop a datastore from the Dataset REST API

Our main services resource is "datastore". I takes the nid of a resource as an argument. The full API should look something like this:

/datastore/: GET - Gives configuration information about the datastore for that resource.
/datastore/: POST - Configures the datastore.
/datastore/: DELETE - Drops the datastore.
/datastore//import: ? - Action to do and import or just to schedule and import?.


dafeder:
Superceded by #2519, closing

PR #2508, opened 2018-05-12
Update dkan.install

ethanteague:
Checks for presence of field_data_field_language and field_data_revision_language in update 7018 prior to table update.


dafeder:
We should bring back the normal array() syntax for consistency and to not introduce unnecessary php5.3-breaking syntax. not sure this is really still needed?

PR #2517, opened 2018-05-14
Allows modules to alter the format of a resource.

fmizzell:


erogray:
@fmizzell should this pr be attached to an issue? and can/should it have QA steps?

dafeder:
We're not going to do this in core. This maybe useful for someone trying to do this in the future, but closing.

PR #2518, opened 2018-05-16
Check for date object item before assigning to variable

ethanteague:
Multiple notices being thrown when date time formatting object items aren't all there. This checks for presence of multiple values before attempting to assign to variables, and prevents cascade of notices. Issue: https://www.drupal.org/project/date/issues/2973379
Patch: https://www.drupal.org/files/issues/2018-05-16/date-empty-val-checks.patch


kimwdavidson:
@fmizzell Do we need this for World Bank? Should this be in code review?

erogray:
Need more info about problem; closing unless this comes up in a new instance

Issue #2495, opened 2018-05-08
Dataset Modified Date not changing on edit dataset


chrodos:
When a dataset is edited the Modified Date  does not change. I tried changing various fields values in the dataset form and adding new resources. The modified date remains the same.

janette:
@chrodos what version of dkan are you using? this is the only thing I can find in recent issues https://github.com/GetDKAN/dkan/issues/1905

chrodos:
@janette Its DKAN (dkan-7.x-1.14.5) from the drops-7 package.

janette:
I have the 7.x-1.14.7 drops-7 locally, (only minor changes from 1.14.5), dkan_workflow enabled, and I'm getting the expected updated modified date when editing datasets...
Try adding some debug code here https://github.com/GetDKAN/dkan/blob/7.x-1.x/modules/dkan/dkan_dataset/dkan_dataset.module#L1301 to see if you can find out what the hang up is

chrodos:
@janette I have started getting somewhere. This is the part of the code that gets activated:
``if (isset($node->field_harvest_source_modified) &&
    !empty($node->field_harvest_source_modified->value())) {
    $date = format_date($node->field_harvest_source_modified->value(), 'iso_8601_date');``

For some reason it detects all datasets as a harvested datasets. None of the datasets though are harvested.

chrodos:
@janette I checked the 2 tables **field_data_field_harvest_source_modified** and **field_data_field_harvest_source_issued** and they are populated. These are the values of the $node->field_harvest_source_modified and $node->field_harvest_source_modified->value() for a specific dataset.

```
field_harvest_source_modified=Property field_harvest_source_modified
field_harvest_source_modified_value=1519208082
```

janette:
@chrodos thank you for taking the time to  investigate that, the PR above should fix the issue.

PR #2526, opened 2018-05-21
Change default harvest source modified date to blank

janette:
connects #2495 
Non-harvested datasets should not have values for 
- **harvest source issued** 
- **harvest source modified**

## How to reproduce

1.  Create a new dataset through the UI. 
2. Check the database and note that there is an entry in the **field_data_field_harvest_source_modified** table for the node.
3. Future edits to the dataset will always use this modified date.

## QA Steps

- [ ] Create a new dataset through the UI.
- [ ] Verify that there is not an entry in the **field_data_field_harvest_source_modified** table for the node
- [ ] If you edit the dataset, an entry will be created in the **field_data_field_harvest_source_modified** table but the value will be NULL

### DIFF
--- a/modules/dkan/dkan_dataset/dkan_dataset.install
+++ b/modules/dkan/dkan_dataset/dkan_dataset.install
@@ -130,38 +130,4 @@ function dkan_dataset_update_7003() {
   taxonomy_vocabulary_delete($vocabulary->vid);
 }
 
-/**
- * Make sure revised settings for harvest dates are applied. See #2495.
- */
-function dkan_dataset_update_7004() {
-  features_revert_module('dkan_dataset_content_types');
-
-  // Get a list of harvested datasets.
-  $harvested_nids = db_select('field_data_field_harvest_source_ref', 'hsr')
-    ->fields('hsr', array('entity_id'))
-    ->execute()
-    ->fetchCol();
-
-  // Get a list of all datasets.
-  $dataset_ids = db_select('node', 'n')
-    ->fields('n', array('nid'))
-    ->condition('type', 'dataset')
-    ->execute()
-    ->fetchCol();
-
-  // Process all datasets to clear the harvest dates from non-harvested entries.
-  $count = 0;
-  foreach ($dataset_ids as $nid) {
-    // Delete the entries for non-harvested datasets.
-    if (!in_array($nid, $harvested_nids)) {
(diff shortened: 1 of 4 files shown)


chrodos:
@janette For the datasets that have already been submitted through the UI, I guess that truncating the field_data_field_harvest_source_modified table (ONLY IF no harvesting is used in the site) will fix the issue for all right?

janette:
@chrodos yes, even if you do have harvested datasets you could truncate the table and just re-harvest.
```
drush sql-query 'TRUNCATE TABLE field_data_field_harvest_source_modified;'
drush sql-query 'TRUNCATE TABLE field_data_field_harvest_source_issued;'
```

chrodos:
@janette Tested it. It still adds an entry in the field_data_field_harvest_source_modified

janette:
@chrodos is there a value in the `field_harvest_source_issued_value`? Looks like editing a dataset will create an entry but the value is NULL

chrodos:
@janette You are right. Both of these values were set to "now" as default. I changed them to no default values and it works. I will run some more tests.

janette:
Thanks, I need to add a hook_update for that.

modules/dkan/dkan_dataset/dkan_dataset.install
```diff
@@ -129,3 +129,39 @@ function dkan_dataset_update_7003() {
   $vocabulary = taxonomy_vocabulary_machine_name_load('odfe_theme');
   taxonomy_vocabulary_delete($vocabulary->vid);
 }
+
```
dafeder: @janette our standard is to only add update functions to dkan.install since the repo consolidation

test/features/resource.editor.feature
```diff
@@ -92,7 +92,7 @@ Feature: Resource
     When I click "Manage Datastore"
     Then I should see "There is nothing to manage! You need to upload or link to a file in order to use the datastore."
 
-  @resource_editor_6 @datastore @noworkflow @javascript
+  @resource_editor_6 @datastore @noworkflow @javascript @fixme
```
dafeder: Is this related to the PR/bug or just a test that is randomly failing?

janette: Not related to the PR, just a redundant test that consistently fails due to the 30 second time limit.

PR #2532, opened 2018-05-22
Refactor datastore access logic

### LINKED ISSUE #2523: Permissions for Datastore
We should refactor the permissions for the Datastore and use them in the new API. 

## Acceptance Criteria

* Access callback should not make reference to feeds
* dkan_datastore_download_access should no longer be needed
* Generally clean up the page and permissions hooks in dkan_datastore.module to fit new paradigm as much as possible
* API resources should use same permissions logic as UI equivilants

dafeder:
Fixes #2523

This PR refactors the access logic for the datastore pages to fit our new model. 

## QA Steps

- [ ] Visit a dataset node while logged out. You should not see a "manage datastore" or "data api" tab.
- [ ] Visit a resource node while logged out. You should not see a "manage datastore or data api" tab.
- [ ] Visit a resource node while logged in as a site manager. You _should_ see a "manage datastore" tab and _should not_ see a "data API" tab.
- [ ] Navagate to the "manage datastore" tab. Configure a datastore manager and run the import.
- [ ] Run cron.
- [ ] You should now see a "Data API" tab on the resource node. (NOTE: This page does not yet contain any content.)
- [ ] While logged in, the "Data API" and "Manage Datastore" tabs should not appear on any nodes that are not resource nodes.
- [ ] Log out. You should now see a "Data API" tab on the resource that has been edited.

I'm not sure it's practical to do manual testing of the API permissions, but do review the code. When we write tests for the new datastore stuff we can make sure all of that is working.

## Note

Circle config has been modified slightly to stop installing modules unneccesarily and to change the order (behat after phpunit and linting)


modules/dkan/dkan_datastore/psr-4/Resource.php
```diff
@@ -69,11 +77,63 @@ public static function createFromDrupalNodeNid($nid) {
...
+    // Return false if wrong type.
+    if ($node->type != self::resourceContentType()) {
+      return FALSE;
+    }
     $id = $node->nid;
     $file_path = self::filePath($node);
     return new self($id, $file_path);
   }
 
+  /**
```
dafeder: These following four functions may not be necessary. This is to preserve legacy functionality that allows the content type and fields used with the datastore to be completely switched if someone wants to use the datastore with a completely different module. In theory I think this is good, but it may be simpler to strip out for now while we're in MVP mode. Also, even if we keep it I don't think these alter hooks are the most elegant way to allow customization -- just a variable would probably make more sense.

dafeder:
Remaining test issues:

@resource_editor_5: Fails locally, cannot reproduce outside of behat
@resource_editor_6: Passes locally, cannot reproduce
@dataset_all_7: Fails locally, maybe due to phpunit not cleaning up after itself? (was trying to run phpunit first)
@workflow_03: Passes locally, cannot reproduce

dafeder:
@fmizzell I'm hoping you can take a look and see if you can fix remaining test issues, as you're focused on the datastore right now and I'm on deadline for some other stuff. We'll see if that last commit fixes any of the above issues

dafeder:
(if you merge @fmizzell don't forget to squash and merge; LOT of small commits in this one!)

modules/dkan/dkan_datastore/dkan_datastore.module
```diff
@@ -413,14 +414,12 @@ function dkan_datastore_file_link_field() {
 
 /**
  * Returns name of node type that the Datastore is attached to.
+ *
+ * DEPRECATED: Use Resource::resourceContentType
```
fmizzell: Should we clean this up?

dafeder: @fmizzell this was more in case someone was using this in another module or something.

dafeder:
Back to you @fmizzell

PR #2539, opened 2018-05-29
Datastore tests

### LINKED ISSUE #2481: Write tests for the new Datastore importer API

fmizzell:
Fixes #2481 

**QA:**
- Tests should pass


test/phpunit/dkan_datastore_api/DkanDatastoreAPITest.php
```diff
@@ -163,7 +168,7 @@ public function test_dkan_datstore_api_offset() {
     );
     $params = dkan_datastore_api_get_params($params);
     $result = dkan_datastore_api_query($params);
-    $this->assertEquals($result['result']->records[0]->state_id, 2);
+    $this->assertEquals(2, $result['result']->records[0]->stateid);
```
dafeder: Let's see if we can allow underscores for column names, this would be a regression.

PR #2541, opened 2018-05-30
Update to PHP7 CLI Container

acouch:
This PR is a test to see the performance difference between php 5.6 and php 7 to help evaluate updating to PHP 7. The dockerfile for this is here: https://github.com/GetDKAN/dkan-docker/blob/master/.docker/php7-cli/Dockerfile

I screwed up the conifg of the PHP7 cli so will have to wait to try an actual test.

### DIFF
--- a/.ahoy/docker-compose.common.yml
+++ b/.ahoy/docker-compose.common.yml
@@ -51,7 +51,7 @@ services:
   # Used for all console commands and tools.
   cli:
     hostname: cli
-    image: getdkan/dkan-docker:php5-cli
+    image: getdkan/dkan-docker:php7-cli
     environment:
       - XDEBUG_CONFIG=idekey=PHPSTORM
     env_file:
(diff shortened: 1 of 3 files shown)


janette:
`PHP Fatal error:  Cannot use 'String' as class name as it is reserved in /var/www/dkan/libraries/json-schema/src/JsonSchema/Constraints/String.php on line 18`

- We are using json-schema 1.37 https://github.com/GetDKAN/open_data_schema_map/blob/7.x-2.x/open_data_schema_map.make#L12
- PHP error fixed in 1.4.0 https://github.com/justinrainbow/json-schema/pull/136
- Latest version of json-schema is 5.2.7)

janette:
**dataset_rest_api.feature**
` Type error: Argument 1 passed to DkanDatastore::setupSourceBackground() must be an instance of FeedsSource, boolean given, called in /var/www/dkan/modules/dkan/dkan_datastore/includes/DkanDatastore.inc on line 431 (Behat\Testwork\Call\Exception\FatalThrowableError)`

PR #2557, opened 2018-06-11
Wrap Dkan defined string to improve multilingual support

rhabbachi:
We support a couple of Dkan installs in Spanish. This is a simple patch to wrap a number of user facing strings with the standard t() function. 

## QA Steps

The changes should be strait forward but this would be a bit tricky to test with the default Dkan setup since multilingual is not yet part of it. We worked on a [Dkan module (dkan_l10n_es_co)](https://github.com/angrycactus/dkan_l10n_es_co) that enables the spanish language but  requires a number of i18n Drupal modules.

- [ ] Add a language using the i18n interface.
- [ ] Make sure that the newly wrapped strings are translated in the i18n translation interface.
- [ ] Enable the new language for the current context (as default or for the logged in user).
- [ ] On the /search page. The text form should show a translated "search" string.
- [ ] On any dataset view page. The License displayed should be translated.
- [ ] On any dataset edit page. The Licenses displayed in the select list should be translated.

## Reminders

- [ ] ~There is test for the issue.~ Not sure translations are supported in default Dkan yet. 
- [x] CHANGELOG updated.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,7 +1,3 @@
-7.x-1.x
-----------
-- #2557 Wrap user facing strings with Drupal's t() function.
-
 7.x-1.15.1
 ----------
 - #2533 #2546 Updates docker-compose to use updated cli container
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,7 @@
+7.x-1.x
+----------
+- #2557 Wrap user facing strings with Drupal's t() function.
+
 7.x-1.15.1
 ----------
 - #2533 #2546 Updates docker-compose to use updated cli container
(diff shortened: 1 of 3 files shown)


dafeder:
@rhabbachi maybe I'm doing something wrong but after pulling up your code on probo and enabling Locale, the searching for several strings in the license names, i'm not seeing anything?
image

dafeder:
Also don't worry about that one failing test, that's on our end.

rhabbachi:
@dafeder I'm able to see the License strings in the translation interface.

image

I'm suspecting a cache issue or a translation module not enabled.

This is the list of i18n modules enabled for our setup:
image

Potential suspect: `i18n_strings`.

rhabbachi:
I also don't have access to the Probo site.

dafeder:
Ah, I didn't realize i18n_strings was required, although probably should have from your description. Since the change follows best coding practices anyway and I can see from your screenshot that it works, I'm going to merge this without further QA.

dafeder:
(once we get that one stupid test fixed, which should be imminent!)

PR #2561, opened 2018-06-13
Test improvements

janette:
connects https://github.com/NuCivic/dkan_management/issues/66

https://docs.google.com/spreadsheets/d/19bpU7SGeNWrHssH2_dX7LUOCHjXmzlCBrRRSvLO3nkE/edit#gid=1450917050
Search functionality tests are a common failure, this PR only simplifies background usage and removes duplicate scenarios.

CDT, ND, VA are not indexing new data when testing, this will need to be addressed if it continues in 1.16

### Reduce duplicate search scenarios
`search.feature` and `dataset.all.feature` duplicate many scenarios and tests for facet functionality spend a lot of time building a large background to test a single facet, I'm combining those into a single scenario.

**dataset.all.feature** 
- dataset_all_1: the front page will be custom no point in testing content here
- dataset_all_2: duplicate of search 2
- dataset_all_3, dataset_all_4: These often fail on client sites when dates are the same
- dataset_all_7: moving to search.feature
- dataset_all_8: duplicate of search
- dataset_all_9: moving to search.feature
- dataset_all_11: duplicate of search
- dataset_all_12: duplicate of search
- dataset_all_14: duplicate of search

### DIFF
--- a/test/features/search.feature
+++ b/test/features/search.feature
@@ -2,9 +2,6 @@
 @api
 # features/search.feature
 Feature: Search
-  In order to see a dataset
-  As a website user
-  I need to be able to search for a word
 
   Background:
     Given I am on the homepage
@@ -30,6 +27,10 @@ Feature: Search
     And group memberships:
       | user    | group    | role on group        | membership status |
       | Gabriel | Group 01 | administrator member | Active            |
+    And "Format" terms:
+      | name   |
+      | csv 2  |
+      | html 2 |
     And "Tags" terms:
       | name         |
       | something01  |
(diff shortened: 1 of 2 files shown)


test/features/search.feature
```diff
@@ -39,48 +40,68 @@ Feature: Search
...
+    And I should see "2" items in the "datasets" region
+    When I press "Reset"
+    Then I should see all published search content
 
-  Scenario: Filter by facet tag
+  @search_03
+  # Sites with long lists of facet items will fail unless you filter first.
+  Scenario: Filter by facets
+    # Tag
     Given I am on the "Dataset Search" page
```
dafeder: While this technically works, it's not correct to have steps other than the first in a scenario start with "Given"

See http://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenarios

janette: done

PR #2562, opened 2018-06-13
Fixing the url we are getting docker from for CircleCI.

fmizzell:


dafeder:
That test is failing due to a known issue that will be fixed in another PR.

PR #1590, opened 2016-12-06
Fix socrata file download in harvest

rhabbachi:
Issue: CIVIC-5048

## How to reproduce

Running this script should not result in errors.
```
<?php

/**
 * @file
 * Example file for getting file type info from getRemoteFileInfo.
 */

use dkanDataset\getRemoteFileInfo;

include_once 'getRemoteFileInfo.php';

$url = 'http://www.nd.gov/gis/apps/NDHUB.TraumaCenters.csv';
$url = 'https://data.wa.gov/api/views/mu24-67ke/rows.csv?accessType=DOWNLOAD';

$file_info = new GetRemoteFileInfo($url, 'test', TRUE, '/tmp');
var_dump($file_info->getType());
var_dump($file_info->getName());
```

## Acceptance Criteria
* Update test to account for the error.

PR #2599, opened 2018-07-19
Add new curl function to limit how much of the file is returned

janette:
Connects https://github.com/NuCivic/healthdata/issues/1062
## Recap
When running `$info = $this->getFileInfoHelper($url, FALSE)` the content of the file is copied in full as `curl_setopt($ch, CURLOPT_RANGE, '0-1000'); ` is not honored.
This causes out of memory errors when harvesting.

Creating a second helper function to run if `$info = $this->getFileInfoHelper($url, TRUE)` does not return the headers we need.
 
## QA Steps
- Create harvest source from https://data.cdc.gov/data.json
- run `drush dkan-h data_cdc_gov`
- confirm the harvest finishes AND the resource files are correctly saved as remote files and not as an api


janette:
this is similar to previous code https://github.com/GetDKAN/dkan/pull/1590

modules/dkan/dkan_dataset/includes/getRemoteFileInfo.php
```diff
@@ -163,7 +197,7 @@ private function getFileInfoHelper($url, $no_body = TRUE) {
       curl_setopt($ch, CURLOPT_NOBODY, 1);
     }
     curl_setopt($ch, CURLOPT_FOLLOWLOCATION, TRUE);
-    curl_setopt($ch, CURLOPT_HTTPHEADER, array("Range: bytes=0-1000"));
+    //curl_setopt($ch, CURLOPT_HTTPHEADER, array("Range: bytes=0-1000")); Not Effective.
```
fmizzell: Since this is not working, lets go ahead an remove it.

janette:
ok back to you @fmizzell

PR #2620, opened 2018-08-09
Datastore one-step form

### LINKED ISSUE #2619: Make datastore import a one-step process
Under the new datastore system, there is a two-step form to chose a manager and then import. This should really only be one step. Any options around the datastore should be on a single form.

## Acceptance Criteria

* There is a setting for default datastore manager for any site, which is user-configurable
* When a resource is created, the datastore resource object is created immediately, based on the default manager
* If multiple managers are enabled a selector is visible at the top of the datastore form, as long as the import has not started
* Once the data has been imported, the selector is hidden

dafeder:
Fixes #2619 

Improve the UX of datastore import form and related pages by making the interface a single-page form.

### One importer available

(Normal state upon install)

- [ ] Create a CSV resource
- [ ] Click "manage datastore".
- [ ] Run the import. Confirm that UX is intuitive and no choices are presented about what importer to use
- [ ] Drop the datastore. Confirm that UX is intuitive and datastore returns to initial state and can be re-imported

### Two importers available

- [ ] Enable fast import module
- [ ] Create a CSV resource
- [ ] Click "manage datastore"
- [ ] Confirm that there is an importer selector
- [ ] Run the import
- [ ] Confirm that correct importer was used
- [ ] Drop the datastore
- [ ] Change to a different importer and run the import again
- [ ] Confirm that the selected importer is used.

## Known issues

* Currently no way to set default importer, which was part of original acceptance criteria
* Config object now made on node creation. It's made for _any_ resource, not just ones with CSV files
* Still lacking help text

### DIFF
--- a/modules/dkan/dkan_datastore/dkan_datastore.module
+++ b/modules/dkan/dkan_datastore/dkan_datastore.module
@@ -63,32 +63,11 @@ function dkan_datastore_menu() {
   $items['node/%node/datastore'] = array(
     'title' => 'Manage Datastore',
     'page callback' => 'drupal_get_form',
-    'page arguments' => array('dkan_datastore_pages', 1, "importForm"),
+    'page arguments' => array('dkan_datastore_form', 1),
     'access callback' => 'dkan_datastore_access',
     'access arguments' => array('manage', 1),
     'file' => 'dkan_datastore.pages.inc',
     'weight' => '15',
-    'type' => MENU_LOCAL_TASK,
-  );
-  $items['node/%node/datastore/import'] = array(
-    'title' => 'Import',
-    'page callback' => 'drupal_get_form',
-    'page arguments' => array('dkan_datastore_pages', 1, "importForm"),
-    'access callback' => 'dkan_datastore_access',
-    'access arguments' => array('import', 1),
-    'file' => 'dkan_datastore.pages.inc',
-    'weight' => 10,
-    'type' => MENU_DEFAULT_LOCAL_TASK,
-  );
-
-  $items['node/%node/datastore/drop'] = array(
-    'title' => 'Drop Datastore',
-    'page callback' => 'drupal_get_form',
-    'page arguments' => array('dkan_datastore_pages', 1, "dropForm"),
-    'access callback' => 'dkan_datastore_access',
-    'access arguments' => array('drop', 1),
-    'file' => 'dkan_datastore.pages.inc',
-    'weight' => 12,
     'type' => MENU_LOCAL_TASK,
   );
   return $items;
@@ -225,12 +204,9 @@ function dkan_add_resource($node) {
 function dkan_datastore_node_insert($node) {
   $type = dkan_datastore_node_type();
   if ($node->type == $type) {
-    if (!isset($node->feeds)) {
-      // Feeds only saves if there is a form present. We want this to work
-      // programmatically as well.
-      feeds_node_prepare($node);
-      feeds_node_update($node);
-    }
+    $resource = Resource::createFromDrupalNode($node);
+    $datastore_manager = Factory::create($resource, 'Dkan\Datastore\Manager\SimpleImport\SimpleImport');
+    $datastore_manager->saveState();
   }
 }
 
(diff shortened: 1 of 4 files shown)


modules/dkan/dkan_datastore/src/Pages.php
```diff
@@ -67,14 +69,9 @@ private function chooseManagerForm($form) {
...
-      '#title' => t('Choose a Datastore Manager:'),
+      '#title' => t('Change datastore importer:'),
       '#options' => $options,
-    );
-
-    $form['actions'] = array('#type' => 'actions');
-    $form['actions']['submit'] = array(
-      '#type' => 'submit',
-      '#value' => t("Save"),
+      '#default_value' => "\\$class",
```
dafeder: This seems a little strange/hacky - @fmizzell let me know if you think there's a better way to do this.

modules/dkan/dkan_datastore/src/Pages.php
```diff
@@ -211,93 +232,89 @@ public function batchProcess($datastore_manager, &$context) {
...
     return $form;
   }
 
+  /**
+   * Format the class name to something prettier
+   */
+   private function formatClassName($classname) {
+     foreach(dkan_datastore_managers_info() as $info) {
+       if ('\\' . $classname == $info->getClass()) {
+         return $info->getLabel();
```
dafeder: It should not be this hard to get the label for a manager class! @fmizzell

modules/dkan/dkan_datastore/dkan_datastore.module
```diff
@@ -225,12 +204,9 @@ function dkan_add_resource($node) {
 function dkan_datastore_node_insert($node) {
   $type = dkan_datastore_node_type();
   if ($node->type == $type) {
-    if (!isset($node->feeds)) {
-      // Feeds only saves if there is a form present. We want this to work
-      // programmatically as well.
-      feeds_node_prepare($node);
-      feeds_node_update($node);
-    }
+    $resource = Resource::createFromDrupalNode($node);
```
dafeder: I wonder if it is an issue that this creates a manager state object in the DB for _every_ resource, even though it will only ever be used with datastore-compatable resources. I suppose maybe it doesn't hurt anything to have them all in there?

modules/dkan/dkan_datastore/src/Pages.php
```diff
@@ -109,9 +127,9 @@ public function importForm($form, &$form_state) {
           );
         }
         else {
-          $form["datastore_manager_{$property}"] = [
+          $form['import_options']["datastore_manager_{$property}"] = [
             '#type' => 'textfield',
-            '#title' => "{$property}",
+            '#title' => ucfirst(t("{$property}")),
```
dafeder: Not crazy about this but works

modules/dkan/dkan_datastore/src/Pages.php
```diff
@@ -211,93 +232,89 @@ public function batchProcess($datastore_manager, &$context) {
...
+   */
+   private function formatClassName($classname) {
+     foreach(dkan_datastore_managers_info() as $info) {
+       if ('\\' . $classname == $info->getClass()) {
+         return $info->getLabel();
+       }
+     }
+     // Fallback if this fails for some reason.
+     $nameBits = explode('\\', $classname);
+     return end($nameBits);
```
dafeder: Maybe not necessary?

modules/dkan/dkan_datastore/src/Resource.php
```diff
@@ -94,17 +94,22 @@ private static function getNidFromUuid($uuid) {
...
-      $drupal_uri = $node->field_upload[LANGUAGE_NONE][0]['uri'];
+      // We can't trust that the field_upload array will contain a real uri, so
+      // we need to load the full file object.
+      $file = file_load($node->field_upload[LANGUAGE_NONE][0]['fid']);
+      $drupal_uri = $file->uri;
       return drupal_realpath($drupal_uri);
     }
     if (!empty($node->field_link_remote_file)) {
+      $file = file_load($node->field_link_remote_file[LANGUAGE_NONE][0]['fid']);
       stream_wrapper_restore("https");
```
dafeder: Not sure if this is doing anything?

fmizzell: It is, we modify stream wrapper setting to be able to handle remote files correctly, so we want to set it back when we are done.

modules/dkan/dkan_datastore/dkan_datastore.module
fmizzell: mmm... doesn't the createFromDrupalNode() method checks some level of compatibility?

Issue #2622, opened 2018-08-09
Restore Datastore API pages


dafeder:
The "API" tab for resources has broken in the refactor. Restore it.

## Acceptance Criteria

* The "API" link on resources goes to a similar page to what it used to
* API link is only visible on resources with a loaded datastore

kimwdavidson:
Merged for expediency. Closing.

PR #2626, opened 2018-08-15
Restoring the API page.

fmizzell:
Connects #2622


janette:
QA Steps
- [ ] The "API" link on resources goes to a similar page to what it used to
- [ ] API link is only visible on resources with a loaded datastore

janette:
```
Node 61 doesn't have a proper file path. (Exception)
    │
    │  Current URL: http://web/
    │  Only a Selenium2Driver supports screenshots.
    │
    ╳  Warning: file_put_contents(assets/test-failure-html-dump-2018-08-27-11-27-06_resources-.html): failed to open stream: No such file or directory in vendor/devinci/devinci-behat-extension/src/Devinci/DevinciExtension/Context/DebugContext.php line 106
```

Issue #2631, opened 2018-08-17
Add search sort by created date


erogray:
Currently, out-of-the-box DKAN allows sorting of search results by "Date changed" and "Title". As a data consumer, I'd like to also be able to sort by created date so that I can view search results in the order that they were generated.

See for example https://data.city.kyoto.lg.jp/search?query=&sort_by=created&sort_order=DESC&f%5B0%5D=type%3Adataset&f%5B1%5D=type%3Aresource&op=%E6%A4%9C%E7%B4%A2

<img width="461" alt="screen shot 2018-08-17 at 3 49 04 pm" src="https://user-images.githubusercontent.com/21176376/44291724-2acf5f80-a235-11e8-9351-79883df59832.png">

janette:
@erogray I can have a PR for this by EOD

erogray:
See also https://github.com/GetDKAN/dkan/issues/2396

PR #2636, opened 2018-08-20
Add created date as a sort option on the Search pages

janette:
connects #2631 

## QA Steps
- [ ] Go to the datasets search page, confirm you can sort by date created
- [ ] Go to a group page, confirm you can sort by date created

## Reminders


test/dkanextension/src/Drupal/DKANExtension/Context/DatasetContext.php
```diff
@@ -129,6 +129,10 @@ public function iShouldSeeTheFirstDatasetListInOrder($number, $orderby, $sortdir
     }
 
     switch ($orderby) {
+      case 'Date created':
```
dafeder: @janette looking at this test code, I'm pretty sure that it's going to give a false positive. Where is it actually checking the contents of the first four datasets against expected values? All the assertion at the end of the test seems to check is that at least four datasets exist.

janette: @dafeder do you think this check for the order of result titles does not work? https://github.com/GetDKAN/dkan/blob/a977f6b54f3a8534df2b8643faa03a199b05f392/test/dkanextension/src/Drupal/DKANExtension/Context/DatasetContext.php#L161

dafeder: LOL you're right. I misunderstood the code 🤦.

Issue #2630, opened 2018-08-16
Group membership should skip Pending


erogray:
When DKAN site users are added to Groups, they should become active members of the Group immediately and not need a separate activation. If Editors, they should also immediately become Group admins.

This ought to dramatically reduce user role confusion and make the site easier to use.

See also #2623 and #2624

janette:
Just need to fix the pending membership state: https://www.drupal.org/project/og/issues/2744405

The group role is already automated https://github.com/GetDKAN/dkan/pull/1786

Issue #2574, opened 2018-06-20
Enforce workflow roles vs group roles at the time of role change


erogray:
Per https://github.com/GetDKAN/dkan/issues/2465, it can be confusing for users if a role change doesn't behave as expected, which can happen when workflow roles and group administrator roles are changed independently.

As a site manager, I'd like to be prevented from assigning a content creator/workflow contributor to a group administrator role without changing their workflow role also, so that I'm not confronted with unexpected status changes later.

dafeder:
This was fixed in #2644

PR #2644, opened 2018-08-28
Automate group role assignments

janette:
connects #2630
connects #2574

Group roles (member, administrator member) are confusing site managers, these roles should be handled entirely behind the scenes so that site managers don't have to worry about them.

## Issues

- When a user is assigned to a group from the user profile page, they are only in a pending state. You need to go to the group UI to set the user to 'active'. This is resolved by upgrading OG to 2.10
- The group administrator role should only be assigned to editors or site managers since it involves the ability to edit other user's content. This issue is addressed by adding a check before assigning the group administrator role.

## How to reproduce

- [ ] Go to a user account and edit, assign the user to a group and save
- [ ] Visit the group's 'people' page `group/node/[nid]/admin/people` and confirm that the user just added is in a pending state
- [ ] Go to a group's add user page `group/node/[nid]/admin/people/add-user` and add a content creator user, and check the box to give them the group administrator role. Then go to the user's account page and edit any setting and save.
- [ ] Confirm that the group administrator role was revoked.

## QA Steps

- [x] Create a new user, give them the content creator role and add the user to a group and save.
- [x] Go to the group's people page and confirm that the newly added user is 'active' and not 'pending'

- [x] Now use the modify role operation to assign the group administrator role to the user
- [x] Confirm that you get a message that the user was also assigned the editor role. And the user is a  group administrator.
- [x] Click the link in the message to go to the user edit page, uncheck the editor role and save.
- [x] Confirm that the group administrator role was also removed from the user.
- [x] Enable dkan_workflow and repeat the above steps using workflow roles

### DIFF
--- a/drupal-org.make
+++ b/drupal-org.make
@@ -237,7 +237,7 @@ projects:
       url: https://git.drupal.org/project/multistep.git
       revision: 3b0d40a
   og:
-    version: '2.9'
+    version: '2.10'
     patch:
       1090438: https://drupal.org/files/issues/og-add_users_and_entities_with_drush-1090438-12.patch
       2549071: https://www.drupal.org/files/issues/og_actions-bug-vbo-delete.patch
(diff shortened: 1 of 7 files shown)


erogray:
Specified QA steps appear to work correctly, and thus this PR works as advertised; however I found that the steps below (not listed above) did not:
- Use the people editor page to change a group member from Content Creator to Editor
- Confirm that this group member is now a Group Admin.

This may be unrelated work, but seems like an important omission.

kimwdavidson:
@janette: looks like this is ready to merge then?

PR #2652, opened 2018-09-05
Refactoring of the Datastore Page and fixes to failing and incorrect tests

fmizzell:
The purpose of the refactoring is to make the logic in what used to be the Pages class a little clearer after the UX modifications.


kimwdavidson:
@fmizzell to add details

modules/dkan/dkan_datastore/src/Manager/Factory.php
```diff
@@ -10,45 +10,68 @@
...
-      if ($state) {
-        $class = $state['class'];
-      }
-      else {
-        return NULL;
+  private $resource;
+
+  private $classes = [];
+
+  public function __construct(Resource $resource) {
```
dafeder: Document all these methods

modules/dkan/dkan_datastore/src/Page/Component/ManagerConfiguration.php
```diff
@@ -0,0 +1,61 @@
...
+
+namespace Dkan\Datastore\Page\Component;
+
+use Dkan\Datastore\Manager\ManagerInterface;
+
+class ManagerConfiguration {
+
+  private $datastoreManager;
+
+  public function __construct(ManagerInterface $datastore_manager) {
```
dafeder: use `$manager`

modules/dkan/dkan_datastore/src/Page/Page.php
```diff
@@ -0,0 +1,186 @@
...
+    $this->form = $form;
+    $this->formState = $form_state;
+  }
+
+  public function get() {
+    try {
+      $resource = Resource::createFromDrupalNode($this->node);
+
+      /* @var $datastore_manager ManagerInterface */
+      $datastore_manager = (new Factory($resource))->get();
```
dafeder: `$manager`

modules/dkan/dkan_datastore/src/Page/Page.php
```diff
@@ -0,0 +1,186 @@
...
+        (new ManagerSelection($resource, $datastore_manager))->submit($value);
+
+        // The manager got configured we have to reload.dk
+        $datastore_manager = (new Factory($resource))->get();
+      }
+
+      $value = [];
+      foreach ($values as $property_name => $v) {
+        if (substr_count($property_name, "datastore_manager_config") > 0) {
+          $value[$property_name] = $v;
```
dafeder: Rename to `$manager_values`

test/features/search.feature
```diff
@@ -61,8 +61,11 @@ Feature: Search
     When I search for "DKANTest"
     Then I should see "4 results"
     And I should see "4" items in the "datasets" region
-    When I press "Reset"
-    Then I should see all published search content
+    # Deactivating for not, pressing reset is causing for all 46 indexed items to
```
dafeder: Let's have @janette take a quick look at this - if we really can't fix we can merge like this but let's make sure to try reactivating soon.

janette:
@fmizzell @dafeder we stopped setting the search page default to filter by dataset, so the $search_results count is counting all content, not just datasets. So I'm wondering why this test didn't fail before.
https://github.com/GetDKAN/dkan/pull/2576/files#diff-84dc305e0b21b857a093c1ae5d649028R494

_search_02


So two ways to fix it:
add_step

OR

datasetcontext

janette:
@dafeder the test is passing, ready for merge

dafeder:
Install file looks like it wasn't linted - comments above still open.

Also variable names in tests and install file are not fixed

PR #2664, opened 2018-09-13
Add UI to change allowed extensions

janette:
Currently custom file extensions are added to custom_config with a hook_form_alter() and when updates are made to DKAN the downstream sites do not get those updates as the override holds true.

By adding a UI where site managers can make adjustments it is easy to get new updates and make changes without a code change.

form

### QA Steps
- [ ] Log in as sitemanager
- [ ] Navigate to DKAN > DKAN Dataset Forms
- [ ] Add an extension 
- [ ] Create a new resource
- [ ] Confirm you can upload a file with that extension


erogray:
Seems to work correctly with some known extensions (pptx, for instance) but other less predictable ones (RPP, js) fail, though they're shown in the helptext.

There also seems to be a character count limit in the extensions input field that's significantly shorter than the field display width.

dafeder:
We want to add the hook @fmizzell wrote for a different project to let this use the same method for overriding.

fmizzell:
This might be related and relevant https://github.com/GetDKAN/dkan/pull/2517

erogray:
Decided to use variables, not hook

Issue #2684, opened 2018-09-29
More control over datastore progress


fmizzell:
In the new datastore in branch ```datastore-improvements``` we use a hybrid approach of trying to import in a batch process in the browser and transitioning to imports during cron for large files.

When a resource starts to import with cron, the user does not have much control to what is happening, and the information provided is not very helpful.

It would be great to be able to see when a cron job is actually running and importing (In Progress), versus when it is in the queue for that process but not actively being imported (Paused). 

Finally it would be great to be able to interrupt/pause an active migration if necessary.

PR #2689, opened 2018-10-05
Pause Import

fmizzell:
Connects #2684
**QA**
- Create a resource with a large file, like https://data.ca.gov/sites/default/files/SafeToSwim_2018-09-04.csv
- Start a datastore import from the UI
- Run cron from drush
- Go back to the UI, you should see that the datastore is in progress and a stop button should be available
- Press the stop button
- A message will let you know that the import will be stopped soon
- Refres the page
- The status should now be paused

### DIFF
--- a/drupal-org.make
+++ b/drupal-org.make
@@ -5,7 +5,7 @@ includes:
 - https://raw.githubusercontent.com/NuCivic/visualization_entity/7.x-2.0/visualization_entity.make
 - https://raw.githubusercontent.com/NuCivic/open_data_schema_map/7.x-2.1/open_data_schema_map.make
 - https://raw.githubusercontent.com/NuCivic/leaflet_draw_widget/5a5f8faf664aeca02371f6692307580d9fab9116/leaflet_widget.make
-- https://raw.githubusercontent.com/NuCivic/recline/7.x-2.0/recline.make
+- https://raw.githubusercontent.com/NuCivic/recline/working-with-new-datastore/recline.make
 projects:
   admin_menu:
     version: 3.0-rc5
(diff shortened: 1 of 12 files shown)


dafeder:
## QA Report

After trying to add that file as a remote resource, I got this:

image

Uploading it worked without issue and I was able to start an import. We still have the issue where the message we see after the batch is "import finished" when in fact it is not complete. Still, I was able to run cron and re-start the import. 

"Stop" works.

What I'm concerned about (and I think we've discussed this before) is that when you're in that paused state, it's very unclear where to go from there. The only button available is "Drop." It seems like we should have some sort of message to the user explaining that importing has been paused, and will resume on the next cron run. 

I also wonder if we might want to make an import button available in the paused state, where it would let you at least move a bit further ahead again.

dafeder:
Another interesting note: after letting cron run for a long time the second time, it stalled eventually and displayed this:

image

When I run cron again, it completes immediately and does not import any further items. At this point, the datastore is still in an "In Progress" state. Pressing the "Stop" button does not seem to have an effect.

dafeder:
I can no longer import any other resource once the site gets into the state described above ^^^

dafeder:
Back to the remote file issue: the resource works as expected despite warnings after creation. I can import a file to the datastore, resume on cron, and use the stop button. The cron will "time out" in the same manner as with the local file as described above, and leave the site in the same broken state.

fmizzell:
@dafeder I just pushed some code to restrict the datastore from running during UI cron calls. The datastore imports are meant to be long running processes, so they don't make sense to run from the UI. I also changed the message given when batch completes to be less deceiving.

PR #2703, opened 2018-10-18
Update dkan_migrate_base.migrate.inc

acouch:
This fixes a type error that is raised by adding strings using PHP 7+.


acouch:
Closing for #2702

PR #2710, opened 2018-10-24
Fix the data story view to only display stories

janette:
Resource nodes and unpublished nodes are showing up on the /stories view

Replacing the search index view with simpler node type view.

### QA steps
- create a dataset with the title 'Fresh Produce Purchases - Quarterly Totals', see if it turn up on `/stories`


dafeder:
Note that this completely eliminates the need for a story search index and server, which to me seems like a plus. People will need to delete the server and index manually.

This is a heavier touch than I'd usually want to see on just a patch release (this is destined for 1.15.5) but the current functionality is broken on PHP7 and we can't figure out why. Given Acquia's server updates (which go into effect today) it seems like we need to get this in now rather than later.

PR #2730, opened 2018-11-07
Test improvements

### LINKED ISSUE #2646: Create smoke tests
It's not practical to run the full suite of DKAN tests on sites that have been configured, customized, etc. We should have a core group of "smoke tests" that make sure nothing has broken after an upgrade, but don't try to run every core DKAN test because there are too many differences.

## Acceptance Criteria

- Look at Javascript tests and choose a small set that is representative of the major components of DKAN
- Tag added to smoke tests
- Resolve workflow issue
- A minimum number of tests are now used for project sites for testing releases (modify site repos' circleci  config to run only smoke tests)
- Document config changes for sites

janette:
We need to get tests working after these changes. This PR does not necesarrily fix #2646

Connects #2646 

### Issues
File copy failing on circlci
- @group_sm_01
- @resource_cc_26
- @resource_cc_27
- @theme_01
- @theme_02


woodt:
I checked out this branch, built a new environment and confirmed that the behat tests are runnable.

Issue #2333, opened 2018-01-17
JSON- or file-based data dictionaries at resource level


dafeder:
_Work in progress_

## User story

As a data consumer, I want both human-readable and machine-readable data dictionaries for tabular data that explain a dataset at the column level and conform to an established standard.

## Acceptance Criteria
* New module: dkan_data_dictionary
* Creates new content type: data_dictionary
* _Data dictionaries_ are attached to _resources_
* Data dictionaries are _not_ presented as an option on the search
* A data dictionary field contains these fields:
  * Specification: select which specification to use (we'll start with just tabular data) or "other" - see below for details
* A JSON data field for the data itself
* The dictionary/schema is validated against the specification selected
* Sync group ownership and workbench draft status with parent resource

# Follow-up tickets
- [x] User-friendly form to replace raw JSON entry
- [x] Theming/display logic for dictionary
- [ ] File field for optional upload/linking, allow CSV upload

**For discussion:** should this be a paragraph bundle rather than a node?

kimwdavidson:
https://data.nal.usda.gov/data-dictionary-blank-template

kimwdavidson:
Possible standard: frictionlessdata.io/specs/table-schema/

dafeder:
This features currently lives in #2380. Re-opening until that's merged.

dafeder:
Note that most of this work has moved to #2608

dafeder:
Picking this back up in DKAN 2 under different framing.

PR #2738, opened 2018-11-08
Data Dictionaries: Reroll branch

dafeder:
This is a reroll of #2380.
Connects #2333


dafeder:
Thinking we should break this out into another module for now.

PR #2739, opened 2018-11-08
Docker proxy and xdebug updates

dafeder:
This was an old branch found recently that looks like it may contain some useful code we don't want to lose; let's review and check with @kducharm before killing it.


dafeder:
@kducharm do you remember what this branch was for, and if it was replaced with another one that was merged or anything like that?

dafeder:
No longer needed.

PR #2743, opened 2018-11-13
Check CircleCI settings and make sure we're using opcache

dafeder:
Look on circle settings (and maybe SSH in) to confirm that opchache and opcachecli is being used on Circle. Code in this PR may or may not be relevant.

### DIFF
--- a/.ahoy/.scripts/circle-behat.rb
+++ b/.ahoy/.scripts/circle-behat.rb
@@ -24,14 +24,17 @@ error = 0
 parsed = behat_parse_params(ARGV)
 params = behat_join_params parsed[:params]
 files = balance(parsed[:files])
+composer = ""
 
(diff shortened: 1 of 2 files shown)


acouch:
This PR is no longer valid. Currently opcache is enabled for the web:

```bash
root@web:/# php -i | grep opcache
/etc/php/7.1/cli/conf.d/10-opcache.ini,
opcache.blacklist_filename => no value => no value
opcache.consistency_checks => 0 => 0
opcache.dups_fix => Off => Off
opcache.enable => On => On
opcache.enable_cli => Off => Off
```
I created a PR in the docker project to enable it for the cli: https://github.com/GetDKAN/dkan-docker/pull/8

Not sure if this should get closed or if we need another ticket to track this. Also noticed that the web container runs ubuntu 18.04/ php 7.1 while the cli runs ubuntu 16.04/php 7.0 .

Issue #2759, opened 2018-11-30
Update default Workflow content state change emails


erogray:
Workflow moderation state change emails could use some improving; new text has been written (in https://github.com/NuCivic/dkan_management/issues/126) and needs deployment. In addition, "Published to Needs Review" emails are not enabled for the original author, the Workflow Moderator or the Workflow Supervisor. These should be enabled within the code.

## Acceptance Criteria
- [ ]  /admin/config/workbench/email  shows "Published to Needs Review" emails enabled for author, moderator and supervisor
New text is in place in email template at options at /admin/config/workbench/email for each of the following, as specified in https://github.com/NuCivic/dkan_management/issues/126:
- [ ] draft to needs review, original author
- [ ] draft to needs review, Workbench Moderator
- [ ] draft to needs review, Workbench Supervisor
- [ ] needs review to draft, original author
- [ ] needs review to draft, Workbench Moderator
- [ ] needs review to draft, Workbench Supervisor
- [ ] needs review to published, original author
- [ ] needs review to published, Workbench Moderator
- [ ] needs review to published, Workbench Supervisor
- [ ] published to needs review, original author
- [ ] published to needs review, Workbench Moderator
- [ ] published to needs review, Workbench Supervisor

PR #2767, opened 2018-12-07
Issue GetDKAN/dkan#2759: workflow emails changes

dharizza:
connects GetDKAN/dkan/issues/2759

Improvements for text in workflow moderation state change emails.

### DIFF
--- a/modules/dkan/dkan_workflow/dkan_workflow.features.workbench_email.inc
+++ b/modules/dkan/dkan_workflow/dkan_workflow.features.workbench_email.inc
@@ -118,28 +118,27 @@ For assistance or support, please contact a Site Manager or an administrator fro
       'to_name' => 'draft',
       'author' => 0,
       'automatic' => 1,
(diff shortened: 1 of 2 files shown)


janette:
@stefaniefgray I've removed references to Page and Blog content since page content is not handled by the workbench interface (only site managers can create/edit) and blog content does not exist in OOB dkan.

Data story is also not configured to be run through the workbench UI but I think that is a separate issue.

janette:
@stefaniefgray I noticed a problem with the instructions for where to find the "My workbench" link and I've updated the email text, can you update the docs

https://dkan.readthedocs.io/en/latest/components/workflow.html
> When logged in as a user that has been assigned a Workbench role, the “My Workbench” button will be displayed on the site’s main navigation toolbar.

should be

> When logged in as a user that has been assigned a Workbench role, the “My Workbench” button will be displayed in the administration menu at the top.

PR #2780, opened 2018-12-24
Fist attempt at running tests with dktl.

fmizzell:
Connects GetDKAN/dkan-tools#14

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,11 +59,11 @@ jobs:
       - attach_workspace:
           at: ~
       - run:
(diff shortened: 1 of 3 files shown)


woodt:
I think this is ready to review again.  I believe there are still some latent test ordering issues (I will create tickets) and there were two tests I needed to skip (again, I will create a ticket), but I found a value of parallelism (3) that runs the behat tests successfully.

Also now collecting test artifacts (e.g., screen captures) and JUnit style test results in a form that CircleCI can interpret.

Issue #2577, opened 2018-06-21
Confirm that all allowed extensions are also recognized by Harvest


erogray:
It seems like Harvest can get throw errors when harvesting resource filetypes that are allowed by the site. The site-allowed filetypes should also be harvestable.

erogray:
See also https://github.com/NuCivic/cadepttech/issues/60 and https://github.com/GetDKAN/dkan/issues/2575

erogray:
can be tested by adding filetype to overrides that isn't normally allowed, then harvesting with fake json file that includes the unusual filetype.

dafeder:
@erogray can you list some specific examples of this happening (with URLs) with some steps to reproduce?

kimwdavidson:
this may no longer be an issue. @janette to test

janette:
@stefaniefgray 
1. On a demo flavor site, add a custom extension via the UI `admin/dkan/dataset_forms`
2. create a json file to harvest that includes a resource with the custom extension http://myjson.com/
3. create a harvest source using the json you created
4. harvest and confirm the resource with the custom extension  was created

stefaniefgray:
QA pass!

I did things a bit differently:

1. Allowed mp3 files on http://stefanie.dkandemo.nuamsdev.com and http://andrew.dkandemo.nuamsdev.com. (Note: After adding a new file format, clear the cache so that DKAN will allow you to upload that format.)

2. Uploaded a dataset with an mp3 to http://stefanie.dkandemo.nuamsdev.com/dataset/meow-meow-meow

3. Created a new Harvest from http://andrew.dkandemo.nuamsdev.com with keyword = "meow" and proceeded to Harvest from the Stefanie demo site.

4. Success! The mp3 resource is now located at http://andrew.dkandemo.nuamsdev.com/dataset/meow-meow-meow/resource/a60d6662-e62f-4df3-91e0-ec1a5f6ee5f0.

Ticket can be closed.

PR #2781, opened 2018-12-26
Use dkan_allowed_extensions() during harvest

rhabbachi:
connects #2577
## Description.

The `dkan_allowed_extensions()` function (added in #2664 is the way to go when overriding resources allowed extensions. During a harvest a check on the allowed extensions is done to honor the user setting. This PR will update this check (done in HarvestMigration::prepareResourceHelper) to use `dkan_allowed_extensions()` instead of querying the field settings. 

## How to reproduce

Step-by-step instructions for reproducing the bug:
1. Add harvest source that have resources with extensions that are not part of the default field setting but allowed via the `dkan_allowed_extensions()`.
2. Run Harvest on the source.

**Expected**:
All the resources are harvested and imported correctly.

**Currently**:
Resources with non default extensions introduced via `dkan_allowed_extensions()` fail to be harvested.

## QA Steps

1. Add harvest source that have resources with extensions that are not part of the default field setting.
2. Make sure to allowed the non default extension via the `dkan_allowed_extensions()` mechanism.
2. Run Harvest on the source.
 
The Harvest should proceed without issues.

## Merge process

- N/A

## Reminders

### DIFF
--- a/modules/dkan/dkan_harvest/dkan_harvest.migrate.inc
+++ b/modules/dkan/dkan_harvest/dkan_harvest.migrate.inc
@@ -1144,12 +1144,11 @@ class HarvestMigration extends MigrateDKAN {
       // page.
       if (!is_null($remoteFileInfo->getType())
         && !in_array($remoteFileInfo->getExtension(), $html_extensions)) {
-
         // Check if this file extension is allowed for field_link_remote_file.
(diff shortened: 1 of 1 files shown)


janette:
Thank you @rhabbachi it works!

Issue #2635, opened 2018-08-20
Bug: harvest machine names should validate for length and allowed characters


erogray:
When creating a Harvest, the machine name must be shorter than 64 characters (including the already-33-chars-long prefix `migrate_map_dkan_harvest_migrate_`) or the harvest will (without visible errors or other indications) fail to complete migration, and will not create datasets properly. We should put some validation on the harvest name field so that it's not too long.

Another issue that sometimes causes problems is when the machine name contains a hyphen. When a site manager adds a Harvest with a title that contains a hyphen, the machine name will contain a hyphen also, and running this harvest will then result in errors; it's also difficult to correct the machine name after the fact. The easiest way to prevent this would be to sanitize the input so that hyphens are not included in the machine name in the first place.
<img width="281" alt="screen shot 2018-10-29 at 10 22 20 am" src="https://user-images.githubusercontent.com/21176376/47668083-c863db00-db64-11e8-8fd4-4c8a6ff8ee23.png">
<img width="1165" alt="screen shot 2018-10-29 at 10 23 07 am" src="https://user-images.githubusercontent.com/21176376/47668084-c863db00-db64-11e8-8077-06af3ccccd5b.png">

PR #2784, opened 2018-12-29
Update Harvest node machine name field settings to limit length and replace single hypen

rhabbachi:
connects #2635 

We ran into similar issues as stated in #2635, when a site admin tries to be too explicit with the Harvest Source name and hits database field limitations in terms of length and allowed characters.

## How to reproduce
See #2635

## QA Steps

1. Open node/add/harvest-source.
2. Input "harvest-machine-name-that-is-too-long-to-be-true-because-this-is-really-really-really-long" into the `Title` field.

*Result*: the machine name should show "harvest_machine_name_that_is_to"
 

## Merge process

- N/A

## Reminders


janette:
This is great @rhabbachi can you also update the help text to include the character limit in case the user is adding similar long source names

`A URL-safe version of the text. It may only contain lowercase letters, numbers and underscores and may not exceed 31 characters. Leave blank to re-generate.`

warning

rhabbachi:
@janette Done!

janette:
Fabulous! Thank you @rhabbachi

Issue #2802, opened 2019-01-17
Disable Feeds module


dafeder:
The new Datastore module should make the Feeds module unnecessary. Let's include an update hook that will disable the module.

We probably need to keep the module in the codebase for a few more releases so that it doesn't break.

dafeder:
Assigning back to @janette for testing

PR #2803, opened 2019-01-17
Remove feeds modules

janette:
clean up db


erogray:
https://github.com/GetDKAN/dkan/issues/2802 disables feeds; do we want to also remove it in 1.18?

PR #2834, opened 2019-02-22
Drush command for revision prune

### LINKED ISSUE #2825: Harvest revision clean up
Create a script to purge duplicate revisions from existing datasets.

For datasets with multiple revisions with the same timestamp, keep a single revision and delete the rest.

dafeder:
Fixes #2825 

Running `drush ddrp` will find all duplicate dataset revisions from the same timestamp, and reduce them to a single revision. You can limit the number done in one run using `--node-max` and `--revision-max` options.

### DIFF
--- a/modules/dkan/dkan_dataset/dkan_dataset.drush.inc
+++ b/modules/dkan/dkan_dataset/dkan_dataset.drush.inc
@@ -1,0 +1,91 @@
+<?php
+
+/**
+ * @file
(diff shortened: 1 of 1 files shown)


dharizza:
Notes: 
- The command created for this is `drush ddpr` which is alias for `drush dkan-dataset-prune-revisions`.
- When both options `--node-max` and `--revision-max` are present, the `--revision-max` is the one which limits the amount of revisions being deleted.

PR #2839, opened 2019-02-27
Drush command for deleting orphaned legacy datastore tables

dharizza:
## User story

The ghost datastore tables are those for which there doesn't exist a related node, for example there exist a table `feeds_datastore_dkan_file_10000` but there isn't a node with id `10000`. Here we're adding a new drush command `drush dkan-datastore-delete-ghost-tables` with alias `drush dkan-datastore-dgt` to find ghost datastore tables and delete them.

## QA Steps

- When you run `drush dkan-datastore-drop-orphan-tables` or `drush dkan-datastore-dot` you'll get the following behaviour:
  - [ ] If there are no datastore tables, you'll get a message saying "There are currently no datastore tables on this site." (this will happen on a fresh install when you haven't imported anything to datastore).
  - [ ] If there are datastore tables, you'll get a message displaying the node IDs it is looking for (node IDs for which there is a datastore table).
  - [ ] Number of ghost tables found.
  - [ ] If there are ghost tables, it deletes them and shows the respective table names.

Connects NuCivic/dkan_management#171

### DIFF
--- a/modules/dkan/dkan_datastore/dkan_datastore.drush.inc
+++ b/modules/dkan/dkan_datastore/dkan_datastore.drush.inc
@@ -20,6 +20,13 @@ function dkan_datastore_drush_command() {
     ),
   );
 
+  // Delete ghost datastore tables.
+  $items['dkan-datastore-delete-ghost-tables'] = array(
+    'aliases' => array('dkan-datastore-dgt'),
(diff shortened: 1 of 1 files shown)


dharizza:
@dafeder fixed.

PR #2862, opened 2019-04-12
Import manager improvements

dharizza:
If in a site we need to execute any steps right after the import to datastore process, right now we don't have an easy way to do it, so in this PR we are introducing a new hook. This hook is called hook_datastore_post_import and it is fired when an import process is completed.

Right now, the dkan datastore tables don't have a primary key field, which may lead to issues when someone needs to index the items from a datastore, so in this PR we're also altering the datastore table schema to include a serial `entry_id` field which will work as the primary key.

## QA Steps

- [ ] Import a resource to the datastore, the resulting table should contain a field `entry_id` as primary key.
- [ ] Implement a hook_datastore_post_import and make sure it fires after a resource is imported into the datastore.

### DIFF
--- a/modules/dkan/dkan_datastore/src/Manager/Manager.php
+++ b/modules/dkan/dkan_datastore/src/Manager/Manager.php
@@ -245,6 +245,7 @@ abstract class Manager implements ManagerInterface {
     if ($import_state === self::DATA_IMPORT_DONE) {
       $this->stateDataImport = self::DATA_IMPORT_DONE;
       $this->saveState();
+      module_invoke_all('datastore_post_import', $this->resource);
     }
     elseif ($import_state === self::DATA_IMPORT_PAUSED) {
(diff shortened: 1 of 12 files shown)


modules/dkan/dkan_datastore/src/Manager/Manager.php
```diff
@@ -235,6 +245,7 @@ public function import() {
     if ($import_state === self::DATA_IMPORT_DONE) {
       $this->stateDataImport = self::DATA_IMPORT_DONE;
       $this->saveState();
+      module_invoke_all('datastore_post_import', $this->resource);
```
janette: @dharizza do we still need this?

dharizza: Yes we do, because that's what will allow us to execute custom actions after a datastore import is completed.

Issue #2850, opened 2019-03-22
Harvest should allow mediaType to specify resource type


erogray:
As a site manager harvesting from a remote JSON source, I want resources that are created to use a resource type that is specified by the mediaType in the JSON, when it is specified, so that in cases where the JSON and remote server mimetype conflict, the resulting resource is correctly typed according to what the JSON intends.

Add a brief description about the purpose of the issue.

## Steps to Reproduce
Add steps for reproducing the issue including any file(s) needed to reproduce.

1. Create a harvest from https://bythenumbers.sco.ca.gov/data.json
2. Confirm that some resources in the JSON use "mediaType": "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"
}
3. Confirm that these resources show up as type "html" or "Web page", though, if downloaded, they're XLSX files

## Acceptance Criteria
Add the goals and desired results that this issue should accomplish when complete.

- [ ] for each resource, if a JSON specifies a mediaType, the harvest should use the type that's specified as the Resource type
- [ ] for each resource, if the JSON does not specify a mediaType, the harvest should use the getfileinfo value (as is currently the default behavior)

erogray:
Originally prompted by research on https://github.com/NuCivic/cadepttech/issues/165 in which the JSON specifies a mediaType that is not the same as the remote server. We think that it is more POD-correct to follow the JSON's specification.

PR #2863, opened 2019-04-15
Allow mediaType to specify resource type during Harvest

dharizza:
Connects #2850 

Description. If mediaType is defined, use its respective extension as the format for resources harvested.

## User story

As a site manager harvesting from a remote JSON source, I want resources that are created to use a resource type that is specified by the mediaType in the JSON, when it is specified, so that in cases where the JSON and remote server mimetype conflict, the resulting resource is correctly typed according to what the JSON intends.

## How to reproduce

1. Create a harvest from https://bythenumbers.sco.ca.gov/data.json
2. Confirm that some resources in the JSON use "mediaType": "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"
3. Confirm that these resources show up as type "html" or "Web page", though, if downloaded, they're XLSX files

## QA Steps

- [x] Create a harvest from https://bythenumbers.sco.ca.gov/data.json
- [x] Confirm that some resources in the JSON use "mediaType": "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"
- [x] Run the harvest
- [x] Confirm that resources with mediaType are using that to get the resource format.
- [x] Confirm that resources with mediaType set to `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet` are shown as resources with format `xlsx`.

### DIFF
--- a/modules/dkan/dkan_harvest/modules/dkan_harvest_datajson/dkan_harvest_datajson.migrate.inc
+++ b/modules/dkan/dkan_harvest/modules/dkan_harvest_datajson/dkan_harvest_datajson.migrate.inc
@@ -257,18 +257,22 @@ class DatajsonHarvestMigration extends HarvestMigration {
 
       $format = $resource_row_data->format;
 
-      // If mediayType is defined, then get the right format based on that mimetype.
-      if (isset($resource_row_data->mediaType)) {
-        include_once DRUPAL_ROOT . '/includes/file.mimetypes.inc';
-        $mimetype_mappings = file_mimetype_mapping();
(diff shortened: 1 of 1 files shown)


janette:
@dharizza, two problems
1.  what if the $extensions_lookup is empty
2. what if the lookup is different from the format given as 'format'

After further testing I think we should stick with $resource_row_data->format by default and if that is empty, then use the $mimetype_mappings value.

for example, .lpk is not in the $mimetype_mappings
1.  add `lpk nc dat` to the allowed extensions field on `admin/dkan/dataset_forms`
2. create harvest source from `https://data.marine.gov.scot/data.json` with filter: title = Updating Fisheries Sensitivity Maps in British Waters

extensions

dharizza:
@janette done, can you test again, please?

Issue #2861, opened 2019-04-12
Add National Data Portals for Pacific Island Countries to DKAN Sites List


tavitas:
Hello,

The Secretariat of the Pacific Regional Environment Programme (SPREP) project Inform has launched national data portals for Pacific Islands countries below based on dkan, please add to the dkan sites list. 

American Samoa: [https://americansamoa-data.sprep.org](url)
Cook Islands: [https://cookislands-data.sprep.org](url)
Federated States of Micronesia: [https://fsm-data.sprep.org](url)
Fiji: [https://fiji-data.sprep.org](url)
Kiribati: [https://kiribati-data.sprep.org](url)
Republic of the Marshall Islands: [https://rmi-data.sprep.org](url)
Nauru: [https://nauru-data.sprep.org](url)
New Caledonia: [https://newcaledonia-data.sprep.org](url)
Niue: [https://niue-data.sprep.org](url)
Palau: [https://palau-data.sprep.org](url)
Papua New Guinea: [https://png-data.sprep.org](url)
Samoa: [https://samoa-data.sprep.org](url)
Solomon Islands: [https://solomonislands-data.sprep.org](url)
Tonga: [https://tonga-data.sprep.org](url)
Tuvalu: [https://tuvalu-data.sprep.org](url)
Vanuatu: [https://vanuatu-data.sprep.org](url)

Pacific Environment Data Portal: [https://pacific-data.sprep.org](url)

stefaniefgray:
See https://github.com/GetDKAN/dkan/pull/2869

tavitas:
Thanks guys! Really happy to see them listed

stefaniefgray:
Page just needs a final review of my edits and then it's good to go

PR #2868, opened 2019-05-01
Add pacific islands data sites to the docs

janette:
## QA
Confirm sites from #2861 have been added


stefaniefgray:
See https://github.com/GetDKAN/dkan/pull/2869

PR #2874, opened 2019-05-10
Security update for Drupal to 7.67 (SA-CORE-2019-007)

rhabbachi:
Security advisory for Drupal core - Moderately critical - Third-party libraries
https://www.drupal.org/sa-core-2019-007


janette:
Thanks @rhabbachi

PR #2879, opened 2019-05-14
Delete current revisions on unpublished nodes

dharizza:
It isn't possible to delete current revisions on unpublished nodes, when a user needs to "reject" a revision for an unpublished node it should be possible for them to completely delete the node. Right now for unpublished nodes those drafts are there forever and user would need to delete each node one by one. Here we are adding a new view on workbench moderation page for showing the drafts created for unpublished nodes.

## Steps to test:
- [ ] There is a new tab in admin/workbench with the title "Unpublished nodes draft", it should show _only_ unpublished nodes with revisions (there should be a badge at the top of the tab with the amount of revisions that met the conditions).
- [ ] In this page there should be bulk operation to delete the nodes completely, it should work correctly.
- [ ] The "My drafts" page should work as usual.


janette:
I'm getting an undefined index error on the confirmation screen
error

dharizza:
@janette can you check again? Just fixed it. Also, removed reference to dkan_feedback logo in the logos section at the bottom of workflow pages.

PR #2888, opened 2019-06-02
Datastore library

fmizzell:
We have a decoupled version of our datastore that currently lives at https://github.com/fmizzell/datastore and is being used in https://github.com/getdkan/dkan2.

This PR connects that library to the datastore implementation.

**Gotchas:**

- The datastore library does not have an interrupt mechanism, that feature needs to be added.
- Recline has some datastore logic hardcoded that needs to be replaced for resource pages to work:
`$resource = Resource::createFromDrupalNode($node);
      /* @var  $manager ManagerInterface */
      $manager = \Drupal\Dkan\Datastore\Util::getManager($resource);`

This is in recline.theme.inc around line 545


erogray:
@fmizzell are additional tickets needed for the Gotchas?

erogray:
@fmizzell is this PR still in action? Or has it been surpassed by other work?

janette:
iceboxing

PR #2903, opened 2019-06-15
Hide some harvest messages

dharizza:
Hide groups and datastore messages during harvest runs and from the status page right after it finishes, because those messages are not helpful for the user when multiple datasets/resources are added during harvest.

## How to reproduce

1. Create a harvest source from https://open.fda.gov/data.json. Click through to "Harvest Now".
2. Once the harvest has completed, note that there are a large number of messages in green, some of them are related to nodes (like  "Node 43 doesn't have a proper file path.") but some don't have an ID so nothing is actionable from "This filemime type (application/zip) can be added as a resource, but cannot be imported to our datastore." or "Groups were updated on 1 resource(s).".

## QA Steps

- [ ] Create a harvest source from https://open.fda.gov/data.json. Click through to "Harvest Now".
- [ ] While harvest is running, try to load any other pages, you shouldn't see messages like "This filemime type (SOME FILEMIME) can be added as a resource, but cannot be imported to our datastore." or "Groups were updated on X resource(s).".
- [ ] Once the harvest has completed, on the dkan harvest dashboard you shouldn't see messages like "This filemime type (SOME FILEMIME) can be added as a resource, but cannot be imported to our datastore." or "Groups were updated on X resource(s).".

### DIFF
--- a/modules/dkan/dkan_harvest/dkan_harvest.module
+++ b/modules/dkan/dkan_harvest/dkan_harvest.module
@@ -412,6 +412,7 @@ function dkan_harvest_cache_sources_in_batch(array $harvest_sources = array(), $
  *   The context associated with the batch process.
  */
 function dkan_harvest_migrate_source_batch_op($harvest_source, array $options = array(), array &$context = array()) {
+  $_SESSION['dkan_harvest_running'] = TRUE;
 
   // Load migration.
   $migration = $harvest_source->getMigration();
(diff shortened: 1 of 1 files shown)


janette:
@dharizza this looks great but getting an undefined offset error message, we might need an override in nuboot_radix for `Implements theme_status_messages()`

radix

dharizza:
Talked to @janette and we decided to merge this one and create another ticket to handle the offset error message, the corresponding ticket is: https://github.com/GetDKAN/dkan/issues/2909

PR #2908, opened 2019-06-22
Implement periodic updates functionality

dharizza:
Added dkan_periodic_updates module.

As a site manager I can upload a manifest for a periodic import of resources.
There should be an administrative page where user can upload the manifest, which is a csv file with format resource_id,frequency,file_url.
The file_url will be used to download the file in a temporary directory, then that file will be used to update the resource with the resource_id specified.
The frequency can be daily, weekly or monthly.

## Steps to test
- [x] Enable dkan_periodic_updates.
- Go to `admin/dkan/periodic-updates`:
  - [x] Try to upload a file that is not a csv, you should get an error message.
  - [x] Try to upload a csv file with any headers, it should throw a message as it doesn't has the right headers.
  - [x] Upload a CSV files with the columns `resource_id`, `frequency` and `file_url`, it should be saved correctly.
  - [x] Mark the checkbox `Enable periodic updates.` and click on `Save settings`.
- Go to `admin/dkan/periodic-updates/status`:
  - [x] If the "Enable periodic updates" checkbox is not marked, then you should get a message saying "Periodic updates are disabled.".
  - [x] If the "Enable periodic updates" checkbox is marked but no manifest has been uploaded, then you should get a message saying "No manifest was found.".
  - [ ] If the updates are enabled and a valid manifest is uploaded, you should see a list of the elements included in the previously uploaded manifest file, it should include the `UUID`, the `File URL`, the `Update frequency`, `Status` (which represents the update status), `Last update` (which is the date in which the resource was last updated) and a link to see the resource.
- [ ] Once the updates are enabled and there is a valid manifest uploaded, on every cron run the manifest will be evaluated to see what resources needs to be updated based on the last update date and the frequency specified in the manifest (the frequency can be `monthly` or `weekly`, every other value will be applied as `daily`).
- [ ] Run `drush cron run` to trigger the updates and keep checking `admin/dkan/periodic-updates/status` to see the differences in the updates status.
- For every updated resource check:
  - [ ] on each node just one file is loaded
  - [ ] the updated date on the node matches the current date 
  - [ ] make sure the data is imported into the datastore (only time this can change is when the periodic update status page reports an error when importing to the datastore, in these cases, user would need to re-import manually).

Note: documentation as added and also PHP Unit tests.

### DIFF
--- a/docs/components/periodic-updates.rst
+++ b/docs/components/periodic-updates.rst
@@ -13,7 +13,7 @@ The module requires a CSV file to execute the updates, this file is known as "th
 Periodic Updates Settings
 -------------------------
 
-You can find the periodic updates settings under DKAN --> Periodic Updates (`/admin/dkan/periodic-updates`).
+You can find the periodic updates settings under DKAN --> Periodic Updates --> Settings (`/admin/dkan/periodic-updates`).
 The settings for this module includes:
 
   - Periodic updates state
--- a/docs/components/periodic-updates.rst
+++ b/docs/components/periodic-updates.rst
@@ -1,0 +1,55 @@
+DKAN Periodic Updates
+=====================
+
+The DKAN Periodic Updates module provides the user the ability to execute automated updates on existing resources in a daily, weekly or monthly basis. It also provides a UI in which you can check at the status of the updates.
+The module requires a CSV file to execute the updates, this file is known as "the manifest", when you upload it and enable the updates, this is what happens:
+
+- The manifest is evaluated on every cron run to determine what resources needs to be updated (based on the frequency defined in the manifest and the "last update" date).
+- For each resource that needs an update, the system will:
+  - Download the file URL specified in the manifest (or link it, according to which file field is used on the corresponding resource).
+  - Update the resource node to use the new file.
+  - If the file is a CSV or TSV, then the resource is imported into the datastore.
+
+Periodic Updates Settings
+-------------------------
+
+You can find the periodic updates settings under DKAN --> Periodic Updates (`/admin/dkan/periodic-updates`).
+The settings for this module includes:
+
+  - Periodic updates state
+  - Manifest for periodic updates
+
+1- Periodic updates state
+*************************
+
+To turn periodic updates on you need to be in the Periodic Updates page, once there you should mark the option "Enable periodic updates" and click on "Save settings". You'll also need to upload a valid manifest in the section "Manifest for periodic updates", otherwise the updates cannot be executed.
+If you already have a manifest in the system but need to pause the periodic updates, all you need to do is unmark the option "Enable periodic updates" and save settings.
+
+2- Manifest for periodic updates
+********************************
+
+The manifest is a CSV file in which you define all the resources that need to be updated periodically in the system, this file should contain the following column names:
+
+``resource_id,frequency,file_url``
+
+:resource_id: is the UUID related to the resource node.
+:frequency: it can be set to `daily`, `weekly` or `monthly`. If you leave it empty or put any other string, then the system asumes it has to be imported daily.
+:file_url: the URL of the remote file that needs to be downloaded or linked and saved to the node.
+
+Once you upload the manifest and click on "Add file", the file will be recognized by the system.
+
+Note: for the periodic updates to be executed you need to have uploaded a valid manifest and enabled the periodic updates. If you only have one of these items, the updates will not be executed.
+
+Periodic Updates Status
+-------------------------
+
+The status page can be found under DKAN --> Periodic Updates --> Status (`/admin/dkan/periodic-updates/status`). In this page you'll get information about all the resources specified in the manifest.
+If the "Enable periodic updates" checkbox from Settings is not marked, then you'll get a message saying "Periodic updates are disabled.". If it is marked but no manifest has been uploaded, then you'll get a message saying "No manifest was found.".
+If the updates are enabled and a valid manifest is uploaded, then you'll see a list of the elements included in the manifest file, that list includes:
+
+  - the node UUID,
+  - the File URL,
+  - the Update frequency,
+  - the Status: this represents the messages generated on each update, it will tell you wether the resource has been updated, if the process finished correctly or if/what errors were produced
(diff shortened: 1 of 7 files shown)


janette:
undefined-index

modules/dkan/dkan_periodic_updates/dkan_periodic_updates.module
```diff
@@ -0,0 +1,459 @@
...
+      }
+      fclose($handle);
+      $output['state'] = [
+        '#theme' => 'table',
+        '#header' => $table_headers,
+        '#rows' => $table_rows,
+      ];
+    } else {
+      $output['state'] = [
+        '#markup' => '<p>' . t('No manifest was found.') . '</p>',
```
janette: Can you add these classes to the `<p>` tag: `<p class="alert alert-warning">`

modules/dkan/dkan_periodic_updates/dkan_periodic_updates.module
```diff
@@ -0,0 +1,459 @@
...
+    '#value' => t('Add file'),
+  ];
+
+  // Current manifest file.
+  $prev_file_name = '- None -';
+  $message = '<p>Current manifest: ' . $prev_file_name . '</p>';
+  $prev_file = variable_get('dkan_periodic_updates_manifest');
+  if ($prev_file) {
+    $file = file_load($prev_file);
+    $prev_file_name = $file->filename;
```
janette: Probably not a common occurrence but if the user uploads a manifest file then deletes it via the `admin/content/file` UI then they will get warnings here
```
Notice: Trying to get property of non-object in dkan_periodic_updates_form() (line 87 of /var/www/dkan/modules/dkan/dkan_periodic_updates/dkan_periodic_updates.module).
Notice: Trying to get property of non-object in dkan_periodic_updates_form() (line 88 of /var/www/dkan/modules/dkan/dkan_periodic_updates/dkan_periodic_updates.module).
```
if you wrap lines 87-89 in an `if ($file) {}` statement, we still run into problems because of the **dkan_periodic_updates_manifest** variable, so perhaps we need to check that the variable exists, then confirm the file exists, if not then wipe out the variable.... maybe set the message to 'file was removed' or something...

docs/components/periodic-updates.rst
```diff
@@ -0,0 +1,54 @@
+DKAN Periodic Updates
+=====================
+
+The DKAN Periodic Updates module provides the user the ability to execute automated updates on existing resources in a daily, weekly or monthly basis. It also provides a UI in which you can check at the status of the updates.
```
janette: Documentation suggested changes will be in bold only for clarity.

...existing resources **on** a daily, weekly...

docs/components/periodic-updates.rst
```diff
@@ -0,0 +1,54 @@
...
+The module requires a CSV file to execute the updates, this file is known as "the manifest", when you upload it and enable the updates, this is what happens:
+
+- The manifest is evaluated on every cron run to determine what resources needs to be updated (based on the frequency defined in the manifest and the "last update" date).
```
janette: ...**which** resources **need** to be updated...

docs/components/periodic-updates.rst
```diff
@@ -0,0 +1,54 @@
...
+- For each resource that needs an update, the system will:
+  - Download the file URL specified in the manifest (or link it, according to which file field is used on the corresponding resource).
+  - Update the resource node to use the new file.
+  - If the file is a CSV or TSV, then the resource is imported into the datastore.
+
+Periodic Updates Settings
+-------------------------
+
+You can find the periodic updates settings under DKAN --> Periodic Updates --> Settings (`/admin/dkan/periodic-updates`).
+The settings for this module includes:
```
janette: The settings for this module **include**:

docs/components/periodic-updates.rst
```diff
@@ -0,0 +1,54 @@
...
+
+  - Periodic updates state
+  - Manifest for periodic updates
+
+1- Periodic updates state
+*************************
+
+To turn periodic updates on you need to be in the Periodic Updates page, once there you should mark the option "Enable periodic updates" and click on "Save settings". You'll also need to upload a valid manifest in the section "Manifest for periodic updates", otherwise the updates cannot be executed.
```
janette: ~To turn periodic updates on you need to be in the Periodic Updates page, once there you should mark the option "Enable periodic updates" and click on "Save settings".~

Enable periodic updates by checking the "Enable periodic updates" box on the Periodic Updates page, and click on "Save settings".

docs/components/periodic-updates.rst
```diff
@@ -0,0 +1,54 @@
...
+If you already have a manifest in the system but need to pause the periodic updates, all you need to do is unmark the option "Enable periodic updates" and save settings.
```
janette: ... is **uncheck** the option ...

docs/components/periodic-updates.rst
```diff
@@ -0,0 +1,54 @@
...
+
+2- Manifest for periodic updates
+********************************
+
+The manifest is a CSV file in which you define all the resources that need to be updated periodically in the system, this file should contain the following column names:
```
janette: ...you define all **of** the resources...

docs/components/periodic-updates.rst
```diff
@@ -0,0 +1,54 @@
...
+:file_url: the URL of the remote file that needs to be downloaded or linked and saved to the node.
+
+Once you upload the manifest and click on "Add file", the file will be recognized by the system.
+
+Note: for the periodic updates to be executed you need to have uploaded a valid manifest and enabled the periodic updates. If you only have one of these items, the updates will not be executed.
+
+Periodic Updates Status
+-------------------------
+
+The status page can be found under DKAN --> Periodic Updates --> Status (`/admin/dkan/periodic-updates/status`). In this page you'll get information about all the resources specified in the manifest.
```
janette: information about ~all~ the resources specified

dharizza:
Thanks @janette, just did the changes, can you check at this again, please?

janette:
Thank you @dharizza, looks great, scheduled for the 1.17 release

PR #2958, opened 2019-09-11
Stop harvest from checking resource file extention if web page

dafeder:
- [ ] Add unit test for accessURLs that do not have an extension


erogray:
@dafeder does this ticket need more detail?

dharizza:
Closing this as we can't really test it because html are already skipped.

PR #2961, opened 2019-09-16
Make group.all.feature not to be a smoketest.

dharizza:
Remove @smoketest tag to group.all.feature as it contains tests that relies on the amount of groups present in the site. This is to be able to have smoketest work correctly in other dkan based sites.


dharizza:
Closed in favor of #2965 .

PR #2968, opened 2019-10-01
Change the way we're handling dataset resource updates to changed time

dafeder:
This PR changes the logic of how DKAN determines whether to create a "revision log" update to a dataset when a resource is changed. We get a lot of extra revisions to datasets based on changes to resources. Rather than trying to check if the update happened during a harvest, this simply checks if the resource revision is more than a minute older than the dataset's updated date. If not, it's not worth recording.

This should provide more consistency and accuracy, and reduce extra revisions created through automated processes. 

Finally, this removes the queue functionality and makes the update immediately. If this creates much longer wait times for harvests to finish we can revisit, but as implemented, the queueing defeats the whole purpose of the revision logging! If the logging is queued and happens hours (or days) later, the updated timestamp will be still inaccurate, just for different reasons.

# QA Steps

I'm not sure how to QA this beyond the tests. Let's discuss if there are concerns!

### DIFF
--- a/modules/dkan/dkan_dataset/dkan_dataset.module
+++ b/modules/dkan/dkan_dataset/dkan_dataset.module
@@ -537,33 +537,21 @@ function dkan_dataset_node_update($node) {
       dkan_dataset_sync_groups($node);
       break;
 
-    case 'resource':
+      case 'resource':
       // Add changelog on associated datasets.
       $resource = entity_metadata_wrapper('node', $node);
       $datasets = $resource->field_dataset_ref->raw();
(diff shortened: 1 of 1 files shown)


erogray:
@dafeder does this ticket need more detail?

dafeder:
Description added.

janette:
- [x] Tested locally via the UI and works but does take a long time.
- [ ] Test automated **Run Harvest multistep** on staging server

dharizza:
Merging as @janette already did the testing on dev and everything went fine.

PR #2977, opened 2019-10-25
move doc images to S3

### LINKED ISSUE #2166: Reduce repo size by moving documentation images to S3
The repo is 100MB, which while not huge is enough to slow down some operations. While there are local workarounds (such as using git clone with --reference) it would be nice to see if there is a more universal option.

Most of the volume is coming from the /docs directory, especially the images. We want to keep this as part of the project, so documentation can be tracked and reviewed in lockstep with code changes. The simplest method would be to replace the images with git-lfs (https://git-lfs.github.com/) which would avoid the repository bloat. To reduce repository size a script would need to replace the images with git-lfs references in the entire history, so this change would need to be done at some point when feature branches are all merged (or as close to that point as possible, to avoid disruption).

janette:
fixes #2166 

- Reduce size of repo by removing doc images
- Update docs to use remote images

## QA Steps

- [ ] Browse docs/build/index.html


docs/components/workflow.rst
```diff
@@ -36,12 +36,12 @@ DKAN Workflow is included on all out-of-the-box DKAN sites; however, it is not e
 
    drush en dkan_workflow -y
 
-Enabling DKAN workflow will automatically enable all other required modules and add the Workflow Supervisor role to all users already assigned the site manager role. (More information available in the `Workflow Roles`_
+Enabling DKAN workflow will automatically enable all other required modules and add the Workflow Supervisor role to all users already assigned the site manager role. (More information available in the :ref:`Workflow Roles`
```
dharizza: This reference is not working properly, see https://docs.getdkan.com/en/2166-rm-doc-images/components/workflow.html it directs the user to the same place.

janette: Thanks for catching that @dharizza, adding one last commit...I hope

PR #2978, opened 2019-10-30
Reduce unneccesary revisions

dafeder:
Stop creating new revisions throughout the dataset creation process in Harvest, so that we have a cleaner history revision history on newly harvested datasets.

## QA Steps

If tests pass, try running an HHS or LKY harvest using this branch to make sure no ill-effects. You should see a reduced number of revisions (actually, just one revision) on new datasets.


dafeder:
Deprecated in favor of #2983

PR #2983, opened 2019-11-14
Separate harvest classes into separate files and do some light cleanup

dafeder:
Stop creating new revisions throughout the dataset creation process in Harvest, so that we have a cleaner history revision history on newly harvested datasets.

Also, refactoring to make code easier to work with.

Note: The cache dir logic now goes in HarvestMigration by default, and does not need to be added in additional migration classes. See https://github.com/GetDKAN/dkan/pull/2983/files#diff-4e04b5611bd1607adc9902f7c27dbbbcL20-L21

## QA Steps

If tests pass, try running an HHS or LKY harvest using this branch to make sure no ill-effects. You should see a reduced number of revisions (actually, just one revision) on new datasets.

Everything should work the same.


dafeder:
Now includes commits from #2978

PR #2984, opened 2019-11-15
Remove feeds modules

janette:
Feeds no longer needed in the code base.


erogray:
Need to rebase against develop.

PR #2985, opened 2019-11-15
Remove deprecated dkan roles perms module

janette:
cleaning up the code base


erogray:
Need to rebase against develop.

PR #2998, opened 2019-12-20
Move check for file size into the proxy function

janette:
## Out of memory issues

see https://github.com/NuCivic/healthdata/issues/1155

Why are we printing the file? Can we just always pass the link?

### DIFF
--- a/modules/dkan/dkan_datastore/dkan_datastore.pages.inc
+++ b/modules/dkan/dkan_datastore/dkan_datastore.pages.inc
@@ -25,12 +25,12 @@ function dkan_datastore_proxy($node) {
   $type = recline_get_data_type($remote_file['filemime']);
 
   if (in_array($type, $allowed_types)) {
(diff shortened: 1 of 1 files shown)


modules/dkan/dkan_datastore/dkan_datastore.pages.inc
```diff
@@ -25,9 +25,19 @@ function dkan_datastore_proxy($node) {
   $type = recline_get_data_type($remote_file['filemime']);
 
   if (in_array($type, $allowed_types)) {
-    drupal_add_http_header('Content-Type', $mime);
-    drupal_add_http_header('Content-Disposition', 'attachment; filename=' . $filename);
-    print file_get_contents($uri);
+    $size = isset($node_wrapper->field_link_remote_file->value()['filesize']) ? (int) $node_wrapper->field_link_remote_file->value()['filesize'] : 0;
```
fmizzell: @janette if we do not know the size, would it be better to assume that it is huge?

janette: yes, good point, added the change

PR #3009, opened 2020-01-14
Change behat context to specifically index the saved entity.

beeyayjay:
The raw DKAN entity behat context attempts to index new or updated entities using the process() function from SearchAPIContext, which indexes up to 10 entities that haven't yet been indexed. This is fine if the indexes are up-to-date, but if there are enough items that need indexing, it will never get to the one(s) being tested.

This PR solves that problem by explicitly indexing the new/updated entity instead. It may be less efficient, but ensures that test entities will be indexed.

## QA Steps

- [ ] Run behat tests that use the save function from RawDKANEntityContext (e.g. `Given Pages:`, `Given Resources:`, `Given Datasets:`, `Given harvest sources:`). The tests should run correctly.


janette:
@dafeder do you want to review this or should Dharizza or I do it?

dafeder:
Not using search in tests anymore; closing

PR #3012, opened 2020-01-23
Prevent infinite loop in Harvest Source preview page

beeyayjay:
## Description

When a harvest cache doesn't exist, the harvest source preview page calls a batch to generate it. When the batch is done, it loads the preview page. If the cache can't be created, this creates an infinite loop.

This PR checks that  the Content URI is reachable, and if the header is available, checks that it's a json response. If either check fails, it displays a message instead of triggering the batch

## How to reproduce

1.  Create a harvest source with an invalid source URL. A batch display will keep flashing until you browse away from the page

## QA Steps

- Create a test Harvest Source with an invalid URL and save. Confirm that a message appears and it is appropriate. 
- Change the URL one that goes to a non-json page. Confirm that an appropriate message appears.
- Try valid JSON URLs (including, if possible, one that doesn't return headers). Confirm that caches are generated and the preview page presents a summary of them as before.

### DIFF
--- a/modules/dkan/dkan_harvest/dkan_harvest.module
+++ b/modules/dkan/dkan_harvest/dkan_harvest.module
@@ -281,7 +281,7 @@ function dkan_harvest_page_preview($node) {
       $output .= t("<div>Could not connect to <em>@url</em>. Please check the value of 'Source URI' and resubmit.</div>",
           ['@url' => $uri]);
     }
-    elseif (strpos($info->getType(), 'json') === FALSE) {
+    elseif (!empty($info->getType()) && strpos($info->getType(), 'json') === FALSE) {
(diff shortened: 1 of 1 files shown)


janette:
@beeyayjay I'm testing with a source that is a json endpoint but not public (403), and I get the "... is not a json endpoint" message, can we spell out that it is an access denied issue? or use the first message.

beeyayjay:
@janette, GetRemoteFileInfo doesn't provide response codes. I could use that source to roll my own solution  or just make a single, generic message to cover all cases.

janette:
@beeyayjay yes good idea, we don't need to pinpoint the exact problem just report that we can't get to the specified url.

PR #3013, opened 2020-01-27
Avoid proxy file printing.

dharizza:
## Issue
When the remote host does not supply a valid content header, namely Content-Length. We cannot get an accurate file size without downloading the file first.
Since we can not reliably determine file size, we end up trying to print large files that are reported as much smaller files. This is causing out of memory errors, so removing the proxy and call to `file_get_contents`.

This will "disable" the data preview for all remote csv files, it can be restored by adding the file to the datastore.

remote-csv

## Steps to test:

- [ ] Whenever you go to `node/%dkan_datastore_resource/data` you should get the file download, wether it is a CSV or not.


erogray:
Closing because not every DKAN site will want this -- it could be useful for some folks with lots of large remote files, but for newer sites or less data-heavy sites it's unnecessary

PR #3021, opened 2020-02-12
Check accrualPeriodicity against list of valid frequency values.

beeyayjay:
Checks the validity of the accrualPeriodicity field in `prepareRow()`.

## User story

If a json harvest has an `accrualPeriodicity` value, that value gets saved to the `frequency` field, even if the value isn't valid for the field.

The pr adds a check to prepare row. If the value is a valid key value for field_frequency, there is no change. If it is a valid display value, it is set to the key value. If it is neither, it is set to null.

## How to reproduce

1.  Run a harvest against a harvest cache with an invalid value for `accrualPeriodicity`.
2. Browse to the dataset page. The invalid value will be displayed as the frequency.
3. Edit the page. The select value for the frequency field will be set to '- None -'.

## QA Steps

1. Create/modify three harvest cache files with `accrualPeriodicity`  fields. 
a. Give one a valid Frequency key (e.g. `R/P1D`). 
b. Give another a valid Frequency value (e.g. `Daily`). 
c. Give the third an invalid value
2. Run the harvest.
3. Check the harvested datasets. Frequency should appear with its correct value for the first two on both the dataset page and its edit form. Frequency should not appear on the dataset page for the third.

### DIFF
--- a/modules/dkan/dkan_harvest/includes/HarvestMigration.php
+++ b/modules/dkan/dkan_harvest/includes/HarvestMigration.php
@@ -372,6 +372,22 @@ class HarvestMigration extends MigrateDKAN {
         }
       }
     }
+
+    // Check the accrualPeriodicity field.
+    // If it's in the frequency map as a value, set it to the associated key.
+    // If it's not there, either as a key or value, set it to null.
+    if (isset($row->accrualPeriodicity) && $row->accrualPeriodicity != '') {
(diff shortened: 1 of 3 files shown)


beeyayjay:
Wait on this. I just realized the code probably belongs in dkan_migrate_base_data_json.inc instead.

PR #3032, opened 2020-03-06
Use jquery 1.12 by default

dafeder:
jquery 1.10 contains the security issue reported here. It's fixed in 1.12. See http://blog.jquery.com/2016/01/08/jquery-2-2-and-1-12-released/ 

Let's see how much havoc it would create to switch.


dafeder:
Note: Recline field tabs break.

PR #3052, opened 2020-04-27
405 docs revamp

janette:
- module renaming and grouping


janette:
closing in favor of #3084

PR #3056, opened 2020-04-28
Support import_to_datastore field on manifest for periodic updates.

dharizza:
Sometimes the CSV files have unescaped quotes or are way to big and the user doesn't want them to be in the datastore, by default, the periodic updates module takes the manifest and tries to import everything so when a resource has an issue in the corresponding CSV the datastore gets stuck, so the idea is to allow the users to decide what elements from their manifest should be imported to the datastore after being updated.

## User story
As a sitemanager I can decide what resources from the periodic updates manifest should be imported to the datastore.

## QA Steps
- [x] Go to admin/dkan/periodic-updates, enable periodic updates and upload a manifest containing a new column name "import_to_datastore", its possible values are "Y" and "N". Confirm you can upload the file correctly.
- [x] Go to "Status" tab, you should see the state of the elements in your manifest.
- [x] Run cron (from drush you can run dktl drush cron run).
- [x] Confirm that the resources that had a "Y" in the field "import_to_datastore" are updated and imported to the datastore. All the others shouldn't be imported to the datastore.

### DIFF
--- a/modules/dkan/dkan_periodic_updates/dkan_periodic_updates.module
+++ b/modules/dkan/dkan_periodic_updates/dkan_periodic_updates.module
@@ -62,7 +62,7 @@ function dkan_periodic_updates_form($form, &$form_state) {
     '#title' => 'Manifest for periodic updates',
   ];
   $form['manifest-file']['info'] = [
-    '#markup' => '<p>' . t('The manifest file should contain the following column names: resource_id, frequency and file_url. The allowed frequency values are daily, weekly, or monthly. If left blank, the frequency will default to daily.') . '</p>',
+    '#markup' => '<p>' . t('The manifest file should contain the following column names: resource_id, frequency, file_url and import_to_datastore. The allowed frequency values are daily, weekly, or monthly. If left blank, the frequency will default to daily.') . '</p>',
   ];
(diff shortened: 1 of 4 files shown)


janette:
@dharizza can you update the docs and tests as well

modules/dkan/dkan_periodic_updates/dkan_periodic_updates.module
```diff
@@ -248,8 +254,10 @@ function dkan_periodic_updates_get_items_to_update($file) {
 function dkan_periodic_updates_execute_update($resources) {
   module_load_include('inc', 'dkan_datastore', 'resources/dkan_datastore_resource');
   if (!empty($resources)) {
-    foreach ($resources as $uuid => $file_url) {
+    foreach ($resources as $uuid => $resource) {
       $nids = entity_get_id_by_uuid('node', [$uuid]);
+      $file_url = $resource['file_url'];
+      $datastore = $resource['datastore'] === 'Y' ? TRUE : FALSE;
```
janette: I have a manifest with values: Y,n,n,blah in the 'import_to_datastore' column and none of the files are added to the datastore.

PR #3072, opened 2020-05-01
New struct

fmizzell:


janette (changes requested):
- common.info.yml
  * change desc: `Provides common utilities, functions, and base api endpoint used by other DKAN modules.`
- metastore_form.info.yml
   * ~~change name: `name: Metastore Form`~~ (If we follow the same pattern as the other modules this would be "Form", not good, maybe the machine name should be metastore_metadata_form)
   * change desc: `description: Provides an embedded React UI for creating and editing dataset metadata.`
- sample_content.info.yml
  * change name: `name: Sample Content`
  * change desc: `description: Provides sample content for tests or standing up a demo instance.`
- metastore.info.yml
  * change desc: `Provides integration with the metastore library and creates the "Data" content type for storing metadata.`
- modules/harvest/harvest.info.yml
  * change desc: `Provides integration with the harvest library.`
- frontend.info.yml
  * change desc: description: 'Provides the routing connection between Drupal and a decoupled frontend.'

janette:
- metastore_admin/metastore_admin.install
  * change "ground_control" to "metastore_admin"
- metastore.links.menu.yml
  * fix the link `admin/config/dkan/properties`
- missing `/admin/config/dkan/sql_endpoint`?

PR #3089, opened 2020-05-09
Clean up references to old docs

dafeder:
1. Removes any remaining references to docs.getdkan.com and replaces with dkan.readthedocs.io.
2. Removes entire releases section from docs, which is incomplete


janette (changes requested):
Let's also remove `releases/index` from the TOC on index.rst

dafeder:
Done.

PR #3092, opened 2020-05-11
Fix a class namespace in dkan_alt_api routing

thierrydallacroce:


thierrydallacroce:
See https://github.com/GetDKAN/dkan/pull/3093

PR #3104, opened 2020-05-21
ResourceMapper

### LINKED ISSUE #3108: ResourceMapper
One of the most useful features of DKAN is its capability to allow users to explore the data to some degree to help them figure out whether it is the right data for them before downloading it. To allow this data previewing we get a copy of a remote file, parse it, and store it in a database table.

In other scenarios, we might be consuming data from "private" sources and we might need to get a local copy of the files not only for previewing but to simply make the available. In this instances we might also need to know what the local url for public access to the file is. 

The point is that a file can be seen from many perspectives withing DKAN:
1. A remote file
1. A local copy
1. A datastore table
1. A local URL

The FileMapper is an attempt at unifying the management of these different perspectives.

### LINKED ISSUE #3186: Simplifying the communication of referenced information in metadata.
### Problem
Currently, we have 3ish ways to see our metadata:
1. Original
```
{
    "title": "Hello",
    "distribution": {
        "title": "Title",
        "url": "http://site.com/file.csv
    }
}
```
2. DKAN's internal storage 
```
{
    "title": "Hello",
    "distribution": unique-identifier
}
```
3. Original with information about DKAN's internal references to other objects
```
{
    "title": "Hello",
    "distribution": {
        "identifier": unique-identifier,
        "data": {
            "title": "Title",
            "url": "http://site.com/file.csv
        }
    }
}
```

2 is an internal representation and is not really accessible to the users, or even programmatically through the API, but 1 and 3 are accessible.

The current mechanism to choose which version you see/get is a global static variable. This solution has served us well so far, but from the initial implementation, we wanted to do something different. Global flags bring uncertainty and it makes code harder to reason about given its side-effects. Here is some pseudo-code of how this global flag can land us in trouble:

```
set_show_references_global_flag(TRUE);
$distribution = getDataset('123')->getDistribution();
saveDistribution($distribution);
```
Any code that retrieves that dataset after our manipulation of the distribution will most likely fail. The problem here is that we are retrieving data in a format that looks like 3, and saving/storing it. DKAN expects the data that it processes to always look like 1, but we do not have, yet, the mechanism to guarantee that for internal references, like distributions, our data respects the schema (If our pseudo-code tried to save the full dataset, DKAN would have told us that the data did not comply with our schema).

### Solution
The solution we want to implement, eventually, is to validate these pieces of JSON data that are being saved individually against their respective schema. This will protect our data, but the global flag is still a bad way to ask for what one needs. So we are going to solve that first since it is an easier problem to address.

The proposed solution here is to have a single internal representation of our data. Here is what it will look like:
```
{
    "title": "Hello",
    "distribution": {
        "title": "Title",
        "url": "http://site.com/file.csv
    }
    "%Refdistribution": {
        "identifier": unique-identifier,
        "data": {
            "title": "Title",
            "url": "http://site.com/file.csv
        }
    }
}
```
This new format gives everyone that needs information about the internal references in DKAN what they need, while still keeping the original metadata in a state that should not put the system in a bad state even if we accidentally tried to store it.

fmizzell:


composer.json
```diff
@@ -9,6 +9,9 @@
         "composer/installers": "^1.2",
         "cweagans/composer-patches": "^1.5.0",
         "drush/drush": "^10.2.2",
+        "aws/aws-sdk-php": "^3.138",
+        "league/flysystem": "^1.0",
+        "league/flysystem-aws-s3-v3": "^1.0",
```
thierrydallacroce: Were these packages accidentally included?

thierrydallacroce (approved):

modules/common/src/Resource.php
```diff
@@ -0,0 +1,151 @@
...
+  private $filePath;
+  private $identifier;
+  private $mimeType;
+  private $perspective;
+  private $version;
+
+  /**
+   * Constructor.
+   */
+  public function __construct($file_path, $mime_type, $perspective = 'source') {
```
dafeder: Use constant for `'source'`

modules/common/src/Resource.php
```diff
@@ -0,0 +1,151 @@
+<?php
+
+namespace Drupal\common;
+
+use Procrastinator\HydratableTrait;
+use Procrastinator\JsonSerializeTrait;
+
+/**
+ * Class Resource.
```
dafeder: Add some documentation to explain more clearly the concept of a resource, relationship between versions and perspectives.

thierrydallacroce (approved):

modules/datastore/tests/src/EventSubscriber/SubscriberTest.php
```diff
@@ -0,0 +1,58 @@
...
+ */
+class SubscriberTest extends TestCase {
+  use ServiceCheckTrait;
+
+  /**
+   *
+   */
+  public function test() {
+    $url = 'http://hello.world/file.csv';
+    $resource = new Resource('http://hello.world/file.csv', 'text/csv');
```
thierrydallacroce: Probably meant to re-use `$url`:

`$resource = new Resource($url, 'text/csv');`

modules/common/tests/src/Storage/JobStoreTest.php
```diff
@@ -0,0 +1,165 @@
...
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @coversDefaultClass \Drupal\common\Storage\JobStore
+ * @group datastore
+ */
+class JobStoreTest extends TestCase {
+
+  /**
+   *
```
dharizza, line 25: This file has a lot of empty comments. Can we add them?

modules/common/tests/src/Unit/Mocks/FileSystem.php
```diff
@@ -0,0 +1,28 @@
...
+use Psr\Log\LoggerInterface;
+use Drupal\Core\Site\Settings;
+
+/**
+ *
+ */
+class FileSystem extends DrupalFileSystem {
+
+  /**
+   *
```
dharizza, line 18: Empty comments.

modules/datastore/tests/src/Service/ResourceLocalizerTest.php
```diff
@@ -0,0 +1,52 @@
...
+use MockChain\Sequence;
+use PHPUnit\Framework\TestCase;
+use Drupal\common\Resource;
+use Drupal\Core\File\FileSystem;
+use MockChain\Chain;
+use Drupal\common\Storage\JobStore;
+use Drupal\common\Storage\JobStoreFactory;
+
+/**
+ *
```
dharizza, line 15: More empty comments here.

modules/datastore/tests/src/WebServiceApiTest.php
```diff
@@ -18,14 +18,14 @@ class WebServiceApiTest extends TestCase {
...
 
-    $webServiceApi = WebServiceApi::create($container);
-    $result = $webServiceApi->deleteMultiple(["1", "2"]);
+  $webServiceApi = WebServiceApi::create($container);
+  $result = $webServiceApi->deleteMultiple(["1", "2"]);
 
-    $this->assertTrue($result instanceof JsonResponse);
-  }
+  $this->assertTrue($result instanceof JsonResponse);
+  }*/
```
dharizza: This code is all commented, should we remove it?

modules/harvest/tests/src/ServiceTest.php
```diff
@@ -194,18 +194,29 @@ public function destroy() {
...
+   * Private.
+   */
   private function getMetastoreMockChain() {
     return (new Chain($this))
       ->add(Metastore::class, 'publish', '1')
       ->getMock();
   }
 
+  /**
+   *
```
dharizza, line 207: More empty comments.

modules/metastore/metastore.module
```diff
@@ -110,3 +57,32 @@ function metastore_entity_bundle_field_info_alter(&$fields, EntityTypeInterface
     }
   }
 }
+
+/**
+ *
```
dharizza: More empty comments.

modules/metastore/modules/metastore_admin/metastore_admin.module
```diff
@@ -108,6 +108,9 @@ function metastore_admin_preprocess_views_view_field(&$vars) {
   }
 }
 
+/**
+ *
```
dharizza: Another empty comment.

modules/metastore/src/Plugin/QueueWorker/OrphanReferenceProcessor.php
```diff
@@ -66,31 +71,35 @@ public static function create(ContainerInterface $container, array $configuratio
    * {@inheritdoc}
    */
   public function processItem($data) {
-    $property_id = $data[0];
-    $uuid = $data[1];
+    $this->debug(json_encode($data));
```
dharizza: Do we need this call to debug here?

modules/metastore/src/Reference/OrphanChecker.php
```diff
@@ -127,6 +129,11 @@ private function processReferencesInUpdatedProperty($property_id, $old_value, $n
       $old_value = [$old_value];
       $new_value = [$new_value];
     }
+
+    $this->debug($property_id);
+    $this->debug(json_encode('old_value'));
+    $this->debug(json_encode($new_value));
```
dharizza: Are these calls needed?

modules/metastore/src/Storage/Data.php
```diff
@@ -276,6 +256,9 @@ private function updateExistingNode($node, $data) {
     $node->set('moderation_state', $this->getDefaultModerationState());
 
     $node->save();
+
+    $this->debugNode(__FUNCTION__, $node);
```
dharizza: Needed?

modules/metastore/src/Traits/FileMapperTrait.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace Drupal\metastore\Traits;
+
+use Drupal\metastore\ResourceMapper;
+
+/**
+ * ResourceMapper Trait.
+ */
+trait FileMapperTrait {
```
dharizza: Thought we were renaming to ResourceMapper.

modules/metastore/tests/src/NodeWrapper/DataTest.php
```diff
@@ -0,0 +1,42 @@
...
+use MockChain\Chain;
+use PHPUnit\Framework\TestCase;
+
+/**
+ *
+ */
+class DataTest extends TestCase {
+
+  /**
+   *
```
dharizza, line 17: Empty comments.

modules/metastore/tests/src/files/README.txt
```diff
@@ -0,0 +1 @@
+In an attempt at making some tests more sociable and realistic, we are reducing some mocking and trying to use real classes. One of this examples is the FileMapper tests in which many Drupal classes (StreamWrapper, FileSystem, etc) are used. On of this clsses requires a folder to exist, and that is why this empty folder named file exists. The code that checks for the folder is in Drupal's LocalStream class.
```
dharizza: The are some typos here.

dharizza (commented):
I left some comments in the code, they are not blockers but i think it is good to take a look.
Also, there is a file in modules/common/tests/files/hello.txt that is empty, do we really need it?

dharizza (approved):

PR #3124, opened 2020-06-09
Use node storage over entity type manager

thierrydallacroce:
Whenever possible and where intended to be used with `getStorage('node')` (i.e. not in metastore_search), `EntityTypeManager` can be replaced with `NodeStorage` to get us closer to what we actually need, using less custom code hence less mocking in our unit tests.

### DIFF
--- a/modules/common/common.services.yml
+++ b/modules/common/common.services.yml
@@ -9,3 +9,7 @@ services:
     arguments:
       - '@module_handler'
       - '@request_stack'
+  dkan.common.node_storage:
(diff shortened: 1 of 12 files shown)


thierrydallacroce:
@fmizzell @dafeder Please let me know if the following CI error is something I can do anything about within this PR or if it's external.

`Not enough arguments (missing: "code_climate_reporter_id"). `

Thanks in advance.

thierrydallacroce:
Closing and re-opening a PR for the same code, but originating from this repo.

PR #3141, opened 2020-06-23
LoggerTrait

thierrydallacroce:
Minor clean-up in a couple classes using LoggerTrait where its default value or functionality is not being taken advantage of, and another one where it is not being used.


thierrydallacroce:
Not entirely understanding Code Climate's code diff percentage considering 554821c does not add new code, only removes some use-statements and redundant function parameters.

thierrydallacroce:
* Add conditional wrappers for notice, warning and error functions
* Add unit testing

PR #3144, opened 2020-06-26
Remove unneeded distribution schema fields

### LINKED ISSUE #3143: Clean up dataset json schema file
## User story
When I manually add a dataset, I should not see any fields that don't require human data entry, so that I'm not confused

## Acceptance criteria
When I manually add a dataset
I should not see a distribution identifier field 
I should not see a media type field

## Details/Tasks
- remove distribution identifier
- remove file size
- add suggested list for format and media type
- hide media type

janette:
fixes #3143 

## Changes
- Removed identifier field from Publisher
- Removed identifier field from Distributions
- Removed Media Size field from Distributions
- Removed Release Date field from Distributions
- Removed Publisher description and image fields, we should not set these values per dataset, we need a global approach here, possibly load select options on the Publisher name field from a json file.
- Set Media Type field to hidden (future plan is to auto-populate)
- Adds suggestions to the Format field, works like an autocomplete


susisiman:
@fmizzell could you review this?

PR #3150, opened 2020-07-01
Add missing punctuation for the correct short form of exempli gratia

tgurr:


kimwdavidson:
incorporated into #3144

PR #3151, opened 2020-07-01
Add missing punctuation

tgurr:
Not sure about the "Distribution identifier." though, please advise if I should update the pull request to rather exclude it.


kimwdavidson:
incorporated into #3144

PR #3155, opened 2020-07-09
Revise Schema files

janette:


janette:
duplicate of https://github.com/GetDKAN/dkan/pull/3144

PR #3157, opened 2020-07-14
Make frontend module work with Gatsby or Create React App

### LINKED ISSUE #3097: Make front end work with React App
## User Story
As a developer with a lot of datasets in my catalog, I want to use a create-react-app frontend, so I can update a page without having to rebuild and make changes quickly.

## Acceptance Criteria
- [ ] Both dynamicly- and staticly-generated routing is supported
- [ ] Frontend module detects which method by looking at how many files there are
- [ ] If staticly-generated (many files), a Drupal route is created for each static page
- [ ] If dynamically-generated (only one file exists), all Drupal routes (except admin/*) are redirected to that file

## Details/Tasks
- [ ] make DKTL choose between gatsby and react app (captured in GetDKAN/dkan-tools#227)

Note: This is essentially a quick-fix to switch from Gatsby to create-react-app, we expect to deprecate in the future

dgading:
Create new routing system for Create React App frontend that doesn't use the same dynamic routing system as the Gatsby version. 

QA Steps:
- `dktl make --frontend --branch=react-app-module`
- `dktl install --demo` This will install the frontend module and also install the base config included with the module that is setup for Gatsby (data-catalog-frontend)
- Site should be working
- In the docroot folder of the site, `git clone https://github.com/GetDKAN/data-catalog-react frontend` to add the React app to `docroot/frontend`
- cd into the frontend app and run `yarn install && yarn build` or `npm install && npm build`
- Go to: `/admin/config/development/configuration/single/import` and import the following with the label of `frontend.config`
```
static: False
build_folder: '/build'
frontend_path: '/frontend'
routes:
  - home,/home
  - about,/about
  - api,/api
  - dataset,/dataset/{id}
  - datasetapi,/dataset/{id}/api
  - search,/search
  - publishers,/publishers 
```
- `dktl drush cr`
- Site should now be serving the React app.

If not merged yet, use `git clone --single-branch --branch add-env-files --depth 1 https://github.com/GetDKAN/data-catalog-react.git` to have .env files setup so build works correctly. The site should still load without being a dynamic React app, but the calls on the homepage, dataset, and search will fail as the axios call may be wrong.


dgading:
In the react-app package.json, the following line needs to be added until it is in the base repo:
`"homepage": "frontend/build"`

This will tell it to load at the correct path when css/js files are called.

dgading:
This update adds a new `config/install` file to the frontend module. To make use of existing code, the routes for a non static site are now set using the install file. This can be updated currently with a config edit and possibly in the future a config form. Also in the config file is a string for build folder and frontend path. The frontend path is the folder in the docroot where the frontend lives, like `data-catalog-frontend` or `frontend`. The build folder is the location of the index.html file that runs the site. For Gatsby this is the public folder, but for Create React App it is the build folder as public is for global assets. 

This update should allow for easier changes to kinds of frontend tools used, as it removes most hardcoded paths and routes from the module code.

dharizza (approved):

PR #3162, opened 2020-07-20
Fix to show 'Add dataset' button when no results

janette:
## QA steps
- build without demo data or delete all datasets
- confirm that you still see the 'Add new dataset' button at `admin/content/datasets`


rhabbachi:
image

Clicking the button will keep the same URL `/admin/dkan/dataset` but will show the Dataset creation form.

rhabbachi (approved):
LGTM

PR #3179, opened 2020-08-12
Force updates on resources from manifest

dharizza:
## Description
The module dkan_periodic_updates provides functionality so that data publishers can upload a manifest file which is used for executing periodic imports of the specified resources. That manifest file is a csv file with format resource_id,frequency,file_url,import_to_datastore.

Currently the manifest supports specifying a frequency that can be daily, weekly or monthly, but sometimes there are urgent changes that need to be pulled into the site so the publishers need to execute the update earlier than specified in the manifest. This PR adds a new drush command to the module dkan_periodic_updates so we can force the update on all of the resources from the manifest (ignoring the frequency).

## Steps to test
- Go to admin/dkan/periodic-updates:
  - Click on the checkbox "Enable periodic updates" and click on the button "Save settings"
  - Upload a manifest that contains some test data (remember the headers should be resource_id,frequency,file_url,import_to_datastore make sure the resource_id is the UUID of an existing resource node, the frequency can be "weekly", "montly" or "daily" and import_to_datastore should have the value "Y" if you want it to be imported), click on "Add file"
- Go to admin/dkan/periodic-updates/status
  - No file should be imported at this point.
  - Run cron (via drush or cron, it doesn't matter), then reload the status page, all of the resources from the manifest should be updated
  - Run cron again, this time no resources should be updated (make sure the "Last update" date/time isn't changed)
  - Do some changes to the manifest (maybe point to another URL in some resource or just update one of the files).
  - [ ] Run the drush command `drush dkan-force-resources-update` (you can run it with -v for verbose mode).
  - [ ] Confirm the elements were updated, "Last update" column should be updated.

### DIFF
--- a/modules/dkan/dkan_periodic_updates/dkan_periodic_updates.drush.inc
+++ b/modules/dkan/dkan_periodic_updates/dkan_periodic_updates.drush.inc
@@ -1,0 +1,51 @@
+<?php
+
+/**
+ * @file
+ * Code for dkan_periodic_updates drush commands.
+ */
+
+/**
+ * Implements hook_drush_command().
+ */
+function dkan_periodic_updates_drush_command() {
+  $commands['dkan-force-resources-update'] = array(
+    'description' => 'Force resources from manifest to be updated.',
+    'aliases' => array('dfru'),
+  );
+
+  return $commands;
+}
+
+/**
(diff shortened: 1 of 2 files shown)


rhabbachi (approved):
LGTM

PR #3198, opened 2020-09-02
Fix undefined class constant when attempting drush dkan:datastore:drop

### LINKED ISSUE #3197: Error running drush dkan:datastore:drop
Attempting to run `drush dkan:datastore:drop <UUID>` caused the following error:

`[error]  Error: Undefined class constant 'DEREFERENCE_OUTPUT_REFERENCE_IDS' in Drupal\datastore\Drush->drop() (line 137)...`

thierrydallacroce:
Attempt to address #3197

### DIFF
--- a/modules/datastore/src/Drush.php
+++ b/modules/datastore/src/Drush.php
@@ -4,7 +4,6 @@ namespace Drupal\datastore;
 
 use Consolidation\OutputFormatters\StructuredData\RowsOfFields;
 use Consolidation\OutputFormatters\StructuredData\UnstructuredListData;
-use Drupal\metastore\ResourceMapper;
 use Drush\Commands\DrushCommands;
 
 /**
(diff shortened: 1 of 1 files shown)


modules/datastore/src/Drush.php
```diff
@@ -45,7 +45,7 @@ public function import($uuid, $deferred = FALSE) {
 
     try {
       // Load metadata with both identifier and data for this request.
-      drupal_static('metastore_dereference_method', Dereferencer::DEREFERENCE_OUTPUT_REFERENCE_IDS);
+      drupal_static('metastore_dereference_method', ResourceMapper::DEREFERENCE_YES);
```
fmizzell: @thierrydallacroce I believe we can simply remove these calls to drupal_static. Dereferencing effectively always happens.

fmizzell (changes requested):

modules/datastore/src/Drush.php
thierrydallacroce: Done in de7f9c0.

PR #3213, opened 2020-09-21
JSON endpoint for datastore queries

dafeder:
Will create a new query action in main datastore service/api, that accepts a JSON version of a query object. Will also support additional select operators.

Update: now working: or, comparison, between operators.

Try:

```http
POST /api/1/datastore/query/2ab8b036-689e-5f28-831f-46808950ce39 HTTP/1.1
Host: datastorebool.localtest.me
Content-Type: text/plain

{
    "orConditions": [
        {
            "property":"tax",
            "value":["0.20","0.42"],
            "operator":"BETWEEN"
        },
        {
            "property":"state_abbreviation",
            "value": "CO",
        }
    ],
    "showDbColumns":true,
    "count":true
}
```


modules/common/src/Storage/AbstractDatabaseTable.php
```diff
@@ -203,16 +207,8 @@ public function count(): int {
...
-
-    $this->setQueryConditions($db_query, $query);
-    $this->setQueryOrderBy($db_query, $query);
-    $this->setQueryLimitAndOffset($db_query, $query);
-
-    if ($query->count) {
-      $db_query = $db_query->countQuery();
-    }
+    $query->collection = $this->getTableName();
+    $db_query = SelectFactory::create($query, $this->connection);
```
brianseek, line 211: @dafeder Somewhere in the selectFactory the table alias isn't getting passed to the the SQL select. Lets pick this up again in a co-working session

grugnog:

PR #3228, opened 2020-10-02
Dashboard

fmizzell:


rhabbachi:
**Interface**
- Feels repetitive to have a different table solely for the status, could be a column in the 2nd table
image
- Column headers not sticky.
- Table would be more readable with alternating row background.
- Missing more information about the failure.
- In the example I used, `testy` harvest have all it's datasets marked as failed by the global status is marked as success.
image
- Would be more readable with relative time (X time ago) for dates within couple week / months.
image
- Can't access previous runs reports to compare and pinpoint when a dateset started failing. Not sure we keep previous runs reports in the db.

**Code**
- Getting a PHP exception after a harvest run timed out during a file fetch.
image
image


Attached db snapshot with all the test harvest plans.
[dashboard-error.sql.gz](https://github.com/GetDKAN/dkan/files/5458043/dashboard-error.sql.gz)

janette:
- I do not care for "X time ago" values at all, put my vote on full date value like it is. 
- +1 on moving the dataset harvest status into the other table with the ID, title, modified, resources (what is that one? count of distributions?)
- Why are the Title and Modified Dates columns empty?
- For the harvest status, classic has: Not started, In progress, Finished, and Finished with errors. Are those the same in v2?
- Can we get the Processed, Updated, Created info in that table?
- Classic has an Errors tab the will then list each dataset or distribution error, can we log that too.
- Wondering what a harvest of 1000 datasets would look like, does the table have a pager once we hit a certain #? Sorting?

harvest_error_log

harvest_event_log

rhabbachi:
> Wondering what a harvest of 1000 datasets would look like, does the table have a pager once we hit a certain #? Sorting?

If we set a pager, we need a search box to do lookups :smile: .

modules/harvest/src/Drush.php
```diff
@@ -134,7 +134,8 @@ public function run($id) {
   public function runAll() {
 
     $ids = $this->harvestService
-      ->getAllHarvestIds();;
+      ->getAllHarvestIds();
+    ;
```
thierrydallacroce: Extra semi-colon

modules/harvest/tests/src/Storage/DatabaseTableFactoryTest.php
```diff
@@ -18,7 +18,7 @@ class DatabaseTableFactoryTest extends UnitTestCase {
    */
   public function test() {
     $factory = new DatabaseTableFactory($this->getConnection());
-    $this->assertNotNull($factory->getInstance('blah', []));
+    $this->assertNotNULL($factory->getInstance('blah', []));
```
thierrydallacroce: Not sure why the casing of this function was changed

modules/metastore/src/LifeCycle/Data.php
```diff
@@ -80,6 +80,7 @@ protected function distributionLoad() {
 
     if (isset($downloadUrl) && !filter_var($downloadUrl, FILTER_VALIDATE_URL)) {
       $resourceIdentifier = $downloadUrl;
+      $ref = NULL; $original = NULL;
```
thierrydallacroce, line 83: My guess is you added these because Code Climate thinks these are undefined...

modules/metastore/tests/src/WebServiceApiTest.php
```diff
@@ -78,7 +79,8 @@ public function testGetResourcesException() {
    *
    */
   public function testGetException() {
-    $mockChain = $this->getCommonMockChain();;
+    $mockChain = $this->getCommonMockChain();
+    ;
```
thierrydallacroce: Extra semi colon.

modules/metastore/tests/src/WebServiceApiTest.php
```diff
@@ -66,7 +66,8 @@ public function testGetResources() {
    *
    */
   public function testGetResourcesException() {
-    $mockChain = $this->getCommonMockChain();;
+    $mockChain = $this->getCommonMockChain();
+    ;
```
thierrydallacroce: Extra semi colon.

modules/metastore/tests/src/ResourceMapperTest.php
```diff
@@ -64,7 +64,7 @@ public function test() {
       $mapper->get($resource1->getIdentifier(), 'local_url', $resource1->getVersion())
         ->getFilePath()
     );
-    $this->assertNull($mapper->get($resource1v2->getIdentifier(), 'local_url', $resource1v2->getVersion()));
+    $this->assertNULL($mapper->get($resource1v2->getIdentifier(), 'local_url', $resource1v2->getVersion()));
```
thierrydallacroce: Casing again

modules/metastore/src/WebServiceApiDocs.php
```diff
@@ -170,7 +170,8 @@ private function modifyDatasetEndpoint($path, $operations, $identifier) {
       return [$path, $operations];
     }
 
-    $newPath = str_replace("{identifier}", $identifier, $path);;
+    $newPath = str_replace("{identifier}", $identifier, $path);
+    ;
```
thierrydallacroce: Extra semi-colon.

modules/metastore/src/Reference/Dereferencer.php
```diff
@@ -88,6 +91,7 @@ private function dereferenceProperty(string $property_id, $uuid) {
   private function dereferenceMultiple(string $property_id, array $uuids) : array {
     $result = [];
     $reference = [];
+    $ref = NULL; $actual = NULL;
```
thierrydallacroce, line 94: Code climate, argh!

modules/metastore/src/Reference/Dereferencer.php
```diff
@@ -33,8 +32,12 @@ public function __construct(ConfigFactoryInterface $configService, NodeStorageIn
    * @return mixed
    *   Modified json metadata object.
    */
-  public function dereference(stdClass $data) {
+  public function dereference($data) {
+    if (!is_object($data)) {
+      throw new \Exception("data must be an object.");
+    }
     // Cycle through the dataset properties we seek to dereference.
+    $ref = NULL; $actual = NULL;
```
thierrydallacroce, line 40: Code climate, argh!

modules/datastore/tests/src/Storage/DatabaseTableTest.php
```diff
@@ -253,7 +254,7 @@ public function testDestroy() {
   /**
    *
    */
-  public function testPrepareDataJsonDecodeNull() {
+  public function testPrepareDataJsonDecodeNULL() {
```
thierrydallacroce: Weird casing again. Maybe it stemmed from doing a replace-all on `Null` to `NULL`

modules/datastore/tests/data/importer.json
```diff
@@ -3,7 +3,7 @@
     "result": {
         "status": "done",
         "data": "{\"recordNumber\":2970,\"chunksProcessed\":4572}",
-        "error": null
+        "error": NULL
     },
```
thierrydallacroce: Seems strange, I'm pretty sure in JSON I've always seen `null` lowercase.

modules/datastore/tests/data/filefetcher.json
```diff
@@ -3,7 +3,7 @@
     "result": {
         "status":"done",
         "data":"{\"source\":\"modules\\\/custom\\\/datastore\\\/tests\\\/data\\\/countries.csv\",\"total_bytes_copied\":150,\"temporary\":false,\"destination\":\"modules\\\/custom\\\/datastore\\\/tests\\\/data\\\/countries.csv\",\"total_bytes\":150}",
-        "error":null
+        "error":NULL
```
thierrydallacroce: Seems strange, I'm pretty sure in JSON I've always seen `null` lowercase.

thierrydallacroce (approved):
Not entirely sure about all the JS files with new spaces.

PR #3239, opened 2020-10-13
Add a dataset publication event

thierrydallacroce:


modules/metastore/src/Storage/Data.php
```diff
@@ -16,6 +18,8 @@
  */
 class Data implements StorerInterface, RetrieverInterface, BulkRetrieverInterface, RemoverInterface {
 
+  const EVENT_DATASET_PUBLICATION = 'dkan_metastore_dataset_publication';
```
fmizzell, line 21: @thierrydallacroce This event is happening in the data class. Is it appropriate to call it "dataset publication"?

PR #3244, opened 2020-10-26
Periodic updates: Add the ability to force import in the next cron execution

apupiales:


modules/dkan/dkan_periodic_updates/dkan_periodic_updates.module
```diff
@@ -408,6 +416,91 @@ function dkan_periodic_updates_state() {
...
+  if ($active) {
+    $file = variable_get('dkan_periodic_updates_manifest');
+    $file = file_load($file);
+    if ($file) {
+      $handle = fopen($file->uri, 'r');
+      $manifest_options = [];
+      $headers = fgetcsv($handle, 0, ',');
+      while (($data = fgetcsv($handle, 0, ",")) !== FALSE) {
+        $resource_data = array_combine($headers, $data);
+        if (!empty(variable_get('dkan_periodic_updates_' . $resource_data['resource_id'], ''))) {
```
dharizza, line 439: On a fresh install when no resource has been updated this looks a little bit broken because it shows the dropdown but without options, I think we can choose if we want to:
1- hide the form and say there is nothing to reset because no resource has been updated yet, or
2- change the condition in order to load all the options.

I'll leave this up to you.

modules/dkan/dkan_periodic_updates/dkan_periodic_updates.module
```diff
@@ -408,6 +416,91 @@ function dkan_periodic_updates_state() {
...
+          $manifest_options[$resource_data['resource_id']] = $resource_data['file_url'];
```
dharizza, line 440: Can we make the options a little bit more verbose? Right now it is just the url, maybe we can have a string with the url and the resource ID? Though, it is up to you if we do this, we can discuss in scrum.

modules/dkan/dkan_periodic_updates/dkan_periodic_updates.module
```diff
@@ -408,6 +416,91 @@ function dkan_periodic_updates_state() {
...
+ * Submit function for periodic updates reset form.
+ *
+ * @param $form
+ * @param $form_state
+ */
+function dkan_periodic_updates_reset_form_submit($form, &$form_state) {
+  if (empty($form_state['values']['resource_id'])) {
+    form_set_error('resource_id', t('You must select a resource'));
+  }
+  variable_set('dkan_periodic_updates_' . $form_state['values']['resource_id'], '');
```
dharizza, line 500: As filling the field is mandatory, I think we should add the required indication to it.

modules/dkan/dkan_periodic_updates/dkan_periodic_updates.module
dharizza, line 439: On a fresh install when no resource has been updated this looks a little bit broken because it shows the dropdown but without options, I think we can choose if we want to:
1- hide the form and say there is nothing to reset because no resource has been updated yet, or
2- change the condition in order to load all the options.

I'll leave this up to you.

dharizza (changes requested):

modules/dkan/dkan_periodic_updates/dkan_periodic_updates.module
apupiales: Change included

modules/dkan/dkan_periodic_updates/dkan_periodic_updates.module
apupiales: Option 1 included

modules/dkan/dkan_periodic_updates/dkan_periodic_updates.module
apupiales: As we talk, is easier for the user to look by url and adding the resource ID would make the text too long.

dharizza (approved):

PR #3272, opened 2020-11-20
Add metastore entity module

janette:
## QA Steps

- [ ] `dktl drush en metastore_entity -y`
- [ ] go to `metadata/add` url and create an entity
- [ ] navigate to `admin/structure/metadata/list`


grugnog:

PR #3295, opened 2020-12-21
Bringing harvest and datastore back.

fmizzell:


docs_assets/pages/3.2_datastore.md
```diff
@@ -0,0 +1,14 @@
+@page datastore Datastore
+
+The DKAN **Datastore** module provides integration with the [datastore](https://github.com/GetDKAN/datastore) library. With it you are able to parse CSV files and save the data in database tables. This allows users to query the data through a public API. Note that the CSV must be in UTF-8 format to parse correctly.
+
+When a dataset is added to the catalog, the mimetype value of each distribution object is checked to determine if the distribution is a CSV file or not. CSV distributions will be added to the import queue to be parsed and imported into the datastore.
+
+The import will happen in batches during cron runs. The import will itterate over the file, if it does not finish, the job will be returned to the queue. Be sure to have cron running on a regular basis so that large file imports will complete. If there is a change to the metadata it will trigger a new import to the datastore.
+
+You can manually import file data into the datastore via drush with the identifier of the distribution. Use the [API](guide-dataset-api.html#identifiers) to get the identifier of the file you want to import.
```
janette, line 9: Remove the link, what is the best way to find the identifier to use with the Drush command?

fmizzell: sadly, it is the api. We should probably create some metastore drush commands to get this kind of info.

janette: So the link would be `/api/1/metastore/schemas/dataset/items/?show-reference-ids` and look for the distribution.0.identifier value?

PR #3335, opened 2021-01-27
Make identifier use dkan_uuid and replace metastore_form.

### LINKED ISSUE #3311: UUID widget
Make the identifier field use the UUID widget. 

## Background ##
We hide the identifier field from the data publisher, because it already should have a Drupal UUID as an identifier. The widget hides the field and adds the UUID. We need to make the identifier use that widget.

But this currently breaks tests, because the React JSON schema form doesn't recognize the widget. 

## Tasks ##
- Remove React JSON schema form
- Make identifier use UUID widget

## Acceptance Criteria ##
- tests pass
- when you add a new data node
- then you do not see an identifier field
- and when you save
- then you see a uuid in the URL

dharizza:
Fixes #3311 

Removed the metadata_form functionality in favor of json_form_widget. The module is still there but should be completely deleted for the next DKAN release.
Made the field_json_metadata use the JSON Form Widget and added default config for it.
Hide the identifier field.

## Preparation steps
You can try the QA steps on the following scenarios:
1. New dkan installation.
2. You have dkan already installed, in that case you need to:
  - [ ] dktl drush updb -y
  - [ ] dktl drush cr

## QA Steps
- [ ] Go to node/add/data, you should see the full form populated with fields for each element from the schema.
- [ ] The field identifier should not be visible.
- [ ] You should be able to submit the form without errors.
- [ ] The module metadata_form should be disabled and shown as deprecated.


modules/metastore/modules/metastore_admin/metastore_admin.install
```diff
@@ -25,3 +25,10 @@ function metastore_admin_uninstall() {
   $config->save(TRUE);
   drupal_flush_all_caches();
 }
+
```
dafeder, line 28: @dharizza I think this needs to enable the new module as well? I tried the upgrade QA path and got:
```
[error]  Drupal\Core\Config\ConfigImporterException: There were errors validating the config synchronization.
Configuration <em class="placeholder">core.entity_form_display.node.data.default</em> depends on the <em class="placeholder">JSON Form Widget</em> module that will not be installed after import. in Drupal\Core\Config\ConfigImporter->validate() (line 748 of /var/www/docroot/core/lib/Drupal/Core/Config/ConfigImporter.php). 
```

dafeder (changes requested):
@dharizza looks good on a new site but is there a good upgrade path here? I see that the update hook disables the old form module but does it enable the new one? The form widget is changed in the default config but will that be applied to an existing site?

modules/metastore/modules/metastore_admin/metastore_admin.install
dharizza: Good catch, just fixed it. Thank you!

dafeder:
Sorry, the PR that never ends. The updatedb seems to work, but I get this output:

image

Also, if i go to the datasets page and use the new dataset button, I get page not found. Starting to think we should leave the update hook out of this PR and just provide instructions to switch manually, and move the update hook to another issue?

dafeder:
Looks like this error does not affect functionality, we're going to merge and just warn people in release notes.

PR #3336, opened 2021-01-29
Fix email validation on json form widget

### LINKED ISSUE #3312: Fix email validation on json form widget
## Background ##
The schema requires the format "mailto:(email address)"

But most users will expect to just put their email address in without "mailto:" in front of it. We want to add "mailto:" for them. 

We also need to ensure validate that entered emails include .domain.

## Tasks ##
- support option in schema UI
- prepend to the value

## Acceptance Criteria ##
- When I create dataset
- And go to the contact section
- And enter an email address in the format "name@service.whatever"
- And the rest of the form is filled out properly
- I don't see an error
- And I see my email display properly on the dataset

dharizza:
fixes #3312 

## Preparation steps
- [x] Enable the json_form_widget module.
- [x] Go to Structure -> Content types -> Data -> Manage form display, select the widget "JSON Form" for the field "JSON Metadata", make the schema be "dataset" and save the configuration.
- [ ] In the codebase go to the JSON schema UI for dataset `dataset.ui.json` and add a new entry with the following:

```
"identifier": {
    "ui:options": {
      "widget": "dkan_uuid"
    }
  }
```

## QA Steps

- [x] Go to node/add/data
- [x] Fill up the form
- [x] In the field Email (from section Project Open Data Contactpoint VCARD) try the following:
  - [x] Leave the field empty and submit the form, you should get an error saying that field should not be empty
  - [x] Type `test` and submit the form, you should get an error saying that specific field is wrong formatted
  - [x] Type `test@test` and submit the form, you should get an error saying that specific field is wrong formatted
  - [x] Type `test@test.com` and submit the form, the node should be saved correctly without any issues.

### DIFF
--- a/modules/json_form_widget/json_form_widget.services.yml
+++ b/modules/json_form_widget/json_form_widget.services.yml
@@ -9,6 +9,7 @@ services:
     class: \Drupal\json_form_widget\ValueHandler
   json_form.string_helper:
     class: \Drupal\json_form_widget\StringHelper
+    arguments: ['@email.validator']
   json_form.object_helper:
     class: \Drupal\json_form_widget\ObjectHelper
   json_form.array_helper:
--- a/modules/json_form_widget/src/StringHelper.php
+++ b/modules/json_form_widget/src/StringHelper.php
@@ -2,14 +2,19 @@
 
 namespace Drupal\json_form_widget;
 
+use Drupal\Component\Utility\EmailValidator;
+use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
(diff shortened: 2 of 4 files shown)


janette:
@dharizza if I enter `mailto:test@test.com` I get the "Please enter an email address" error, we need that as an allowable pattern.

janette:
After further discussion, a 'mailto:...' string is not a valid email and we prefer to keep the email validation. We can add the mailto prefix on the frontend if we want to keep that feature. 

Future code may be needed once we get into one time migrations/harvesting of datasets that have the email properties in this format.

PR #3350, opened 2021-02-09
Make our facets endpoint more granular and performant

### LINKED ISSUE #3346: Make our facets endpoint more granular and performant
Our current methodology to give the user information about facets is very inefficient. For every possible facet on a site we query the SeachAPI system for information on how many datasets match that facet (keyword, topic, etc). For an "empty" search (we have to return all of the possible facets anyways), the current approach is the best that we know how to do. In situation in which some filtering is happening, we can optimize the facets information procedure by looking at matching records instead of all possible facets. 

This issue will implement this alternative way of calculating facets, and create some logic to use the most effective method in a given context.

fmizzell:
#3346


modules/metastore/modules/metastore_search/tests/src/Functional/SearchTest.php
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * @package Drupal\Tests\metastore_search\Functional
+ * @group metastore_search
+ */
+class SearchTest extends ExistingSiteBase {
+
+  /**
+   *
+   */
+  public function testDushCommands() {
```
thierrydallacroce: Typo `Dush`

modules/metastore/modules/metastore_search/tests/src/Functional/SearchTest.php
```diff
@@ -0,0 +1,60 @@
...
+    $output = NULL;
+    $return = NULL;
+    exec('drush dkan:metastore-search:rebuild-tracker', $output, $return);
+    $this->assertEquals(0, $return);
```
thierrydallacroce, line 25: Why is there code after this non-expectException assertion?

thierrydallacroce (changes requested):

modules/metastore/modules/metastore_search/tests/src/Functional/SearchTest.php
fmizzell: @thierrydallacroce did you read exec() as expectException() ?

PR #3351, opened 2021-02-09
Eliminate drush dependency

dafeder:
It doesn't really make sense to have drush as a module dependency for DKAN, it should be managed at the project level. This removes it from composer.json

## Merging instructions

1. Merge GetDKAN/dkan-tools#287
2. New DKAN tools release
3. Change dktl version to new release in this branch's circleci config
4. Merge if all still passes


janette:
This was done in https://github.com/GetDKAN/dkan/pull/3412

PR #3341, opened 2021-02-05
add check for uuid

janette:
part one of  #3328

PR #3361, opened 2021-02-17
fix: catch TypeErrors on dkan:datastore:drop-all

rhabbachi:
Fixes an issue when the datastore registered items are corrupt and contain entity with empty identifiers.

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

- Get the datastore to a corrupt state (need a consistant way of doing this, but probably by harvesting a harvest plan that adds entities to the datastore, and then revert the harvest).
- The try dkan:datastore:drop-all
- An error should show up for entities with empty UUIDs.

### DIFF
--- a/modules/datastore/src/Drush.php
+++ b/modules/datastore/src/Drush.php
@@ -183,6 +183,14 @@ class Drush extends DrushCommands {
         $this->logger->debug($e->getMessage());
         continue;
       }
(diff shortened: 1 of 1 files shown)


janette:
@rhabbachi ha this is a nicer solution than https://github.com/GetDKAN/dkan/pull/3341

PR #3393, opened 2021-03-08
Issue #3392 - change chunked import from fread to fgets

### LINKED ISSUE #3392: DKAN Datastore Simple Import produces warning messages (7.x)
## Describe the bug
The DKAN Datastore Simple import may produce (incorrect) warning messages like "Source file is not in "UTF-8" encoding".
The Simple import does a chunked import, chunking the file into 32 Byte parts. The import does an UTF-8 check with mb_check_encoding() on these 32 byte chunks. Now it is possible that the chunk splits an UTF8 char in the middle. If this is happening the UTF-8 check will fail and warning "Source file is not in "UTF-8" encoding" is given.
The import itself will not fail, because beside checking the UTF-8 and giving the warning the data is processed unaltered.

I think, it does not make sense to perform an UTF-8 check on the chunked parts of the file in the import.

Maybe read the file line by line and do UTF-8 check on each line. Or do the UTF-8 check on the whole file?

This is the code line where the check happens: https://github.com/GetDKAN/dkan/blob/0342af1b8aa91bc1ec3ff347d00be9aa3cabf0f5/modules/dkan/dkan_datastore/modules/dkan_datastore_simple_import/src/SimpleImport.php#L54

stefan-korn:
fixes #3392

## QA Steps

- do a simple import in the DKAN datastore

### DIFF
--- a/modules/dkan/dkan_datastore/modules/dkan_datastore_simple_import/src/SimpleImport.php
+++ b/modules/dkan/dkan_datastore/modules/dkan_datastore_simple_import/src/SimpleImport.php
@@ -44,7 +44,7 @@ class SimpleImport extends Manager {
 
     $finished = TRUE;
     $interrupt = $this->getInterrupt();
-    while ($chunk = fread($h, 32)) {
+    while ($chunk = fgets($h)) {
       if ($interrupt) {
         $finished = FALSE;
         break;


dafeder:
Closing - no new code goes into 7.x

PR #3423, opened 2021-04-06
Removed SAE

### LINKED ISSUE #3366: Controller: Remove SAE engine
The SAE engine performs JSON Schema validation check and if it passes, it gives the JSON to the storage mechanism. It also includes some logic about different actions.

We don't really need this anymore because all the meaningful logic is available from the RootedJsonData library. So we're going to remove SAE completely and have controllers do their own validation with RootedJsonData methods.

## Acceptance criteria

- SAE no longer exists
- Current tests still pass

**Note:** Eventually we want to replace a lot of the methods that take or return strings to use RootedJsonData objects. That will be a separate ticket.

Estimate: 1 day

klimp-drupal:
fixes [GetDKAN/dkan/issues/3366]

Replaces SAE JSON validation against schema with the validation provided by [RootedJsonData plugin](https://github.com/GetDKAN/RootedJsonData) in the metastore controller, service and storage

- [ ] All tests pass
- [ ] All API requests work the same way as before the change
- [ ] JSON data is validated
- [ ] `getdkan/sae` is removed

## QA Steps

- [ ] Run all the tests
- [ ] Run various API requests


modules/metastore/src/Service.php
```diff
@@ -158,37 +185,48 @@ public function get($schema_id, $identifier): string {
...
   public function getResources($schema_id, $identifier): array {
-    $json = $this->getEngine($schema_id)
-      ->get($identifier);
-    $data = json_decode($json);
+    $json_string = $this->getStorage($schema_id)->retrieve($identifier);
+    $data = $this->validMetadataFactory->get($schema_id, $json_string);
+
     /* @todo decouple from POD. */
-    $resources = $data->distribution;
+    $resources = $data->{"$.distribution"};;
```
dharizza: Extra ;

modules/metastore/src/Storage/Data.php
```diff
@@ -309,6 +310,7 @@ private function getType() {
    *   Filtered output.
    */
   private function filterHtml($input) {
+    // TODO: find out if we still need it.
```
dharizza, line 313: Should we find out as part of this work? Or is this really for later?

dharizza (commented):
I left a couple of comments, but just for an extra ; and a comment in the filterHtml function to find out if we still need it, but other than that this is looking good. The issues I had in the form were because of dependencies (again), after a full rebuild of them the form is working correctly. So let's find out about the filterHtml and merge this.

dharizza (approved):

PR #3434, opened 2021-04-14
RootedJsonData

### LINKED ISSUE #3372: Use RootedJsonData for validation in metastore service
Code to be adjusted:
- [x] metastore_search. Facets adjusted to RootedJsonData
- [x] Drupal\Tests\metastore_search\Unit\SearchTest
- [x] Drupal\Tests\metastore_search\Unit\Plugin\search_api\datasource\DkanDatasetTest::test
- [x] Drupal\Tests\datastore\Service\ResourcePurgerTest
- [x] Drupal\Tests\harvest\Load\DatasetTest
- [x] Drupal\Tests\json_form_widget\Unit\SchemaUiHandlerTest
- [ ] Consider using RootedJsonData in Storage 'retrieve'-like methods

Estimate: 1 day

klimp-drupal:
fixes [GetDKAN/dkan/issues/3372]

Facilitates using [RootedJsonData objects](https://github.com/GetDKAN/RootedJsonData) instead of a JSON string in the metastore controller, service and storage.

- [ ] All tests pass
- [ ] All API requests work the same way as before the change
- [ ] JSON data is validated

## QA Steps

- [ ] Run all the tests
- [ ] Run various API requests

### DIFF
--- a/modules/metastore/metastore.module
+++ b/modules/metastore/metastore.module
@@ -117,7 +117,7 @@ function metastore_form_alter(&$form, \Drupal\Core\Form\FormStateInterface $form
     // TODO: use an actual schema id.
     $object = \Drupal::service('dkan.metastore.rooted_json_data_wrapper')->createRootedJsonData(NULL, $json);
     $object = \Drupal\metastore\Service::removeReferences($object);
-    $form['field_json_metadata']['widget'][0]['value']['#default_value'] = $object->pretty();
+    $form['field_json_metadata']['widget'][0]['value']['#default_value'] = $object->__toString();
   }
 }
 
@@ -127,8 +127,8 @@ function metastore_entity_view_alter(array &$build, Drupal\Core\Entity\EntityInt
     if (empty($json)) {
       return;
     }
-    // TODO: use an actual schema id.
-    $object = \Drupal::service('dkan.metastore.rooted_json_data_wrapper')->createRootedJsonData(NULL, $json);
+    $schema_id = $entity->get('field_data_type')->getString();
+    $object = \Drupal::service('dkan.metastore.rooted_json_data_wrapper')->createRootedJsonData($schema_id, $json);
     $object = \Drupal\metastore\Service::removeReferences($object);
     $build['field_json_metadata'][0]['#context']['value'] = $object->pretty();
   }
--- a/modules/metastore/metastore.module
+++ b/modules/metastore/metastore.module
@@ -115,7 +115,7 @@ function metastore_form_alter(&$form, \Drupal\Core\Form\FormStateInterface $form
       return;
     }
     // TODO: use an actual schema id.
-    $object = \Drupal::service('dkan.metastore.rooted_json_data_wrapper')->createRootedJsonData(NULL, $json);
+    $object = \Drupal::service('dkan.metastore.valid_metadata')->get(NULL, $json);
     $object = \Drupal\metastore\Service::removeReferences($object);
     $form['field_json_metadata']['widget'][0]['value']['#default_value'] = $object->__toString();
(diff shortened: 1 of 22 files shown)


modules/metastore/metastore.module
```diff
@@ -114,9 +114,10 @@ function metastore_form_alter(&$form, \Drupal\Core\Form\FormStateInterface $form
     if (empty($json)) {
       return;
     }
-    $object = json_decode($json);
+    // TODO: use an actual schema id.
+    $object = \Drupal::service('dkan.metastore.rooted_json_data_wrapper')->createRootedJsonData(NULL, $json);
     $object = \Drupal\metastore\Service::removeReferences($object);
-    $form['field_json_metadata']['widget'][0]['value']['#default_value'] = json_encode($object);
+    $form['field_json_metadata']['widget'][0]['value']['#default_value'] = $object->pretty();
```
dafeder: Curious why you used pretty() here, just the regular string would be fine I'd think?

modules/metastore/metastore.module
```diff
@@ -114,9 +114,10 @@ function metastore_form_alter(&$form, \Drupal\Core\Form\FormStateInterface $form
     if (empty($json)) {
       return;
     }
-    $object = json_decode($json);
+    // TODO: use an actual schema id.
```
dafeder, line 117: What stopped you from using the schema id now?

modules/metastore/src/WebServiceApi.php
```diff
@@ -134,28 +134,24 @@ private function wantObjectWithReferences() {
   /**
    * Private.
    */
-  private function swapReferences($object) {
-    $array = (array) $object;
-    foreach ($array as $property => $value) {
+  private function swapReferences(RootedJsonData $object): RootedJsonData {
+    $no_schema_object = $this->service->getRootedJsonDataFactory()->createRootedJsonData(NULL, $object->pretty());
```
dafeder: Again, shouldn't need to use pretty() unless displaying the JSON to the user.

dafeder (commented):
A couple minor things here in the inline comments but overall this looks great. There is one thing I am still shaky on and that's the naming. RootedJsonData is a really annoying name I wish we'd done better on but in the end it's just a library. I think We're doing something more specific in this PR: creating permanently valid objects specifically from _metastore data_ that has a schema in the SchemaRetriever system. I think we can use naming conventions that are clearer here.

Maybe RootedJsonDataFactory could become ValidMetadataFactory? The service name could be `dkan.metastore.valid_metadata`. Open to other names but the current naming seems awkward to me. 

Also, I think the general factory pattern would be to have the main method just be called `get` as opposed to `createRootedJsonData`.

modules/metastore/src/WebServiceApi.php
klimp-drupal: replaced to __toString()

modules/metastore/metastore.module
klimp-drupal: I don't have the schema id at this stage.

In my latest commits I used `field_data_type` for the schema id here. However, I left a similar todo in the hook_fiorm_alter() above because I'm not sure if I have the entity object there - I couldn't pause it with a breakpoint to debug

modules/metastore/metastore.module
klimp-drupal: Replaced with __toString()

modules/metastore/metastore.module
dafeder: gotcha, ok

dafeder (approved):

dafeder:
Codeclimate is on the fritz, but merging.

PR #3435, opened 2021-04-14
GET endpoint for datastore query API

dafeder:
Using POST for any datastore query is a pain. This adds an equivalent GET endpoint for datastore queries that can accept the same JSON request types, but encoded for a query string. 

The new RequestParamNormalizer class can hopefully be used to apply this pattern to other parts of the API in the future.

Note: If RootedJsonData used justinrainbow instead of Opis for JSON validation, we could move the type casting logic into RootedJsonData and simplify the controller and normalizer logic from this PR. Would be a nice future improvement if it doesn't break anything RootedJsonData is doing elsewhere.

## QA Steps

- [ ] Stand up a site with sample data
- [ ] Get the distribution ID for bike lane data at api/1/metastore/schemas/dataset/items?show-reference-ids
- [ ] Load api/1/datastore/query/[ID]?results=true&conditions[0][property]=record_number&conditions[0][value]=2 in a browser

### DIFF
--- a/modules/common/src/Util/RequestParamNormalizer.php
+++ b/modules/common/src/Util/RequestParamNormalizer.php
@@ -1,0 +1,84 @@
+<?php
+
+namespace Drupal\common\Util;
+
+use JsonSchema\Validator;
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * Class DkanRequestNormalizer.
+ *
+ * This class provides public static functions to normalize different HTTP
+ * request types into valid JSON data. Provides type casting from a JSON Schema
+ * to avoid data loss from conversion between query string and JSON.
+ *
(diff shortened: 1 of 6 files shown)


modules/common/src/Util/RequestParamNormalizer.php
```diff
@@ -0,0 +1,84 @@
+<?php
+
+namespace Drupal\common\Util;
+
+use JsonSchema\Validator;
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * Class DkanRequestNormalizer.
```
fmizzell: I guess the name changed but not the docs

fmizzell (changes requested):

PR #3447, opened 2021-04-27
Date picker granularity

### LINKED ISSUE #3440: Support time granularity in date picker form widget
When a date has this format `2020-05-11T15:06:39.000Z` the date widget doesn't display the date at all.

Estimate: 0.5 day

dharizza:
fixes #3440 

## Description
- Added support for date time picker.
- Fixed bug where date pickers wouldn't show the right date if the incoming data included a time.

## Acceptance criteria
- When I harvest a dataset that has in the fields "issued" or "modified" dates with the following formats: 
  - [ ] `2020-05-11T15:06:39.000Z`
  - [ ] `2020-05-11 15:06:39.000`
  - [ ] `2020-05-09`
- And I go to the edit form for the harvested datasets
- Then the date picker for the corresponding field should display the right date.

### DIFF
--- a/cypress/integration/07_admin_dataset_json_form.spec.js
+++ b/cypress/integration/07_admin_dataset_json_form.spec.js
@@ -9,7 +9,7 @@ context('Admin dataset json form', () => {
         cy.get('#edit-field-json-metadata-0-value-title').should('have.attr', 'required', 'required')
         cy.get('#edit-field-json-metadata-0-value-description').should('have.attr', 'required', 'required')
         cy.get('#edit-field-json-metadata-0-value-accesslevel').should('have.attr', 'required', 'required')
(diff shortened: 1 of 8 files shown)


janette:
@dharizza this works perfectly. I was wondering, can the date formats match the POD structure? YYYY-MM-DD rather than mm/dd/yyyy?

dharizza:
@janette according to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date "The displayed date format will differ from the actual value — the displayed date is formatted based on the locale of the user's browser, but the parsed value is always formatted yyyy-mm-dd.", so the value we're storing is already matching the POD structure YYYY-MM-DD, the difference is when entering the data but that changes depending on the user locale browser configuration. Also it is not recommended to force the format in the default input type=date element (and there is no way to do so).

PR #3455, opened 2021-05-05
Use get_headers as fallback for the file content-type in order to trigger datastore import.

### LINKED ISSUE #3443: Bug: Datastore import fails when manually creating datasets
## Steps to reproduce
- Create a dataset at `node/add/data`
- Run `drush queue:run datastore_import`
- Confirm output is `Processed 0 items from the datastore_import queue in 0.01 sec.`
- Seems to be a missing step on node creation, only the source perspective is present in the dkan_metastore_resource_mapper table

issues

User Story:

Acceptance criteria: 

Estimate: 2 days

dharizza:
fixes #3443 

Datastore import was not triggered because we rely on the value of media type to see if the file is a csv, by default the new metadata form has the media type field hidden so the default value assumed was `text/plain`. In this PR we're using the get_headers() function as a fallback when no media type value is found.

## Steps to test

- Create a dataset at node/add/data with a resource (try both, file upload and link to remote)
- Run drush queue:run datastore_import
- [x] Confirm output is "Processed 1 items from the datastore_import queue in XXX sec."
- [x] In dkan_metastore_resource_mapper table we should get 3 perspectives for the same resource.
- [x] If you run `drush dkan:datastore:list`, you can see the import was correct.

### DIFF
--- a/modules/metastore/tests/src/Unit/Reference/ReferencerTest.php
+++ b/modules/metastore/tests/src/Unit/Reference/ReferencerTest.php
@@ -2,9 +2,16 @@
 
 namespace Drupal\Tests\metastore\Unit\Reference;
 
+use Drupal\Core\Config\ConfigFactoryInterface;
+use Drupal\Core\Config\ImmutableConfig;
+use Drupal\Core\Logger\LoggerChannelFactory;
 use Drupal\metastore\Reference\Referencer;
+use Drupal\metastore\ResourceMapper;
+use Drupal\node\NodeStorage;
 use MockChain\Chain;
+use MockChain\Options;
 use PHPUnit\Framework\TestCase;
+use stdClass;
 use Symfony\Component\DependencyInjection\Container;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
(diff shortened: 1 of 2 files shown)


modules/metastore/tests/src/Unit/Reference/ReferencerTest.php
```diff
@@ -25,4 +32,95 @@ public function testHostify() {
...
+      ->add(Container::class, 'get', $options)
+      ->add(RequestStack::class, 'getCurrentRequest', Request::class)
+      ->add(Request::class, 'getHost', 'test.test')
+      ->add(ResourceMapper::class, 'register', TRUE, 'resource');
+
+    $container = $container_chain->getMock();
+    \Drupal::setContainer($container);
+
+    $referencer = new Referencer($configService, $nodeStorage);
+    $data = $this->getData('https://dkan-default-content-files.s3.amazonaws.com/phpunit/district_centerpoints_small.csv');
```
dafeder, line 94: @dharizza can we do this with local test data files? Generally we don't use remote files in unit tests

dafeder (changes requested):
QA passed. See one comment about tests

modules/metastore/tests/src/Unit/Reference/ReferencerTest.php
dafeder: Nevermind, seeing that this isn't possible. I'm going to merge for now, but we might want to rethink these tests in the future.

PR #3464, opened 2021-05-09
Lifecycle abstraction

### LINKED ISSUE #3444: Make data lifecycle work with custom entity
Some of the data lifecycle hooks may be node-specific. Datasets and distributions are handled here... 

Looking at metadata wrapper and making sure all of our lifecycle logic goes through that instead of working through nodes directly

Most likely solution: convert lifecycle to overrideable service and a second lifecycle class to ship with custom entities. Couples lifecycle to storage.

Estimate: 2 days

dafeder:
This is a refactor of the "LifeCycle" code in DKAN. The LifeCycle class contains the logic that is used in Drupal's entity hooks, to make changes to the metadata at the time of save or load. To prepare for a move to a custom entity, we are abstracting out any code that is specific to a certain entity type, bundle or field name, and replacing these references with methods that are defined in an interface to be shared with future storage systems.

This PR should not actually change any behavior, but code review should confirm that metastore.module, and the new LifeCycle class (formerly LifeCycle\Data) have none of these specific references "hard-coded". It also contains the beginnings of a new more consistent class and interface naming scheme within the metastore module, to be moved further along in future related PRs.


modules/metastore/metastore.module
```diff
@@ -4,50 +4,42 @@
...
-  catch (DataNodeLifeCycleEntityValidationException $e) {
-  }
+  metastore_data_node_lifecycle($entity, "presave");
 }
 
 /**
- * Implements hook_ENTITY_TYPE_predelete().
+ * Implements hook_entity_predelete().
  */
 function metastore_node_predelete(EntityInterface $entity) {
```
dharizza: Shouldn't this be metastore_entity_predelete?

modules/metastore/src/Reference/Dereferencer.php
```diff
@@ -7,12 +7,17 @@
...
 /**
- * Dereferencer.
+ * Metastore dereferencer.
  */
 class Dereferencer {
   use HelperTrait;
   use LoggerTrait;
 
+  /**
+   * Node storate service.
```
dharizza, line 17: Typo.

dharizza (commented):
@dafeder This looks pretty much good to me. I tested the form functionality and it is still behaving correctly too. I only have questions in the metastore.module file because I see we still have the function metastore_node_predelete and I believe that one should be changed to metastore_entity_predelete. 
I also see there are multiple places where we have the field `field_json_metadata` hardcoded but in the metastore.module we're getting it through a function, should we be doing that in other places too or is it fine to have it hardcoded? 
Apart from that I think we're good to merge, and even these changes (if needed) can be done in other PR. I'll leave that up to you.

dharizza (approved):

PR #3469, opened 2021-05-12
Date range widget for temporal

### LINKED ISSUE #3439: Date range widget
As a data publisher when I go to the temporal field, I can see date pickers to choose the range in which my data applies.

Acceptance criteria:
- In the metadata form, I can use date pickers to select start and end date for the field "temporal" instead of having to enter the info in a textfield.

Estimate: 2 days

dharizza:
fixes #3439 

Added new element type for handling date ranges for the field temporal.

## QA Steps

- [ ] Go to node/add/data, the temporal field should have a date and time entry for the start date and end date.
- Try the following behaviors of the temporal field:
  - [ ] If you fill up only the start date or only the end date, it should trigger an error
  - [ ] If the start date is greater than the end date, it should trigger an error
  - [ ] If both start and end date are filled and the end date is greater than the start date, the node should be saved correctly.


dafeder:
@dharizza looks good but I was able to break it -- if you add a date range where one side has a time specified and the other doesn't, you get a server error on submission (it passes form validation but apparently something else breaks in the back end).

TypeError: Argument 2 passed to           
                                        Drupal\json_form_widget\Element\DateRang  
                                        e::validateInterval() must be an          
                                        instance of                               
                                        Drupal\Core\Datetime\DrupalDateTime,      
                                        array given, called in /var/www/dk

modules/json_form_widget/src/Element/DateRange.php
```diff
@@ -0,0 +1,208 @@
...
+        $date_format = $element['#date_date_element'] != 'none' ? static::getHtml5DateFormat($element) : '';
+        $time_format = $element['#date_time_element'] != 'none' ? static::getHtml5TimeFormat($element) : '';
+        $date_time_format = trim($date_format . ' ' . $time_format);
+
+        $start_time = static::getFormattedTime($input['start_date']['time']);
+        $start_input = $input['start_date']['date'] . ' ' . $start_time;
+        $start = DrupalDateTime::createFromFormat($date_time_format, $start_input, $element['#date_timezone']);
+        $start = $start->format('c', ['timezone' => 'UTC']);
+
+        $end_time = static::getFormattedTime($input['end_date']['time']);
```
klimp-drupal, line 80: This code seems to be the same as for the $start_time. I would move it into a separate method to avoid the repetition

modules/json_form_widget/src/Element/DateRange.php
```diff
@@ -0,0 +1,208 @@
...
+   * @param array $input
+   *   Array with keys for start_date and end_date.
+   * @param array $element
+   *   The element being validated.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The current state of the form.
+   */
+  public static function missingStartDate(array $input, array $element, FormStateInterface $form_state) {
+    if (empty($input['start_date']) && !empty($input['end_date'])) {
+      $form_state->setError($element['end_date'], t('Please enter a start date.'));
```
klimp-drupal: Seems it should be $element['start_date']

modules/json_form_widget/src/Element/DateRange.php
```diff
@@ -0,0 +1,208 @@
...
+class DateRange extends Datetime {
+
+  /**
+   * Get info.
+   */
+  public function getInfo() {
+    $class = get_class($this);
+    $date_format = '';
+    $time_format = '';
+    if (!defined('MAINTENANCE_MODE')) {
```
klimp-drupal, line 26: I'm not aware of the specifics of this work but why we don't want it to be working when the maintenance is on?

modules/json_form_widget/src/Element/DateRange.php
```diff
@@ -0,0 +1,208 @@
...
+      '#date_time_format' => $time_format,
+    ];
+  }
+
+  /**
+   * Callback for getting the value of the date range element.
+   */
+  public static function valueCallback(&$element, $input, FormStateInterface $form_state) {
+    $element += ['#date_timezone' => date_default_timezone_get()];
+    if ($input !== FALSE) {
```
klimp-drupal, line 67: I'd just return `if ($input == FALSE) return $input;` in the first line instead of putting the entire method under this condition.

modules/json_form_widget/src/Element/DateRange.php
```diff
@@ -0,0 +1,208 @@
...
+   * @param array $input
+   *   Array with keys for start_date and end_date.
+   * @param array $element
+   *   The element being validated.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The current state of the form.
+   */
+  public static function missingEndDate(array $input, array $element, FormStateInterface $form_state) {
+    if (!empty($input['start_date']) && empty($input['end_date'])) {
+      $form_state->setError($element['end_date'], t('Please enter an end date.'));
```
klimp-drupal, line 190: don't you have `$this->t()` available?

klimp-drupal (commented):
Added some minor comments

modules/json_form_widget/src/Element/DateRange.php
dharizza: It is because date formats cannot be loaded during install or update. The code Drupal DateTime element does it as well because of it.

modules/json_form_widget/src/Element/DateRange.php
dharizza: No, because it is a static function so we need to use t.

dharizza:
@klimp-drupal this is ready for review again. Thank you.

PR #3475, opened 2021-05-14
Update SAE branch

dafeder:
Just merge the latest changes to 2.x abstracting the lifecycle functionality into the SAE branch.


modules/metastore/src/NodeWrapper/Data.php
```diff
@@ -67,9 +58,7 @@ public function getIdentifier() {
    */
   public function getRawMetadata() {
     $this->fix();
-    if (isset($this->node->rawMetadata)) {
-      return json_decode($this->node->rawMetadata);
-    }
+    return $this->node->get('field_json_metadata')->getString();
```
klimp-drupal, line 61: Can we be sure it is raw metadata? We (used to) call `setMetadata()` to write the dereferenced version into the field

PR #3476, opened 2021-05-15
Abstract data storage in referencer

### LINKED ISSUE #3454: Use abstract DKAN storage with referencer
## Acceptance criteria:
- When we create datasets with a distribution and download URL, multiple nodes are created for the different data types with the right json data.
- The referencer will work as expected whether we're using the old nodes storage or the new custom entity.
- Do a thorough code review to confirm there is no reference to node-specific storage or particular field names. 

## Estimate: 
1 day

dharizza:
fixes #3454 
## QA Steps:
- [ ] Create a new dataset with a distribution and download URL (you can also add topics, tags and publishers), multiple nodes should be created for the different data types with the right json data.
- [ ] If you edit a dataset but you don't update the tags, topics or publisher fields, then no new nodes of those types are created
- [ ] If you edit a dataset and add an extra non-existent tag, topic or publisher, new nodes for the corresponding data type should be created.
- [ ] Do a thorough code review to confirm there is no reference to node-specific storage or particular field names in the referencer system, everything should be in the storage classes.


dafeder (approved):

dafeder:
@dharizza code looks good but after QA seeing this behavior in the content list, which I don't think is right. What happened to keywords and publishers? 
image

PR #3485, opened 2021-05-21
Admin updates

### LINKED ISSUE #3482: Reorganize menu items for DKAN admins
Current DKAN config and admin items are split into strange places. Make the following changes to the menu in /admin:

1. Create a new top level "DKAN" menu item in admin. Bonus points if we can add an icon for it!
2. Move the Datasets menu item from content into DKAN, rename to "Administer DKAN Catalog."
3. Move everything from Configuration > DKAN to DKAN.

See the links.menu.yml files in `metastore-entity` branch for some pointers, as we're already moving in this direction there.

### Extra credit 

Create + add an icon to the DKAN top level admin menu!

### LINKED ISSUE #3483: Improvements to dataset administration page
We have a split experience between the datasets and content pages that doesn't really make sense. Let's consolidate them.

1. Remove override of "content" view completely.
2. Rename title of "Datasets" view to "Administer DKAN catalog" (open to better suggestions for this)
3. Add a data type filter just like the one currently on the overridden Content view. Default to "dataset".
4. Modify node titles in the same way we're currently doing in the content screen. 
5. For now, remove Language filter.
6. Add a data type column

dafeder:
Improves menu organization and consolidates content admin into a single view, leaving "content" as is.

## QA Steps

1. Build a site with content.
2. Go to admin page, confirm that all DKAN content and config can be managed under the DKAN menu item
3. Confirm that the "Datasets" menu item provides all needed functionality to manage the catalog.


modules/metastore/metastore.routing.yml
```diff
@@ -119,7 +119,7 @@ metastore.data_json:
     _auth: ['basic_auth', 'cookie']
 
 dkan.metastore.config_properties:
-  path: '/admin/config/dkan/properties'
+  path: '/admin/dkan/properties'
```
dharizza, line 122: Shouldn't we add a title for the page?

modules/metastore/modules/metastore_admin/config/install/views.view.dkan_dataset_content.yml
```diff
@@ -750,16 +764,17 @@ display:
...
+        - 'config:field.storage.node.field_data_type'
   page_1:
     display_options:
-      path: admin/content/datasets
+      path: admin/dkan/datasets
       menu:
         type: normal
         title: Datasets
-        description: ''
+        description: 'Adminster all content in the DKAN metastore'
```
dharizza: Typo: "Administer"

modules/datastore/datastore.routing.yml
```diff
@@ -172,7 +172,7 @@ datastore.sql_endpoint.post.api:
     _access: 'TRUE'
 
 datastore.sql_endpoint.settings:
-  path: '/admin/config/dkan/sql_endpoint'
+  path: '/admin/dkan/sql_endpoint'
```
dharizza, line 175: We should add a title for this page. It is confusing to just see this: 
image

dharizza (commented):
This is looking good but I think we should add titles to the different pages. Also, when installing it works great but if I try to update from an existing site we don't get the overrides, should we add manual steps to the release notes or automate the update?

dafeder:
@dharizza believe I addressed these issues and added a couple other wording improvements (could still be better i'm sure)

dharizza (approved):
This works great. Only small issue is that when upgrading an existing site the Datasets item is not present in the DKAN tab, so we need to run this command to get it:
```
dktl drush cim --partial --source=modules/contrib/dkan/modules/metastore/modules/metastore_admin/config/install
```
I'm still going to merge and we can add this as a deployment step in the release notes.

PR #3502, opened 2021-05-29
Improve facets and drafts behavior 

fmizzell:
There were some issues with the Facets. I believe the first commit addresses that.

But that did not fix all the issues. In a system where we do not publish things by default (things are created as drafts), currently that is affecting the references while previously references were always published.

We can fix this in 2 ways:
1) Bring the previous behavior back and always publish references on creation, or
2) Make it so references are published with the parent

The fact that currently references are not published on creation is breaking facets in those system that allow datasets to be created as drafts.


susisiman:
Let's revisit and revaluate this PR

PR #3504, opened 2021-06-03
Remove idenitfier/data wrapper, take 2

### LINKED ISSUE #3762: Get rid of data/identifier wrappers in subschemas/child entities
Everything that's not a dataset is currently stored in an object that consists of 

```json
{
  "identifier": "[uuid]",
  "data": {}
}
```

where the actual metadata goes in "data", whether it's a string or an object. This was a sort of stopgap a long time ago to deal with strings and objects without identifiers in them, when really we should have a data model that does not require the identifier explicitly in the JSON.

We of course need to do this in a way that either updates all existing metadata or continues to de-reference legacy items correctly.

dafeder:
Fixes #3762 

- [ ] All tests pass
- [ ] JSON metadata stored in the database is not wrapped with with `[{"data":{},"identifier":""}]`
- [ ] API GET requests return the same JSON format as before (the data is wrapped with `[{"data":{},"identifier":""}]`)

## QA Steps

- [ ] Post a dataset with referenced entities (distributions, publisher, keyword, etc)
- [ ] Make sure that JSON metadata stored in the database is not wrapped with `[{"data":{},"identifier":""}]`
- [ ] Run GET requests to get entities of various schemas. Make sure that JSON output has not changed compared to 2.x.


dafeder:
Someday we'll get this done 😞

PR #3505, opened 2021-06-03
3474 hide record number

### LINKED ISSUE #3474: Hide record number from DKAN
## Due date:
June 2nd

## Acceptance Criteria: 
Upon completion of this work, the Agent/Broker Tracker data will hide the record number from DKAN

## Description: 
The work involved in this ticket is as follows:
A/ front end: hiding the record number in the preview (i.e., dropping column)
(not required with DKAN 3441)
B/ back end: CSV download would need to
This ticket coincides with DKAN 3441 (back end fix)

## Estimate: 
1

klimp-drupal:
fixes [GetDKAN/dkan/issues/3474]

- [ ] Test coverage exists
- [ ] Boolean `rowIds` parameter is added to the query schema
- [ ] `record_number` is hidden by default for the `query` endpoints
- [ ] setting `rowIds=true` - includes `record_number` to the output

## QA Steps

- [ ] run `/api/1/datastore/query/<distribution-id>` with and without `?rowIds=true`. Make sure the parameter affects the `record_number` field. 
- [ ] Do the same with the streaming endpoint `/api/1/datastore/query/<distribution-id>/download?format=csv`. Make sure that the visibility of the `record_number` column in the csv file depends on the `rowIds` parameter.
- [ ] Run the queries above with `?rowIds=` equals any other value but `true`. Make sure you get the validation error message.

### DIFF
--- a/modules/datastore/datastore.services.yml
+++ b/modules/datastore/datastore.services.yml
@@ -6,7 +6,6 @@ services:
       - '@dkan.datastore.service.factory.import'
       - '@queue'
       - '@dkan.common.job_store'
-      - '@request_stack'
 
   dkan.datastore.service.resource_localizer:
     class: \Drupal\datastore\Service\ResourceLocalizer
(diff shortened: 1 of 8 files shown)


dafeder (approved):
Code and tests look great. Will QA now and merge if no issues.

PR #3507, opened 2021-06-04
Updated site config to point 404 and 403 to the frontend.

brianseek:
## QA Steps

- [ ] spin up dkan demo
- [ ] visit it a non-existent  page
- [ ] verify that you are routed to the homepage of the catalog app.


janette:
@dgading is this good UX? Can we add a 404 template instead?

PR #3510, opened 2021-06-07
Dynamic API docs generator with plugins

### LINKED ISSUE #3471: Update Swagger Document API
## Due date:
June 2nd

## Description: 
The purpose of this ticket is to document the API changes, all of which we will use for the user training, as well as help desk support.

## Acceptance Criteria: 
All API changes will be well documented. The documentation will be linked within this ticket and also on Confluence. It will be peer-reviewed. And it will also be reviewed by a non-technical person like Heidi

## Estimate: 
5

dafeder:
This completely replaces our existing OpenAPI spec generation with a more dynamic and modular one. A new plugin type, DkanApiDocs, is available, and documents can add their own sections to the spec. The spec is validated against the OpenAPI 3 spec on generation. The metastore API is more fully documented, and all endpoints have working examples. 

## QA steps
- Open the API page (/api) on the QA site.
- Try all the endpoints. Test out the examples and check that responses are accurate and examples work as expected
- Note whether the endpoints are documented in a way that is clear
- Visit a dataset page, ensure that the API docs for specific dataset still work as expected.
- Check /api/1.yml, for a YAML version of the spec.
- Check /api/1?authentication=false and confirm that authenticated endpoints and related components have been removed.

## Known issues
- The dataset-specific docs really need a refactor. They use the new way to generate the spec but then use a bunch of unnecessary logic on top of it left over from the old version to filter down to the two endpoints we actually use.
- Dredd tests have been disabled. Dredd just does not support enough of OpenAPI 3 to be practical. See #3536 for a possible way forward.

## Other notes
- Introduced a new metastore storage method: getRange(). This should be useful for paginating through datasets on sites with a lot of content, where getAll() will be impractical.
- Changed `JsonResponseTrait::getResponseFromException()` to provide more information including an arbitrary "data" property that tries to extract information from the caught exception object. Currently just supports ValidationException.


modules/common/docs/openapi_spec.json
```diff
@@ -0,0 +1,68 @@
...
+                    },
+                    "data": {
+                        "type": "object",
+                        "description": "Arbitrary object storing more detailed data on the error message."
+                    }
+                }
+            }
+        }
+    }
+}
```
klimp-drupal: No new line

modules/common/src/Controller/AuthCleanupHelper.php
```diff
@@ -0,0 +1,200 @@
...
+   *
+   * @param Drupal\common\Plugin\OpenApiSpec $spec
+   *   Full spec.
+   *
+   * @return Drupal\common\Plugin\OpenApiSpec
+   *   Spec without unneeded schemas.
+   */
+  public static function cleanUpSchemas(OpenApiSpec $spec) {
+    $specArr = $spec->{"$"};
+    if (empty($specArr["components"]) || empty($specArr["components"]["parameters"])) {
```
klimp-drupal: Shouldn't it be schemas instead of parameters?

modules/common/src/Controller/OpenApiController.php
```diff
@@ -0,0 +1,121 @@
...
+   * Constructor.
+   */
+  public function __construct(
+    ModuleHandlerInterface $moduleHandler,
+    RequestStack $requestStack,
+    DkanApiDocsPluginManager $manager
+  ) {
+    $this->moduleHandler = $moduleHandler;
+    $this->requestStack = $requestStack;
+    $this->generator = new DkanApiDocsGenerator($manager);
```
klimp-drupal: I think it should work if you simply inject DkanApiDocsGenerator as a third argument

modules/common/tests/src/Controller/OpenApiControllerTest.php
```diff
@@ -0,0 +1,85 @@
...
+      ->getMock();
+
+    return (new Chain($this))
+      ->add(Container::class, 'get', $options)
+      ->add(DkanApiDocsPluginManager::class, 'getDefinitions', [$pluginDefinition])
+      ->add(DkanApiDocsPluginManager::class, 'createInstance', $pluginMock)
+      ->add(RequestStack::class, 'getCurrentRequest', $request)
+      ->getMock();
+  }
+}
```
klimp-drupal: New line

modules/datastore/src/Plugin/DkanApiDocs/DatastoreApiDocs.php
```diff
@@ -0,0 +1,376 @@
...
+      = $this->queryExample();
+
+    $spec['components']['parameters']['datastoreUuid']["example"] = $exampleIds['resource'];
+
+    $spec["paths"]["/api/1/datastore/sql"]["get"]["parameters"][0]["example"] = $this->sqlQueryExample($exampleIds['distribution']);
+    return $spec;
+  }
+
+  /**
+   * Set up the GET parameters for datastore queries. WIP.
```
klimp-drupal, line 147: Still WIP?

modules/common/src/Controller/OpenApiController.php
dafeder: I did not actually implement DkanApiDocsGenerator as a serivice. Maybe I should do so?

modules/common/src/Controller/AuthCleanupHelper.php
dafeder: Yes, oops.

modules/datastore/src/Plugin/DkanApiDocs/DatastoreApiDocs.php
dafeder: Yes, these get parameters are very hard to document because we are really just passing a converted JSON object as a query string, which doesn't really map to any of the OpenAPI options for documenting query parameters. I want to take a second pass at documenting these well after this gets merged.

modules/common/src/Controller/OpenApiController.php
dafeder: Nevermind, yes I did! Refactoring.

dafeder:
Addressed all code review issues

klimp-drupal:
**Datastore** entrypoints QA findings:

- imports - `POST` , `DELETE`. You put a note that the id format is different from what is returned from GET but what is the format exactly? It'd be useful to have it documented.
- `GET /api/1/datastore/query`, `/api/1/datastore/query/download` - parameters didn't work for me. I tried to add "resources" - it complained that the "resources" should be an array, string given. Either a fix or more instructions needed
- `GET /api/1/datastore/query/{identifier}/download` - streaming is not available for JSON. No way to specify other format

klimp-drupal:
**Harvest** entrypoints QA findings:

- `GET ​/api​/1​/harvest​/runs​/{run_id}`, `/api/1/harvest/runs` - missing 'plan' parameter
- `POST /api/1/harvest/runs` - harvest must be a php object. Only JSON available for input

klimp-drupal:
**Metastore** entrypoints QA findings:

- `POST /api/1/metastore/schemas/dataset/items`, `/api/1/metastore/schemas/distribution/items` - the given example fails schema validation
- `GET /api/1/metastore/schemas/dataset/items/{identifier}`, `/api/1/metastore/schemas/distribution/items/{identifier}` - `show-reference-ids` doesn't seem to affect anything
- All Metastore `PUT` and `PATCH` - identifier in the request body should match the identifier in the "identifier" field.
- `PUT /api/1/metastore/schemas/dataset/items/{identifier}`, `/api/1/metastore/schemas/distribution/items/{identifier}` - an error: `"downloadURL": "string" - "The attribute should match 'uri' format."` Other parameters should have a correct format, not "string"
- `PATCH /api/1/metastore/schemas/dataset/items/{identifier}`, `/api/1/metastore/schemas/distribution/items/{identifier}` -  `"describedBy": "string"`,  `"license": "string"`, and other parameters should have correct formats

dafeder:
> **Harvest** entrypoints QA findings:
> 
>     * `GET ​/api​/1​/harvest​/runs​/{run_id}`, `/api/1/harvest/runs` - missing 'plan' parameter

Fixed.

> 
>     * `POST /api/1/harvest/runs` - harvest must be a php object. Only JSON available for input

OK, this is from the old docs and wasn't touched, will make a ticket to fix later. Not a regression. I think a better example just needs to be added.

dafeder:
> **Datastore** entrypoints QA findings:
> 
>     * imports - `POST` , `DELETE`. You put a note that the id format is different from what is returned from GET but what is the format exactly? It'd be useful to have it documented.

Agreed. I think this needs to be a new ticket though.

>     * `GET /api/1/datastore/query`, `/api/1/datastore/query/download` - parameters didn't work for me. I tried to add "resources" - it complained that the "resources" should be an array, string given. Either a fix or more instructions needed

I really struggled to document these because swagger really doesn't support how we do these parameters. Am hoping that at least the defaults work?
>     * `GET /api/1/datastore/query/{identifier}/download` - streaming is not available for JSON. No way to specify other format

Added a format parameter just for this one, see if it makes sense.

dafeder:
> **Metastore** entrypoints QA findings:
> 
>     * `POST /api/1/metastore/schemas/dataset/items`, `/api/1/metastore/schemas/distribution/items` - the given example fails schema validation

See comments for PUT/PATCH.

 
>     * `GET /api/1/metastore/schemas/dataset/items/{identifier}`, `/api/1/metastore/schemas/distribution/items/{identifier}` - `show-reference-ids` doesn't seem to affect anything

Known issue, see #3531 

>     * All Metastore `PUT` and `PATCH` - identifier in the request body should match the identifier in the "identifier" field.

Still no real mechanism to generate good test values for these metastore write endpoints. I added a note on each one to describe how to test using an existing record as a base.
 
>     * `PUT /api/1/metastore/schemas/dataset/items/{identifier}`, `/api/1/metastore/schemas/distribution/items/{identifier}` - an error: `"downloadURL": "string" - "The attribute should match 'uri' format."` Other parameters should have a correct format, not "string"

Yeah, see above. Limitation of swaggerUi's attempt to generate an example value when you haven't provided one.

>     * `PATCH /api/1/metastore/schemas/dataset/items/{identifier}`, `/api/1/metastore/schemas/distribution/items/{identifier}` -  `"describedBy": "string"`,  `"license": "string"`, and other parameters should have correct formats

See above.

klimp-drupal:
`GET ​/api​/1​/harvest​/runs​/{run_id}` - still getting 
```
{
  "message": "Missing 'plan' query parameter value"
}
```

klimp-drupal:
`GET /api/1/datastore/query/download` - doesn't work with default parameters. `format` is JSON by default resulting in `"message": "Streaming not currently available for JSON responses",`. Should be `csv`

klimp-drupal:
`GET /api/1/datastore/query/download` - there is no link to the file. (probably because it is empty?). Not sure if it is a bug but `GET /api/1/datastore/query/{identifier}/download` has the file link.

dafeder:
> `GET ​/api​/1​/harvest​/runs​/{run_id}` - still getting
> 
> ```
> {
>   "message": "Missing 'plan' query parameter value"
> }
> ```

OK after testing this some more I've determined that this method is actually broken in the controller. I will make a new ticket for fixing both the endpoint and the doc

dafeder:
> `GET /api/1/datastore/query/download` - doesn't work with default parameters. `format` is JSON by default resulting in `"message": "Streaming not currently available for JSON responses",`. Should be `csv`

Will make a separate ticket to address.

dafeder:
Got the OK to merge this from @klimp-drupal, will open issues for outstanding points above

PR #3511, opened 2021-06-07
3506 api dataset creation no identifier

### LINKED ISSUE #3506: API dataset creation w/o identifier results in schema failure
Because we have moved validation closer to the controller, we new datasets added to POST are being rejected if they do not have identifiers, as identifier is a required field. 

There are a few ways this could be addressed, but my suggestion is:

1. Find a way to tell `ValidMetadataFactory` that we are doing a post and/or creating something new when calling it, either by passing a new parameter to `ValidMetadataFactory::get()` or creating a new public method specifically for this situation.
2. If `ValidMetadataFactory` gets the signal that we're posting, it checks the schema for the existence of a required "identifier" field in the schema. If it exists, and the data does not include an identifier, a new one should be generated at that point and injected into the dataset before validation.

At some point, we probably want to abstract this more and have it live with the SchemaRetriever or metadata storage mechanism, so that different schemas can define their own identifier fields, and maybe even additional logic about how and when to do this generation. For now though, this will restore the basic dataset creation functionality.

klimp-drupal:
fixes [GetDKAN/dkan/issues/3506]

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

- [ ] Create a dataset with the Metastore API `POST` endpoint - `/api/1/metastore/schemas/dataset/items`
- [ ] Do not include `identifier` in your JSON payload, e.g.
```
{
  "title": "Dummy title",
  "description": "Dummy description",
  "accessLevel": "public",
  "modified": "2020-02-02",
  "keyword": [
    "test"
  ],
  "publisher": {
    "@type": "org:Organization",
  },
  "distribution": [
    {
      "downloadURL": "file:///var/www/docroot/modules/contrib/dkan/modules/sample_content/files/Bike_Lane.csv",
      "mediaType": "text/csv",
      "format": "csv",
      "title": "Florida Bike Lanes"
    }
  ]
}
```
- [ ] Make sure that a new dataset has been created

### DIFF
--- a/modules/metastore/src/ValidMetadataFactory.php
+++ b/modules/metastore/src/ValidMetadataFactory.php
@@ -54,10 +54,10 @@ class ValidMetadataFactory implements ContainerInjectionInterface {
   /**
    * Converts Json string into RootedJsonData object.
    *
-   * @param string|null $schema_id
-   *   The {schema_id} slug from the HTTP request.
    * @param string $json_string
    *   Json string.
+   * @param string|null $schema_id
+   *   The {schema_id} slug from the HTTP request.
    * @param array $options
    *   Options array.
    *
@@ -66,7 +66,7 @@ class ValidMetadataFactory implements ContainerInjectionInterface {
    *
    * @throws \JsonPath\InvalidJsonException
    */
-  public function get($schema_id = NULL, string $json_string, array $options = []): RootedJsonData {
+  public function get(string $json_string, $schema_id = NULL, array $options = []): RootedJsonData {
 
     // Add identifier for new objects if necessary.
     if (isset($options['method']) && $options['method'] == 'POST') {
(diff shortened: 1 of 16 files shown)


modules/metastore/src/ValidMetadataFactory.php
```diff
@@ -3,13 +3,15 @@
...
 use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
+use Drupal\metastore\Reference\HelperTrait;
 use RootedData\RootedJsonData;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
  * Service.
  */
 class ValidMetadataFactory implements ContainerInjectionInterface {
+  use HelperTrait;
```
dafeder: @klimp-drupal I don't think it makes sense to pull in a trait from the Reference namespace, which is unrelated to ValidMetadataFactory. Also all we're getting from that trait is a reference to the Uuid class, which we could invoke directly. So just use Service\Uuid and then call something like `(new Uuid)->generate($schema_id, $json_string);`

dafeder (changes requested):

modules/metastore/src/ValidMetadataFactory.php
klimp-drupal: Fixed

dafeder (approved):

PR #3516, opened 2021-06-10
Dereferencer. Caught MissingObjectException

### LINKED ISSUE #3513: Avoid throwing exception when a reference has been deleted
Currently a number of pages and endpoints error out if any dataset contains a reference to a deleted node such as a distribution. This can possibly be accomplished by making the dereferencer simply return NULL or FALSE when the storage cannot find a metadata item for the reference ID. 

Possibly we should log a warning or in some other way communicate that a reference has disapeared.

## To reproduce

- Create a dataset that includes a distribution. 
- Go do /admin/dkan/datasets and delete the distribution
- The datasets page, as well as /api/1/metastore/schemas/datasets/items should cause a 500 error.

## Recommended approach

1. Throw a more specific exception from MetastoreStorageInterface when retrieve returns no results (probably Drupal\metastore\Exception\MissingObjectException).
2. Catch that specific exception in Dereferencer::dereferenceSingle() and log to watchdog, returning NULL. This is actually the current intended behavior as you will see in the dereferencer code, but the exception isn't caught.

Except the basics of this to be pretty easy to implement, but we want to be extra sure we aren't breaking something else. So double-check test coverage and make sure we aren't missing anything.

### LINKED ISSUE #3517: Deleting distribution nodes should trigger some db clean up
Deleting distributions will leave a mess if not processed correctly. The orphan_reference_processor is only checking for dataset deletion.

## Steps to reproduce
1. Create a dataset with a csv distribution
2. run `drush queue:run datastore_import`
3. Inspect the database, you will see:
    - a datastore table
    - 3 entries in the _dkan_metastore_resource_mapper_ table
    - 1 entry in the _jobstore_dkan_datastore_importer_ table
    - 1 entry in the _jobstore_filefetcher_filefetcher_ table
4. Scenario 1:
    - Delete the distribution using the UI (no error thanks to #3516)
    - Delete the dataset using the UI and run `drush queue:run orphan_reference_processor`
    - it says it "Processed 4 items"
    - return to the database, you will see that the datastore is still there, all entries in the resource_mapper and jobstore tables remain
5. Scenario 2:
    - Delete the distribution
    - Use the API to add an updated csv
    - Run `drush queue:run datastore_import`
    - The old datastore, with the previous data, will be used in its place

klimp-drupal:
fixes [GetDKAN/dkan/issues/3513]

- [x] Test coverage exists
- [x] Pages listing datasets as well as related API edpoints return 200 without any errors in case if one of the dataset's references is deleted

## QA Steps

- [x]  Create a dataset that includes a distribution.
- [ ]  Go do `/admin/dkan/datasets` and delete the distribution
- [ ]  The datasets page, as well as `/api/1/metastore/schemas/dataset/items` should return a dataset with an empty distribution and not cause a 500 error.
- [ ] Make sure that a new item in the `orphan_resource_remover` queue is created
- [ ] Run the queue - `drush queue:run orphan_resource_remover`
- [ ] Make sure that the resources associated with the deleted distribution are deleted as well (the `datastore_*` table and the records in the `dkan_metastore_resource_mapper` table)


janette:
@dafeder @klimp-drupal  I think we need to implement [public static function ConfigEntityBase::preDelete](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Config%21Entity%21ConfigEntityBase.php/function/ConfigEntityBase%3A%3ApreDelete/8.2.x) to remove all the same things the EVENT_ORPHANING_DISTRIBUTION sets in motion. See #3517

And as you expected, we will need to add checks for null:
```
TypeError: Argument 1 passed to Drupal\metastore\ResourceMapper::remove() must be an instance of Drupal\common\Resource, null given, called in /var/www/docroot/modules/contrib/dkan/modules/metastore/src/EventSubscriber/MetastoreSubscriber.php on line 82 in Drupal\metastore\ResourceMapper->remove() (line 133 of modules/contrib/dkan/modules/metastore/src/ResourceMapper.php). 
```

dafeder (changes requested):
We need to remove empty distribution arrays if there are no actual distributions there. See https://github.com/GetDKAN/dkan/compare/temp-docs-3513-3520#diff-87830ad7dfa5b1aa79a08445294a4724b4bde19e93044403a319abcb13a1bf55R58.

Once this is in place, we may be able to go back to throwing an invalid JSON excpetion without issue, since this was the main thing breaking that validation.

janette:
Don't forget the JobStore tables 
```
jobstore_filefetcher_filefetcher
jobstore_dkan_datastore_import
 ```

dafeder (approved):

janette:
@klimp-drupal It works up to step 3, but after that I get `Queue not found: orphan_resource_remover`

klimp-drupal:
@janette sounds like cache. I should've mentioned that in the test steps

PR #3520, opened 2021-06-12
Remove preview button

### LINKED ISSUE #3519: Hide/Remove preview button when creating new dataset
## User Story

As an authenticated user, I should do not need to see a dataset preview button when creating a new dataset, because it will only show me unformatted JSON.

## Acceptance Criteria
- log in as an authenticated user
- create a new dataset 
- confirm there is no preview button

janette:
fixes #3519

json blobs tend to alarm people

## QA Steps

- [ ] Visit `/node/add/data`, confirm there is no 'Preview' button
- [ ] Edit and existing dataset, confirm there is no 'Preview' button


dafeder (changes requested):
@janette I think the correct way to do this would be to change the content type config, not add a form_alter:

image

We can do this either in DKAN default config, or on a per-site basis. Thoughts?

PR #3524, opened 2021-06-14
Update the dkan admin view to filter by dataset by default

janette:
The title of the dkan content view, is 'Datasets' so let's default to show just the datasets by default

## QA Steps

- [ ] Navigate to `/admin/dkan/datasets` and confirm you only see datasets by default and can use the UI to select other options


susisiman:
@dafeder is working on this ticket. Not a high priority but if there is time he will work on it.  @janette did we test this on an upgrade?

dafeder:
Assigning to @janette to confirm upgrade behavior.

dafeder:
Blocked for now by #3556

dafeder:
Glad we finally got those upgrade tests working!

PR #3528, opened 2021-06-17
Allow query by dataset/distribution-index in addition to distribution for datastore

### LINKED ISSUE #3514: Allow query by dataset/distribution-index in addition to distribution for datastore
In order to perform a datastore query, the user must submit a distribution UUID. This can be hard to find in the UI and even in the API if you don't know what you're looking for. We'd like to add an alternative endpoint that takes just a dataset UUID and an index.

So, if we have a dataset with identifier "abc" which contains a single distribution with identifier "xyz", we would want the endpoint:

/api/1/datastore/query/xyz

...to have an equivilant endpoint that looks like:

/api/1/datastore/query/abc/0

with "0" being their position in the datasets' distribution index.

dharizza:
fixes #3514 

## QA Steps

- If you have a dataset with identifier "abc" which contains a single distribution with identifier "xyz" and you query it using the endpoint `/api/1/datastore/query/xyz`, you should also be able to use the endpoint: `/api/1/datastore/query/abc/0` and get the same results. You can try the following URLs:
  - [ ] https://dkan-qa-3528.ci.civicactions.net/api/1/datastore/query/e1f2ebcd-ee23-454f-87b5-df0306658418/0 should return the same set of results than https://dkan-qa-3528.ci.civicactions.net/api/1/datastore/query/5dfd9d60-4f20-5b1d-bcc5-a385a95f7f30
  - [ ] https://dkan-qa-3528.ci.civicactions.net/api/1/datastore/query/e1f2ebcd-ee23-454f-87b5-df0306658418/1 should return the same set of results than https://dkan-qa-3528.ci.civicactions.net/api/1/datastore/query/ca419d8e-8a29-55b0-a323-ad917741ff6a
- If you query the endpoint `/api/1/datastore/query/abc/0` but there isn't a dataset with the identifier abc, then you should get an error. You can try:
  - [ ] https://dkan-qa-3528.ci.civicactions.net/api/1/datastore/query/e1f2ebcd-ee23-454f-87b5-df0306658417/0 should return "No dataset found with the identifier e1f2ebcd-ee23-454f-87b5-df0306658417".
- If you query the endpoint `/api/1/datastore/query/abc/1` but there isn't a distribution in the index 1 for the dataset with the identifier abc, then you should get an error. You can try:
  - [ ] https://dkan-qa-3528.ci.civicactions.net/api/1/datastore/query/e1f2ebcd-ee23-454f-87b5-df0306658418/2 should return "No resource found at index 2".

### DIFF
--- a/modules/datastore/src/Controller/QueryController.php
+++ b/modules/datastore/src/Controller/QueryController.php
@@ -276,13 +276,10 @@ class QueryController implements ContainerInjectionInterface {
     if (!isset($metadata['latest_revision'])) {
       return $this->getResponse((object) ['message' => "No dataset found with the identifier $dataset"], 400);
     }
-    if (!isset($metadata['latest_revision']['distributions'][$index])) {
+    if (!isset($metadata['latest_revision']['distributions'][$index]['distribution_uuid'])) {
       return $this->getResponse((object) ['message' => "No resource found at index $index"], 400);
     }
     $identifier = $metadata['latest_revision']['distributions'][$index]['distribution_uuid'];
-    if (!$identifier) {
-      return $this->getResponse((object) ['message' => "Distribution not found."], 400);
-    }
     return $this->queryResource($identifier, $stream);
   }
 
--- a/modules/datastore/src/Controller/QueryController.php
+++ b/modules/datastore/src/Controller/QueryController.php
@@ -2,6 +2,7 @@
 
 namespace Drupal\datastore\Controller;
 
+use Drupal\common\DatasetInfo;
 use Drupal\datastore\Service\DatastoreQuery;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpFoundation\RequestStack;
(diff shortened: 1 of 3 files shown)


modules/datastore/src/Controller/QueryController.php
```diff
@@ -49,7 +58,8 @@ public function __construct(Service $datastoreService, RequestStack $requestStac
   public static function create(ContainerInterface $container) {
     $datastoreService = $container->get('dkan.datastore.service');
     $requestStack = $container->get('request_stack');
-    return new QueryController($datastoreService, $requestStack);
+    $datasetInfo = $container->get('dkan.common.dataset_info');
```
dafeder, line 61: I think this is not the ideal service to use here, but I know it's the easiest one. Just noting that we want to consolidate DatasetInfo with the metastore classes.

dafeder (changes requested):
Looks great @dharizza but can you add more routes? Every query route that currently has a /query/{identifier} path should have an equiviland /query/{dataset}/{index} route.

PR #3548, opened 2021-07-06
ValidMetadata argument order in metadata.module

dafeder:
metadata.module still using old argument order. Also reveals gap in test coverage to be fixed at another time.


klimp-drupal (approved):

dafeder:
Duplicate of #3518

PR #3553, opened 2021-07-09
Organizing the metastore's unit tests.

fmizzell:
We have a test suite that runs only the unit tests. That suite assumes that that unit tests are in a directory called Unit which is also a standard practice instituted by Drupal. This PR moves the metastore unit tests to that place. I also deleted a couple of unnecessary tests.


dafeder:
Thanks @fmizzell! Been meaning to do some organization along these lines. Re-running the CI to see if we can get it green.

PR #3554, opened 2021-07-10
By default trigger datastore update when dataset modified date changes

### LINKED ISSUE #3552: By default trigger datastore update when dataset modified date changes
## User Story: 

As a data publisher, I want to be able to replace a file with a file of the same name and trigger a datastore update so that the datastore always reflects the current file. 

As a data publisher, I want to be able to tell DKAN about my remote file update and trigger a datastore update so that the datastore always reflects the current remote file. 

## Acceptance Criteria: 
- The modified date change triggers a datastore update of a dataset

dharizza:
fixes #3552 

## QA Steps

- [ ] Go to admin/dkan/datastore and set the field you want as the datastore triggering property.
- [ ] Create a new dataset and confirm it is imported to the datastore.
- [ ] Edit your dataset and update the value for the field you chose in step 1 (by default we use the Last Update - modified date), save the dataset.
- [ ] Confirm the datastore is updated.


dafeder:
Need test coverage for `retrieveSchemaProperties()`

modules/datastore/src/EventSubscriber/DatastoreSubscriber.php
```diff
@@ -147,4 +148,26 @@ public function drop(Event $event) {
...
+   * React to a preReference to check if datastore update should be triggered.
+   *
+   * @param \Drupal\common\Events\Event $event
+   *   The event object containing the resource uuid.
+   */
+  public function onPreReference(Event $event) {
+    $data = $event->getData();
+    $original = $data->getOriginal();
+    $field = \Drupal::service('config.factory')->getEditable('datastore.settings')->get('triggering_property');
+    $field = $field ? $field : 'modified';
```
dafeder: Let's avoid hard-coding "modified" here. I know the issue suggested otherwise but realizing this should just default to nothing.

modules/datastore/src/Form/DatastoreSettingsForm.php
```diff
@@ -0,0 +1,86 @@
...
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    $form['fieldset'] = [
+      '#type' => 'fieldset',
+      '#title' => $this->t('Triggering property'),
+      '#description' => $this->t('Property to trigger update of the datastore.'),
+    ];
+    $form['fieldset']['triggering_property'] = [
+      '#type' => 'select',
```
dafeder: This should have been specified in the ticket, but after QAing this I think it's clear that this should be checkboxes with the ability to have multiple or no fields chosen.

dafeder (changes requested):
@dharizza after QAing realized we should have thought more about the form specifics -- let's make that small change before shipping this.

janette:
notes:
- Add option for none (default)
- Add test for none option
- Do not hard code property names

clayliddell:
The 1 issue that CodeClimate is complaining about is not possible to resolve without triggering another CodeClimate error... Not sure how to proceed.

dafeder:
@clayliddell we have the ability to approve a PR through code climate despite having a warning -- we do this often for constructors because the dependency injection technique often leads to this issue.

On QA I'm seeing strange behavior when testing having no property checked. I create a dataset and then update the modified date. Despite not having any property checked, I am seeing a new distribution created and a new datastore_import job in the queue. When I run the queue I get to a strange state where the dataset no longer appears in the catalog (it's gone from api/1/metastore/schemas/dataset/items). Inspecting the JSON it looks like in the new distribution that was created, the downloadUrl was simply left blank. 

I think it is working as expected with "modified" checked.

dafeder (approved):

dafeder:
Jenkins failure is a CI server problem, build is fine.

dafeder:
I am seeing a resource_purger job created after patching a dataset with nothing checked, which probably shouldn't happen. But, I do not see any ill effects from this.

PR #3559, opened 2021-07-14
Added CachableResponseTrait

### LINKED ISSUE #3558: DKAN delivers appropriate HTTP cache headers
### User Story
DKAN is [returning the page directly via a Symfony response](https://github.com/GetDKAN/dkan/blob/2.x/modules/frontend/src/Controller/Page.php) rather than a page array (which is the part that contains cache tags/contexts and makes it cachable by Drupal core) or other higher level API that would enable Drupal to set appropriate HTTP headers.

This is occurring on frontend pages as well as /api responses, and perhaps other endpoints also - we should validate that all DKAN pages can be cachable.

### Notes

DKAN will need to switch the implementation to use either:

1. [Drupal's CacheableResponse](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Cache%21CacheableResponse.php/class/CacheableResponse/8.2.x) instead of a vanilla Symfony response - this is the ideal method (especially for the API routes) as it would allow longer cache expiration times since cache tags can be used to enable efficient invalidation (including of the Akamai cache, which supports cache tags) when datasets are updated.

2. [Implement it's own cache headers using](https://symfony.com/doc/current/components/http_foundation.html#managing-the-http-cache|https://symfony.com/doc/current/components/http_foundation.html#managing-the-http-cache) - this is simpler to implement, but doesn't offer easy options for cache invalidation, so would likely necessitate shorter cache expiration times for resources that change frequently.

I did a quick test by adding this to the DKAN frontend [Controller/Page.php](https://github.com/GetDKAN/dkan/blob/2.x/modules/frontend/src/Controller/Page.php) linked above and confirmed that this does add cache headers: 

```
 $response = Response::create($pageContent);
 $response->setCache([ 
 'public' => true, 
 'private' => false, 
 'max_age' => 600, 
 'last_modified' => new \DateTime(), 
 ]);
  $response->setVary('Cookie');
 return $response;
```

This isn't an ideal fix on it's own however, it doesn't include an etag or useful last_modified date - nor deal with cache invalidation using the Acquia/Akamai modules (hence the short max_age) - it also doesn't fix other routes like /api that have the same issue.

We should include Drupal config to enable/disable caches (disabled by default in DKAN core) and set expiry

### Story Points 
1

### Due 
07/14/21

klimp-drupal:
fixes [GetDKAN/dkan/issues/3558]

## Notes
- no test coverage
- cache headers are added in the trait which is quick but not the best approach

## QA Steps
Navigate to the following pages and make sure caching headers are there:
- [x] any HTML page
- [x] any API JSON entrypoint
- [x] any API CSV entrypoint
- [x] any API YAML entrypoint

### DIFF
--- a/modules/common/src/CacheableResponseTrait.php
+++ b/modules/common/src/CacheableResponseTrait.php
@@ -14,7 +14,7 @@ trait CacheableResponseTrait {
    *
    * @var int
    */
-  protected $maxAge;
+  protected $cacheMaxAge;
 
   /**
    * Adds cache headers to the response.
@@ -30,19 +30,33 @@ trait CacheableResponseTrait {
    * @throws \Exception
    */
   private function addCacheHeaders(Response $response) : Response {
-    if (!isset($this->maxAge)) {
-      $this->maxAge = \Drupal::config('system.performance')->get('cache.page.max_age');
-    }
-    if ($this->maxAge !== 0) {
+    $this->setCacheMaxAge();
+
+    if ($this->cacheMaxAge !== 0) {
       $response->setCache([
         'public' => TRUE,
         'private' => FALSE,
-        'max_age' => $this->maxAge,
+        'max_age' => $this->cacheMaxAge,
         'last_modified' => new \DateTime(),
       ]);
       $response->setVary('Cookie');
     }
     return $response;
+  }
+
+  /**
+   * Sets cache max age.
+   */
+  private function setCacheMaxAge() {
+    if (!isset($this->cacheMaxAge)) {
+      // A hack to bypass the controllers' tests.
+      if (\Drupal::hasService('config.factory')) {
+        $this->cacheMaxAge = \Drupal::config('system.performance')->get('cache.page.max_age');
+      }
+      else {
+        $this->cacheMaxAge = 0;
+      }
+    }
   }
 
 }
--- a/modules/common/src/CacheableResponseTrait.php
+++ b/modules/common/src/CacheableResponseTrait.php
@@ -33,13 +33,15 @@ trait CacheableResponseTrait {
     if (!isset($this->maxAge)) {
       $this->maxAge = \Drupal::config('system.performance')->get('cache.page.max_age');
     }
-    $response->setCache([
-      'public' => TRUE,
-      'private' => FALSE,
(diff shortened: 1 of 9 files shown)


grugnog (approved):

modules/common/src/JsonResponseTrait.php
```diff
@@ -13,12 +13,14 @@
  * Json Response Trait.
  */
 trait JsonResponseTrait {
+  use CacheableResponseTrait;
```
dafeder, line 16: I would say we should not use the `CacheableResponseTrait` within the `JsonResponseTrait`. Traits within traits tend to lead to a bad developer experience as it can be really hard to find the actual method you're seeing in the implementing class, and in this case it looks like we're using _both_ the cacheable and the JSON traits in the same controller, so it's redundant anyway. I'd recommend just removing that use statement from `JsonResponseTrait` and make sure `ChaceableResponseTrait` is included everywhere it's needed specifically.

modules/common/src/CacheableResponseTrait.php
```diff
@@ -0,0 +1,62 @@
...
+   * @return \Symfony\Component\HttpFoundation\Response
+   *   Symfony response.
+   *
+   * @throws \Exception
+   */
+  private function addCacheHeaders(Response $response) : Response {
+    $this->setCacheMaxAge();
+
+    if ($this->cacheMaxAge !== 0) {
+      $response->setCache([
```
dafeder, line 36: this section needs test coverage

dafeder (changes requested):
@janette @klimp-drupal this looks great... I am a bit hesitant about the traits-within-traits pattern but i think it's justified enough to merge. But I do think the test coverage should be addressed if possible.

modules/common/src/JsonResponseTrait.php
dafeder: Actually ignore this, I was misreading the code flow.

modules/common/src/CacheableResponseTrait.php
klimp-drupal: Done. Created tests to check the caching headers for JSON, YAML, CSV and HTML responses

PR #3563, opened 2021-07-16
3365 local file trasher option

### LINKED ISSUE #3365: Local file trasher option
## User Story: 
Data files live on a storage system, and data managers don't need them both in DKAN and in that system, so once the data is in the datastore, DKAN no longer needs to store the file as well. Not storing the file reduces space issues and strain on the system. 

When an import has been completed, delete the file from the local resources folder, and remove the localized "perspective" of the resource for the database.

This should be an optional behavior that is off by default, and can be enabled through an admin form.

## Acceptance Criteria:
If the behavior is enabled,
- If I create a dataset with a remote file there is no local copy of the file after the datastore is complete. The download link continues to point to the remote file.
  - In above scenario, the filefetcher job record is also deleted
  - local_file and local_url records are also deleted from the resource mapper table
- If I create a dataset with an uploaded file there is no duplicate file of the dataset after a datastore is complete

If the behavior is not enabled,
- Files remain in the file system as they do now. 
- Download links point to the localized file URL.

## Estimate: 
5 days

klimp-drupal:
fixes [/GetDKAN/dkan/issues/3365]

- [ ] Test coverage exists

## QA Steps

- [ ] Navigate to `/admin/dkan/resources` and check the "Delete local resource" checkbox. Save the form

### Local file
- [ ] 1. Create a dataset with the local resource via UI
- [ ] 2. Run datastore import `drush queue:run datastore_import`
- [ ] 3. Build frontend and find your dataset, for example `/dataset/b4bad0bb-0bfe-41b7-b882-dd3b07086dcb`
- [ ] 4. Make sure that the file gets downloaded from `/sites/default/files/uploaded_resources`
- [ ] 5. Run `dktl drush dkan:dataset-info b4bad0bb-0bfe-41b7-b882-dd3b07086dcb` to find out the resource id and version (in this case it is `cb1f2ad61c477cf14ec401aeff28b379` and `1626463096`)
- [ ] 6. Make sure that the resource folder under `/sites/default/files/resources` doesn't exist : `ll ./docroot/sites/default/files/resources/cb1f2ad61c477cf14ec401aeff28b379_1626463096`
- [ ] 7. Make sure there is no database record in the `jobstore_filefetcher_filefetcher` table : `select * from jobstore_filefetcher_filefetcher where ref_uuid = "cb1f2ad61c477cf14ec401aeff28b379_1626463096";`
- [ ] 8. Make sure there are no records in the `dkan_metastore_resource_mapper` table for the `local_file` or `local_url` prespectives: `select * from dkan_metastore_resource_mapper where identifier = "cb1f2ad61c477cf14ec401aeff28b379" and perspective in ("local_file", "local_url");`

### Remote file
- [ ] 1. Create a dataset with the remote resource
- [ ] 2. Repeat the steps 2-8 from the "Local file" except on step 4 the file should be downloaded from the remote source


klimp-drupal:
local file. The resource folder is empty but doesn't get deleted `docroot/sites/default/files/resources/f7728135a2b783904a13b1d2dd0f1d4d_1626467168/` . For the remote file it gets deleted

UPD: cannot reproduce today. Probably it was a file permission issue on my localhost

klimp-drupal:
The filefetcher record doesn't get deleted for both remote and local

modules/datastore/src/Plugin/QueueWorker/Import.php
```diff
@@ -71,6 +71,13 @@ public function processItem($data) {
       foreach ($results as $result) {
         $queued = $this->processResult($result, $data, $queued);
       }
+
+      // Delete local resource file.
+      if ($this->container->get('config.factory')->get('datastore.settings')->get('delete_local_resource')) {
```
clayliddell: We should be using dependency injection here, instead of fetching the config factory at the last minute from the container. That would allow us to more easily overwrite the "delete_local_resource" setting in tests.

clayliddell (changes requested):

clayliddell (approved):

clayliddell:
Currently, the changes in this PR take care of imported resource files left in the `/resources` directory, but not abandoned resource files left in the `/uploaded_resources` directory (i.e. in the case of a distribution being deleted, or of a distribution file being replaced).

### Update

An issue has been created for this at #3656.

PR #3565, opened 2021-07-18
Initial proof of concept.

### LINKED ISSUE #3564: Apply table schemas on datastore import
We want to support, through documentation and probably some helper classes added to the codebase, the application of basic column-level metadata (data type, description) at the moment of an import. For now this will need to be implemented via custom event listener on a per-site basis, but there should be an established pattern to create the listener and provide Drupal with table schema instructions to pass to the DB server based on [Frictionless Table Schema](https://specs.frictionlessdata.io//table-schema/) field definitions. A helper class could be the best way to pass a single field definition, such as...

```
{}
```

... and get back the correct instructions for the Drupal schema API to apply to the table.

Examples coming...

dharizza:
fixes #3564 

## QA Steps

- [ ] Enable the module
- [ ] Create a dataset with a resource, the resource should contain a CSV file. For testing you could use the data specified in the README of the schema_updater module.
- [ ] Run `dktl drush queue:run datastore_import`

## Pendant
- [ ] Fix table not created.
- [ ] Choose naming.
- [ ] Define if this will be in core.
- [ ] Create method to get the frictionless schema.


dafeder:
Closing in favor of #3668

PR #3576, opened 2021-07-22
Report QA site build status to GitHub

dafeder:
Report a second status to github from Jenkins with a direct link to the QA site. Also removes the "Check QA site" stage which seems to be unnecessary. 

Shoutout to @woodt for the new function added to the Jenkinsfile.

## QA Steps

Click "details" link in the new continuous-integration/jenkins/qa-site status in this PR and confirm it links directly to a site.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,3 +1,5 @@
+import groovy.json.JsonOutput
+
 pipeline {
     agent any
     environment {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -67,8 +67,6 @@ pipeline {
                         export DKTL_DIRECTORY="$WORKSPACE/dkan-tools"
                         echo $DKTL_DIRECTORY
                         dktl init --dkan-local
-                        dktl demo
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -72,19 +74,6 @@ pipeline {
                 }
             }
         }
-        stage('Check QA Site') {
-            when { changeRequest(); }
-            steps {
-                script {
```
dafeder, line 78: Do we actually need this?

grugnog: I think we can drop the  QA_SITE_* stuff that doesn't do anything, but I would lean towards leaving the curl command in (even if we just moved it to the last line of the prior stage), since it's a simple enough check and will alert us to issues earlier.

grugnog (commented):
Everything else looks good and I can see the status

Jenkinsfile
dafeder: Going to restore simplified version

grugnog (approved):

PR #3577, opened 2021-07-22
Support both versions of Symfony events

### LINKED ISSUE #3381: Calling EventDispatcherInterface::dispatch() with a string event name has been deprecated
https://www.drupal.org/node/3154407

Have plan to check drupal version when dispatching an event

dafeder:
See https://www.drupal.org/node/3154407

No code coverage on trait because impossible to test both cases.

## QA Steps

This will need some local CLI-based QA. Build two different DKAN instances, one with 9.0 and one with 9.1. For each, ensure that you get no errors when you:

- [x] Run the full functional test suite
- [x] Run a test harvest, revert it, and run the queues

### DIFF
--- a/modules/common/src/EventDispatcherTrait.php
+++ b/modules/common/src/EventDispatcherTrait.php
@@ -5,9 +5,11 @@ namespace Drupal\common;
 use Drupal\common\Events\Event;
 
 /**
- * Trait EventDispatcherTrait.
+ * Event dispatcher trait.
  *
- * @package Drupal\common
+ * @todo Refactor as service.
+ *
+ * @codeCoverageIgnore
  */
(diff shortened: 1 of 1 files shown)


dafeder:
Testing this with Drush is actually not creating any obvious error output, because Drush suppresses deprecation errors no matter what your PHP settings say to do 😖 . This does not appear to cause any output from PHPUnit tests, at least on my environment. So the best way to check is to debug a harvest run and break on `Drupal\Component\EventDispatcher\ContainerAwareEventDispatcher` line 116, which in Drupal 9.1 should be where it calls `@trigger_error()`.

clayliddell:
I ran through the QA steps and didn't encounter any errors so I think it passed?

PR #3589, opened 2021-07-28
Test DKAN upgrades in CI

### LINKED ISSUE #3556: Upgrade as well as install tests on CI
Our current CI tests run only on a new install scenario. We should be running any integration tests on an upgrade scenario as well.

- Create a parallel job that builds from 2.x and then upgrades to the commit being tested
- Run the cypress and PHP unit functional tests on this build

Per Janette this is a Priority 2 level

grugnog:
fixes #3556

- [x] Test coverage exists
- [x] Documentation exists (in step descriptions)

## QA Steps

- [ ] Review steps in https://github.com/GetDKAN/dkan/blob/3556-upgrade-tests/.circleci/config.yml to validate that they are correct for installing a stable version and upgrading to the target commit
- [ ] Review https://app.circleci.com/pipelines/github/GetDKAN/dkan?branch=3556-upgrade-tests that the steps are named and parallelized appropriately

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,15 +15,6 @@ jobs:
     steps:
       - checkout:
           path: dkan
-      - when:
-          condition: << parameters.upgrade >>
-          steps:
-            - run:
-                name: Switch to latest stable tag to upgrade from
-                command: |
-                  TAG="$(git --git-dir=dkan/.git describe --abbrev=0 --tags)"
-                  echo "Testing upgrade from tag: ${TAG}"
-                  git --git-dir=dkan/.git checkout "${TAG}"
       - run:
           name: Setup composer config
           command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,8 +1,47 @@
...
+            dktl dc up -d
+            sleep 10 && dktl install
+      - when:
+          condition: << parameters.upgrade >>
+          steps:
+            - run:
+                name: Switch to target commit and run DKAN upgrade
+                command: |
+                  git --git-dir=dkan/.git --work-tree=dkan checkout "${CIRCLE_SHA1}"
+                  dktl exec:composer -- require 'getdkan/dkan:dev-HEAD' --no-update --ansi
```
dafeder: @grugnog I'm not familiar with that `dev-HEAD` syntax -- is that a proper version constraint in composer? Are we confident this will always work?

.circleci/config.yml
```diff
@@ -1,8 +1,47 @@
...
+                  dktl make
```
dafeder: This may be counterintuitive but the idea of `dktl make` is you only run it once. It uses `composer install` and what you want here is an update I believe. I think this should change to just `dktl composer update`.

dafeder (changes requested):

grugnog:
@dafeder any idea why this test is now failing?

grugnog:
@dafeder Looks like whatever glitch was breaking the test has been resolved with a rebase (I ran the pipeline twice, just in case it was a fluke), so I think this is good to go!

.circleci/config.yml
grugnog: Updated to specify a SHA

.circleci/config.yml
grugnog: Updated

dafeder (approved):

PR #3596, opened 2021-07-30
Added 'checksum' column for 'dkan_metastore_resource_mapper' table.

### LINKED ISSUE #3583: Add checksum to filefetcher process
Once a remote file is fetched, generate an MD5 checksum and store it.

**Store Checksum**
File information is stored in the [dkan_metastore_resource_mapper](https://github.com/GetDKAN/dkan/blob/2.x/modules/metastore/src/Storage/ResourceMapperDatabaseTable.php) table, most files get stored with three "perspectives." We should add a new column to this table, "checksum", and upon completing the download/"fetch", generate an mD5 string and add it to that column for the "local" file perspective.

**Dispatch Event**
Once the file has completed the download and the checksum stored. We need to dispatch an event that passes the filepath and checksum value. Custom modules can then subscribe to the event and run a comparison to the remote file.

Per Janette this is a Priority 2 level

klimp-drupal:
fixes [GetDKAN/dkan/issues/3583]

## QA Steps

- [ ] run `drush updb`
- [ ] make sure that the 'checksum' field is added to the `dkan_metastore_resource_mapper` table
- [ ] create a new dataset
- [ ] run `drush queue:run datastore_import`
- [ ] query the db `select * from dkan_metastore_resource_mapper;` and make sure that the checksum is added to the `local_file` record


klimp-drupal:
I haven't finished the tests. I stumbled over the part where I need to get a string from `md5_file()`  called on a fake file path. Will need to mock a php stream wrapper

klimp-drupal:
The ticket also has the "dispatch event" part. We already have an event dispatched right after saving the `local_file` record to the resource mapper table - https://github.com/GetDKAN/dkan/blob/c869306e863f6fd4d96393d5ae6a2d8475396f08/modules/metastore/src/ResourceMapper.php#L68
The event gets a resource as the parameter. It already has the file path along with the checksum. I don't think we need another event specifically for the checksum and file path.

PR #3600, opened 2021-08-02
Improve behaviour when detecting MIME type

### LINKED ISSUE #3599: CSV/TSV files not being queued for datastore when site unable to make HTTP requests to itself
## Describe the bug
This is happening because the site is trying make HTTP lookups in `Referencer.php:getMimeType()` to determine the file MIME type from the HTTP headers. When a site has access control, SSO gateway or is otherwise unable to make a HTTP request to the file URL it silently drops adding the file to the datastore - there is a secondary issue of lack of user feedback here.

## Steps To Reproduce

- Enable basic auth on a new install and create a dataset with a CSV file
- Check the datastore queue, which will be empty

## Expected behavior

The datastore queue will have the csv file waiting to load

### LINKED ISSUE #3601: Improve variable naming in hostify()
This function uses the variables $public_url, $host, $url and $parsedUrl. The first 2 of these refer to the servers own files directory URL and server hostname - the latter 2 refer to the resource - this is confusing and led to a bug in implementing #3600.

Suggest we change to something more like:
```
$public_url -> $serverPublicFilesUrl
$host -> $serverHost
$url -> $resourceUrl
$parsedUrl -> $resourceParsedUrl
```

grugnog:
fixes #3599 

This PR adds much more robust mimetype detection for both local and remote files.

## QA Steps

- [ ] `dktl demo`
- [ ] Run dktl drush queue:list
- [ ] Make the files directory inaccessible via http: `echo 'Require all denied' >> docroot/sites/default/files/.htaccess`
- [ ] Login
- [ ] Perform the following with a local CSV File and with a remote CSV File:
  - [ ] Create a new dataset
  - [ ] Run dktl drush queue:list
  - [ ] Confirm the datastore_import queue shows 1 more item
  - [ ] Run the datastore_import queue
  - [ ] Ensure a datastore was created for the new dataset


grugnog:
Added a first pass at a lookup using Drupal managed files. This isn't ideal - we can't use the resource mapper as the MIME type is needed to create the resource mapper entry in the first place. It would be preferable to capture the MIME type when the file is saved and include that in the metadata, however that seems quite distant and will require some more intensive surgery.

clayliddell:
This PR is now ready for review.
I'm not sure what to do about the 1 remaining issue in CodeClimate since it seems it would not be in our best interest to fix it. I tried adding a `@codeCoverageIgnore` tag the the docblock of the offending method, however that didn't seem to fix the issue. I'm open to suggestions.

clayliddell:
Regarding the two CodeClimate issues: the `object` type annotation has been supported since PHP 7.2.

Additionally, the Drupal maintainers recommend you run at least PHP 7.3 with Drupal 8.9, since PHP 7.2 and below have already reached EOL.

PR #3604, opened 2021-08-05
DKAN with Drupal 9.2

### LINKED ISSUE #3615: Drupal 9.2 support
Get tests passing on Drupal 9.2.x.

thierrydallacroce:
Issue #3615

## Merging steps

1. Once this PR is approvecd, merge https://github.com/GetDKAN/recommended-project/pull/4
2. On this PR, set DKTL_VERSION back to whatever the latest release of dkan_tools is in .circleci/config.yml
3. Close getdkan/dkan-tools#301 (do not merge)


dafeder:
Continue to pick at this. Seems like most of the breaking changes have to do with the way we mock DB connections.

dafeder:
After addressing what I think are all the deprecation notices from our own code, I was still getting them from other contrib modules, and even core if I am understanding the output correctly. So while it makes me a bit nervous I've simply disabled deprecation warnings (see changes to phpunit.xml).

dafeder:
Something to note - because of the switch to `Drupal\Component\EventDispatcher\Event` this will completely break Drupal 8.9 compatibility. Because 8.9 has reached end of life I think I'm alright with this, but we should discuss and make a collective decision. Let's not merge this till we have discussed further. If we do proceed without doing more work to achieve backwards compatibility, we should

1. Reflect this in all the .info.yml files (9.1)
2. Bump to 2.12.x after merge
3. 2.12.0 release notes should be loud about this

modules/common/tests/src/Traits/ServiceCheckTrait.php
```diff
@@ -8,7 +8,7 @@
 use MockChain\Options;
 
 /**
- *
+ * @todo Fix for non-service arguments like '%app.root%'.
```
clayliddell: @dafeder @thierrydallacroce This seems important. Is this something we need to fix before merging this PR?

clayliddell (commented):
The changes in this PR look good to me. I've left one comment below on something which may need attention.

modules/common/tests/src/Traits/ServiceCheckTrait.php
dafeder: Depends on your idea of important -- as long as we keep suppressing the deprecation notices this won't really be an issue till D10, but it's also a very easy fix so now that you've flagged it I think we might as well do it.

modules/common/src/Storage/SelectFactory.php
```diff
@@ -226,51 +229,49 @@ private function propertyToString($property) {
...
+   * @param Drupal\Core\Database\Query\Select|Drupal\Core\Database\Query\Condition $statementObj
    *   Drupal DB API select object.
    * @param object $conditionGroup
    *   A condition from the DKAN query object.
    */
-  private function addConditionGroup($db_query, $conditionGroup) {
+  private function addConditionGroup($statementObj, $conditionGroup) {
     $groupMethod = "{$conditionGroup->groupOperator}ConditionGroup";
-    $group = $db_query->$groupMethod();
+    $group = $this->dbQuery->$groupMethod();
```
dafeder, line 274: In case you were wondering, all this variable/property refactoring in this class is because condition objects are not supposed to be instantiated with the `new` keyword anymore, which is what happens when you run one of these condition group methods directly from a condition object. I think it's just a bug in core but whatever it's about this fixes it; always calling the method directly from the main Select object (these methods are used to both add conditions directly to the select object and to recursively add nested conditions to other conditions). See https://www.drupal.org/node/3159568

dafeder:
Note - we should make minimum core requirement 9.1

PR #3607, opened 2021-08-05
Properly escape headers in MySQL fast import

### LINKED ISSUE #3606: MySQL fast import does not properly quote headers in the `LOAD DATA LOCAL INFILE` sql command.
## Describe the bug
Whenever dataset files are imported using the MySQL fast import module, the headers used for the tables created using the `LOAD DATA LOCAL INFILE` command are not properly quoted using backticks (\`). This can cause problems if a column name begins with an integer (see the [identifiers section](https://dev.mysql.com/doc/refman/8.0/en/identifiers.html) of the MySQL manual for reference).

## Steps To Reproduce
Create a dataset with a CSV file which has number column headers and observe how the `datastore_import` queue job created fails with an error message resembling:
```
SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax;
check the manual that corresponds to your MySQL server version for the right syntax to use near '2005,2006,2007,2008,2009,2010,2011,2012,2013,2014,2015) SET record_number = NULL' at line 1:
LOAD DATA LOCAL INFILE '/mnt/gfs/datamedicaiddev/sites/default/files/resources/88451bb21cd9239d8dc71a8a112ff1dc_1628201846/hurricanes.csv'
INTO TABLE datastore_1f8bdb25e92099b6d29fabbe4874eb22
FIELDS TERMINATED BY ','
ENCLOSED BY '\"'
LINES TERMINATED BY '\n'
IGNORE 1 ROWS (month,average,2005,2006,2007,2008,2009,2010,2011,2012,2013,2014,2015)
SET record_number = NULL;
```

## Expected behavior
The `datastore_import` queue job should successfully import the CSV file without the above error.

clayliddell:
fixes #3606

This PR adds a one-character prefix (`_`) to incoming column names that are not compatible with mysql. These include:

- Numeric column names
- Column names matching any [MySQL reserved words](https://dev.mysql.com/doc/refman/5.7/en/keywords.html).

This is essentially a stopgap, as [Drupal 9 takes care of this for us](https://www.drupal.org/project/drupal/issues/2986452). Once we drop support for Drupal 8 (don't worry no plans for this yet) we can revert this.  

## QA Steps

- [x] Ensure you're on the latest version of `GetDKAN/datastore`.
  - [x] Create a dataset with a resource file which has column names starting with integers.
  - [x] Ensure you're able to import the newly created dataset.
  - [x] Ensure all of the columns in the newly created dataset have a `d_` prefix and have been properly escaped.
- [x] Switch to the `34-allow-column-names-for-table-fields` branch on `GetDKAN/datastore`.
  - [x] Create a dataset with a resource file which has column names starting with integers.
  - [x] Ensure you're able to import the newly created dataset.
  - [ ] Ensure all of the columns in the newly created dataset have their original table names.

## Known issues

Test coverage for the mysql importer is still incomplete. This module will remain marked as "experimental".

### DIFF
--- a/modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
+++ b/modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
@@ -44,7 +44,7 @@ class MysqlImport extends Importer {
     $storage = $this->dataStorage;
     $storage->count();
 
-    $sqlStatementLines = $this->getSqlStatement($filename, $storage, $header);
+    $sqlStatementLines = $this->getSqlStatement($filename, $storage, $headers);
 
     $sqlStatement = implode(' ', $sqlStatementLines);
 
@@ -73,7 +73,7 @@ class MysqlImport extends Importer {
   /**
    * Properly escape and format the supplied list of column names.
    *
-   * @param string[] $columnNames
+   * @param string[] $columns
    *   List of column names.
    *
    * @return array
--- a/modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
+++ b/modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
@@ -69,7 +69,8 @@ class MysqlImport extends Importer {
     'undo', 'union', 'unique', 'unlock', 'unsigned', 'update', 'usage', 'use',
     'using', 'utc_date', 'utc_time', 'utc_timestamp', 'values', 'varbinary',
     'varchar', 'varcharacter', 'varying', 'virtual', 'when', 'where', 'while',
-    'window', 'with', 'write', 'xor', 'year_month', 'zerofill'];
+    'window', 'with', 'write', 'xor', 'year_month', 'zerofill',
+  ];
 
   /**
    * Override.
--- a/modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
+++ b/modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
@@ -83,11 +83,14 @@ class MysqlImport extends Importer {
     return array_replace([], ...array_map(function ($column) {
       // Sanitize the supplied table header to generate a unique column name.
       $header = $this->sanitizeHeader($column);
-      // Prepend "d_" to the table column name to prevent numeric column name
-      // parsing issues.
-      //
-      // @see https://github.com/GetDKAN/dkan/issues/3606
-      $header = 'd_' . $header;
+
+      if (is_numeric($header)) {
+        // Prepend "_" to fully numeric table column name to prevent column name
+        // parsing issues - can be dropped after move to Drupal 9.
+        // @see https://github.com/GetDKAN/dkan/issues/3606
+        $header = '_' . $header;
+      }
+
       // Truncate the generated table column name, if necessary, to fit the max
       // column length.
       $header = $this->truncateHeader($header);
--- a/modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
+++ b/modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
@@ -78,7 +78,10 @@ class MysqlImport extends Importer {
         $new = $strings[0] . "_{$token}";
       }
 
-      $row[] = $new;
+      // Enclose all headers in backticks to allow non-standard headers.
+      //
+      // @see https://dev.mysql.com/doc/refman/8.0/en/identifiers.html
+      $row[] = '`' . $new . '`';
     }
     return $row;
   }
--- a/modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
+++ b/modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
@@ -14,6 +14,13 @@ use Procrastinator\Result;
 class MysqlImport extends Importer {
 
   /**
+   * The maximum length of a MySQL table column name.
+   *
+   * @var int
+   */
+  protected const MAX_COLUMN_LENGTH = 64;
+
+  /**
    * Override.
    *
    * {@inheritdoc}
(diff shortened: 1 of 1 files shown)


grugnog:
I tested this on my local and it's working for me!

grugnog (changes requested):
While this works (and unblocks users for now), we shouldn't be building SQL through concatenation (as in getSqlStatement) since it is vulnerable to SQL injection. I think we should fix that here by switching to $database->query() and [placeholder arrays](https://www.drupal.org/docs/drupal-apis/database-api/static-queries#placeholder-arrays) for the headers, which will ensure they are properly quoted and escaped.

clayliddell:
This PR and https://github.com/GetDKAN/datastore/pull/35 complement each other. They should be tested both together, and separately.

clayliddell:
This is a non-BC-breaking implementation of the "option b" suggested by @grugnog (using a `d_` prefix to allow columns starting with an integer).

The https://github.com/GetDKAN/datastore/pull/35 PR which I suggested be tested with this PR adds support for preview column names, which differ from SQL table column names, and are derived from the header row of the CSV file provided for a dataset.

### Update

I just found a dataset that preview column names are not being generated properly for:
https://people.sc.fsu.edu/~jburkardt/data/csv/hurricanes.csv

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
```diff
@@ -61,26 +71,74 @@ private function getDatabaseConnectionCapableOfDataLoad() {
...
-      $row[] = $new;
+  private function generateTableHeaders(array $columns): array {
+    return array_replace([], ...array_map(function ($column) {
+      // Sanitize the supplied table header to generate a unique column name.
+      $header = $this->sanitizeHeader($column);
+      // Prepend "d_" to the table column name to prevent numeric column name
+      // parsing issues.
+      //
+      // @see https://github.com/GetDKAN/dkan/issues/3606
+      $header = 'd_' . $header;
```
grugnog: Will this prefix show up in the api & front-end? I wonder if we should be trying to remove this in column names for reads/queries, since it's not really relevant to end users? I also wonder if the regular importer should do add this prefix too for consistency?

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
```diff
@@ -61,26 +71,74 @@ private function getDatabaseConnectionCapableOfDataLoad() {
...
+   * @returns string
+   *   Truncated column name.
+   */
+  protected function truncateHeader(string $column): string {
+    // If the supplied table column name is longer than the max column length,
+    // truncate the column name to 5 characters under the max length and
+    // substitute the truncated characters with a unique hash.
+    if (strlen($column) > self::MAX_COLUMN_LENGTH) {
+      $field = substr($column, 0, self::MAX_COLUMN_LENGTH - 5);
+      $hash = substr(md5($column), 0, 4);
```
grugnog, line 192: Similar question to the d_ prefix here - should this be made transparent to users?

grugnog (commented):
I had a couple of questions about the effect of these changes

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
grugnog: I should have looked at https://github.com/GetDKAN/datastore/pull/35 first! I think that addresses this for the most part, except perhaps that we may still want to update the regular importer with the same heuristic?

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
dafeder: Not completely up to date on this issue, but in general, users want control over their columns and column names. If anything comes out different than it went in, people are going to be unhappy.

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
clayliddell: Yes, agreed. The regular importer also seems to be afflicted by the same bug as well.

dafeder: Maybe something for another PR, and we can do this another time, but I think there is consensus that we should bring the datastore library into DKAN, as it so far seems to provide little value outside DKAN and the separation makes development harder.

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
clayliddell: This particular line pre-existed this PR. It's a constraint of MySQL. MySQL column names can be no longer than 64 characters and this appears to solve that constraint by replacing the last 4 characters with a segment from an md5 hash generated using the whole column name.
Regardless, this PR in conjunction with https://github.com/GetDKAN/datastore/pull/35 solves this issue by using the actual CSV column name provided as the preview column name for datasets while only using the truncated `d_` prefixed column name in MySQL.
TL;DR: the user shouldn't encounter a modified column name with this approach.

grugnog:
After my comment on https://github.com/GetDKAN/datastore/pull/35#issuecomment-895594753 @dafeder had the idea of only prefixing fully-numeric columns, rather than all of them. Since this has a much more minimal user and API impact (especially since fully-numeric columns aren't working previously) it seems like it can probably go ahead even without https://github.com/GetDKAN/datastore/pull/35 (although that does seem to be working also).

dafeder:
It looks like we need to do this for column names that use reserved words (like "index") as well

grugnog:
Since this now has more minimal impacts we are merging without GetDKAN/datastore#35 for now. Still experimental status, so no tests - created #3608 as a follow-up for that.

Issue #3603, opened 2021-08-05
Distribution files that failed to import will prevent future imports of the same file


janette:
## Describe the bug
If a dataset is created with a csv distribution, and there is a failure on the resource importer, attempting to use that file again will point the new use case to the failed datastore import.

## Steps To Reproduce
- make your sites/default/files directory unwritable
- create a dataset with a csv distribution
- run cron
- fix write access to sites/default/files
- create a new dataset and use the same csv file for the distribution

grugnog:
Was finally able to reproduce this as part of working on #3605 - using "https://download.medicaid.gov/data/newNADACPrice.csv" as a remote URL on 2.x creates a distribution with an MIME type that will not trigger the importer.

If you then fix that issue and create a new dataset reusing the same "https://download.medicaid.gov/data/newNADACPrice.csv" URL you will notice that no new distribution is created. This is because it is reusing the reference from the prior dataset.

If you go and look at the original distribution you can see that it also hasn't been updated with the new MIME type.

Because no new distribution or distribution version is being stored the ResourceMapper:EVENT_REGISTRATION event is not triggered, which is what the DatastoreSubscriber listens to for queuing import tasks.

The code in question is in https://github.com/GetDKAN/dkan/blob/d1b8249da2719ebbe68b4a7d93172ac47394ffd5/modules/metastore/src/Reference/Referencer.php#L196

The line `if ($info[0]->perspective == Resource::DEFAULT_SOURCE_PERSPECTIVE && resource_mapper_new_revision() == 1) {` returns false, causing it to return with the existing datastore identifier.

grugnog:
Added a PR that fixes this - I am not very sure if this is the right place/approach though - this approach only accounts for changing MIME types, I am not sure if there are other things we need to look at also. In general it seems quite mysterious if/how a new version would ever get saved in the first place?

dafeder:
@grugnog works great but needs tests, see PR. I can write the tests if you're busy with other things now.

WRT your question, this PR would ensure that a new resource version (which means a new datastore import job is queued) is created under the following conditions:

1. The downloadURL changes
2. One of the "triggering" dataset properties changes (see #3552)

If the downloadUrl changes, that creates a completely new resource, which will have essentially the same effect from the user standpoint. So these are the three ways to tigger a datastore re-import. Obviously, this all needs to be documented more clearly!

PR #3613, opened 2021-08-11
Set csv/tsv mime types based on format field.

### LINKED ISSUE #3605: If remote server provides non csv/tsv MIME type the file is not imported
## Describe the bug

Some web servers don't serve the correct MIME type for csv/tsv files. An example of this is Github raw, which serves text/plain for all files.
Because the remote server content type header is currently the only source for if the file should be imported or not, this means these files are not queued, with no indication for users as to why.

## Steps To Reproduce

* Create a new dataset
* Set the "format" to csv
* Select "Remote file URL"
* Enter https://raw.githubusercontent.com/datasets/unece-units-of-measure/master/data/units-of-measure.csv as the URL
* Save
* Run drush queue:list
* No items will be listed as queued

## Expected behavior

- The "format" field should probably take priority if it is set
- We may want to add some kind of feedback for users if their dataset was queued for import or not (and why not)
- Later on we may also want to look at the filename or check the first few KB of data to check the format

grugnog:
fixes #3605 

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

- [ ] Add a dataset - (a) set format to "csv" and (b) use the remote URL `https://download.medicaid.gov/data/newNADACPrice.csv`
- [ ] Run the import queue

With this patch, the data should be imported. Without the patch **or** without manually setting a "csv" format the import job will not be created.

Note that if you have an existing dataset with this URL that was created before this patch you will need to test in combination with the PR to #3603


janette (approved):
@grugnog this is a nice hotfix, tested and got a successful import

dafeder:
Added a bunch of new tests to cover all scenarios.

PR #3614, opened 2021-08-11
Save new distribution version when MIME type changes

### LINKED ISSUE #3603: Distribution files that failed to import will prevent future imports of the same file

grugnog:
Updating the distribution saving logic to handle the case of a distribution MIME type being updated.

fixes #3603 

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

- [ ] Switch to the 2.x branch
- [ ] Create a dataset using the remote URL https://download.medicaid.gov/data/newNADACPrice.csv
- [ ] Check the distribution mime type is `application/octet-stream`
- [ ] Run the import queue - no import should run
- [ ] Switch to this branch AND cherry pick the commit from #3613
- [ ] Create a new dataset using the remote URL https://download.medicaid.gov/data/newNADACPrice.csv
- [ ] Check the distribution mime type is now `text/csv`
- [ ] Run the import queue - the import should run and data should be available in preview

### DIFF
--- a/modules/common/src/Resource.php
+++ b/modules/common/src/Resource.php
@@ -94,6 +94,13 @@ class Resource implements \JsonSerializable {
   }
 
   /**
+   * Change MIME type.
+   */
+  public function changeMimeType($newMimeType) {
+    $this->mimeType = $newMimeType;
+  }
+
+  /**
    * Private.
    */
   private function createCommon($property, $value) {
--- a/modules/metastore/src/Reference/Referencer.php
+++ b/modules/metastore/src/Reference/Referencer.php
@@ -204,7 +204,7 @@ class Referencer {
       if (isset($info[0]->identifier)) {
         /** @var \Drupal\common\Resource $stored */
         $stored = $this->getFileMapper()->get($info[0]->identifier, Resource::DEFAULT_SOURCE_PERSPECTIVE);
-        $downloadUrl = $this->handleExistingResource($info, $stored);
+        $downloadUrl = $this->handleExistingResource($info, $stored, $mimeType);
       }
     }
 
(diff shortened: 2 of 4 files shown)


janette:
That works, I got a new record in dkan_metastore_resource_mapper table for the updated mimetype.

dafeder:
I think we should just reimport if the distribution has changed in any way, less field-specific, simpler...

grugnog:
@dafeder based on my testing, I think new distributions are already being created if any of the user editable fields change, so that part is already behaving.

I think the addition here is only needed because the MIME field is not purely user set, but has some additional detection logic that happens as part of the reference process - so it's not triggering the mechanism that detects a change. I haven't been able to wrap my head around how/where the change detection mechanism is - it may be there is a more straightforward way of triggering this.

dafeder:
@grugnog right -- new distributions are created on any JSON change but do not trigger a re-import. The motivation for suggesting this is that we really should be trying to keep fieldnames out of the DKAN code. We had made some progress on this in the spring and have definitely backslid a bit recently due to time pressure on just getting certain features working, but I think with any new code we merge we should think if there is an alternative with a reasonable LOE that would let us avoid adding more hardcoded field names in the codebase.

Another, probably better approach, would be to extend the new admin interface for chosing dataset properties that trigger a reimport (#3554) to do the same with distribution fields.

dafeder:
Have come back to this after reviewing and merging #3613 and now realize that the concerns I was raising here are really more specific to _that_ ticket, which we've merged for now anyway. We should go back to that at some point and abstract out all the distribution/field-specific stuff if possible, but this one is actually fine! 🙃

dafeder:
Rebased and tests were failing. Found a logic error, fixed (not clear on how this was passing before actually, but 🤷🏽 ).

dafeder (changes requested):
Although we're getting a 100% score on coverage for the diff here, I don't feel the tests are sufficient. We manage to run all the new lines in our existing functional tests because the check for a changed mimetype in just being added as an OR condition to something that was already covered. We should have an additional test, preferably a unit test in ReferencerTest, that will check to see if a new distribution is triggered by changing the format but not  by changing another field (e.g. "title").

PR #3617, opened 2021-08-13
Add support for non-LF (`\n`) newline character sequences

### LINKED ISSUE #3616: MySQL Fast import improperly imports CSV files with non-LF (`\n`) carriage returns.
## Describe the bug
The MySQL fast import module does not currently support Unix carriage return sequences. It only supports the `\n` or LF newline character.

See the following line from MysqlImport for reference:
https://github.com/GetDKAN/dkan/blob/d1b8249da2719ebbe68b4a7d93172ac47394ffd5/modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php#L95

If a non-LF carriage return character sequence is used, the CSV file will be imported, but not properly.

## Steps To Reproduce
1. Attempt to create a dataset using a CSV file with the CRLF or just CR carriage return sequence.
2. Observe how the some rows are combined or extended in strange manners.

## Expected behavior
Any combination of LF (`\n`), CRLF (`\r\n`), and CR (`\r`) carriage return sequences should be supported by the MySQL Fast Import module. We can use the first line of the CSV file for detecting the appropriate character sequence.

clayliddell:
fixes #3616 

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

- [ ] Create a dataset using a CSV resource file with Unix style newlines (`\n`) (can be generated using standard CSV exports in modern Excel).
- [ ] Ensure the dataset is properly imported, and is displayed properly in the preview table.
- [ ] Create a dataset using a CSV resource file with Windows style newlines (`\r\n`) (can be generated using "Windows comma separated" CSV exports in modern Excel).
- [ ] Ensure the dataset is properly imported, and is displayed properly in the preview table.
- [ ] Create a dataset using a CSV resource file with MacOS style newlines (`\r`) (can be generated using Excel 2011 for Mac).
- [ ] Ensure the dataset is properly imported, and is displayed properly in the preview table.

If you're having trouble creating files with any of the above line endings, contact me and I can send them to you for testing.


modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
```diff
@@ -110,6 +118,31 @@ protected function runIt() {
...
+   * Attempt to detect the EOL character for the given line.
+   *
+   * @param string $line
+   *   Line being analyzed.
+   *
+   * @return string|null
+   *   The EOL character for the given line, or NULL on failure.
+   */
+  protected function getEol(string $line): ?string {
+    $eol = NULL;
```
janette, line 178: I think the default should be unix endings rather than null

```
TypeError: Argument 1 passed to Drupal\datastore_mysql_import\Service\MysqlImport::getEol() must be of the type string, null given, called in /var/www/dkan/modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php on line 101 in /var/www/dkan/modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php on line 130 #0 /var/www/dkan/modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php(101): Drupal\datastore_mysql_import\Service\MysqlImport->getEol(NULL)
```

janette (changes requested):

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
clayliddell: @janette I think the latest changes I just pushed should fix this error, but since I've thus far been unable to replicate it, I'll need you to confirm.

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
```diff
@@ -110,6 +126,63 @@ protected function runIt() {
...
+   */
+  protected function getFirstLineFromFile(string $file_path): string {
+    // Read the first (header) line from the CSV file.
+    $f = fopen($file_path, 'r');
+    // Ensure the file could be successfully opened.
+    if (!isset($f) || $f === FALSE) {
+      throw new FileException(sprintf('Failed to open resource file "%s".', $file_path));
+    }
+    // Attempt to retrieve the first line from the resource file.
+    $header_line = fgets($f);
```
grugnog, line 157: Note that auto-detect-line-endings is disabled by default - not site how typical it is to configure it?
https://www.php.net/manual/en/filesystem.configuration.php#ini.auto-detect-line-endings

clayliddell: @grugnog Good point! I've updated the the `MysqlImport::getFirstLineFromFile()` method to temporarily enable this ini setting before opening the file.

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
```diff
@@ -134,7 +214,7 @@ private function getDatabaseConnectionCapableOfDataLoad() {
   private function generateTableHeaders(array $columns): array {
     return array_replace([], ...array_map(function ($column) {
       // Sanitize the supplied table header to generate a unique column name.
-      $header = $this->sanitizeHeader($column);
+      $header = $this->sanitizeHeader($column ?? '');
```
grugnog, line 218: This looks odd - could use a comment explaining?

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
grugnog: Looks good - does this setting persist through the fgets() that actually reads what it deems a "line"... perhaps you tested, but the docs aren't clear!

grugnog (changes requested):

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
clayliddell: This has to do with some strange undocumented behavior I encountered with the `str_getcsv()` function in PHP.
Essentially, if an empty string is encountered, it returns an array with a single `NULL` element; in any other case, an empty string is used to represent empty headers.
This takes that odd behavior into account by null-coalescing any potentially `NULL` column names to an empty string.
I'll add a comment to the code too.

clayliddell:
> Looks good - does this setting persist through the fgets() that actually reads what it deems a "line"... perhaps you tested, but the docs aren't clear!

Yes, I suppose this is undocumented behavior which could potentially change, but the newline detection behavior appears to currently be determined during `fopen()`.

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
clayliddell: Yes, I suppose this is undocumented behavior which could potentially change, but the newline detection behavior appears to currently be determined during `fopen()`.

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
clayliddell: I moved the check for `NULL` column names up to where the headers are first fetched and added a comment in an attempt to make things a bit more clear.

janette:
@clayliddell @grugnog tested with problem files on the client dev site and works like a charm.

PR #3621, opened 2021-08-14
Dev docs

fmizzell:
As I am going through our tests and the whole codebase, I still, often, feel like we do not have a north star nor well defined boundaries to help us organize our codebase. I think that some documentation that more specifically looks at what a module is supposed to do and how it is organize could serve as the boundary setter, and also as a point of conversation to what belongs or does not.

This PR creates a possible way for us to start organizing this type of documentation.


modules/datastore/modules/datastore_mysql_import/README.md
```diff
@@ -1,12 +1,8 @@
-# DKAN Datastore MySQL Import
+# DKAN Datastore MySQL Import {#mysqlimport}
 
-This is an **expiremental** module to import CSVs into the DKAN datastore using 
-MySQL's native LOAD DATA function, similar to running a `mysqlimport`
-command. It is over 5000% faster than DKAN's default datastore importer, but
-requires a MySQL database and that the Drupal DB user have permissions to use
+This an **expiremental** module to import CSVs into the DKAN datastore using MySQL's native LOAD DATA function, similar to running a `mysqlimport` command. It is over 5000% faster than DKAN's default datastore importer, but requires a MySQL database and that the %Drupal DB user have permissions to use
```
clayliddell: Is this still considered to be experimental? We should also probably mention that it doesn't currently have TSV support.

docs_assets/pages/guide_data_dictionaries.md
```diff
@@ -1,8 +1,9 @@
-@page tut_data_dictionaries How to use Data Dictionaries
+# How to use Data Dictionaries {#guide_data_dictionaries}
```
clayliddell, line 1: I love that we now have this.

modules/metastore/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Entrypoints
+
+There are 3 main ways to interact with the metastore module:
+- Drush commands
+- Endpoints
+- Services
+
+### Drush commands
+To see the Drush commands that are made available by this module look a the Drupal::metastore::Commands::MetastoreCommands class.
```
clayliddell:
```suggestion
To see the Drush commands that are made available by this module, look at the Drupal::metastore::Commands::MetastoreCommands class.
```

clayliddell (approved):

PR #3624, opened 2021-08-17
Fix dataset files with spaces in name not importing

### LINKED ISSUE #3623: Resource files with spaces in their names fail to import
Search existing issues to make sure the bug has not been reported already.

## Describe the bug
When attempting to upload a dataset file with a space in it's name, the following error is encountered (and the dataset fails to import):
```
Import for 00c2a6408acd9130792913c673c1873c__1629214354 returned an error: Error opening file: public://uploaded_resources/some_resource%20with_a_space.csv
```
## Steps To Reproduce
Create a dataset with a resource file which has a space in it's name.

## Expected behavior
The dataset should be successfully imported.

clayliddell:
fixes #3623

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

- [x] Create a dataset with a CSV file which has special characters in it's name (see https://en.wikipedia.org/wiki/Percent-encoding for reference).
- [x] Ensure the dataset is successfully imported by the `datastore_import` queue.

### DIFF
--- a/modules/common/src/UrlHostTokenResolver.php
+++ b/modules/common/src/UrlHostTokenResolver.php
@@ -57,8 +57,8 @@ class UrlHostTokenResolver {
    *   Resolved public file path.
    */
   public static function resolveFilePath(string $resourceUrl): string {
-    return preg_replace('/^' . preg_quote(self::getServerPublicFilesUrl(), '/') . '/',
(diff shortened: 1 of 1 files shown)


grugnog (approved):
Tested with a wide range of filenames, including characters with hard and pre-encoded spaces, files with different parens, backticks, backslash, star and more. All behaving well.

PR #3640, opened 2021-08-27
Prevent datastore_import queue worker from being run by cron

### LINKED ISSUE #3639: A single dataset can be imported multiple times by the same queue job
## Describe the bug
Due to a bug in the implementation of the `\Drupal\Core\Cron` service, a queue item is only ever leased for 1 second (there's a Drupal Core issue to resolve this [here](https://www.drupal.org/project/drupal/issues/2893933)).
This can be a problem if a single dataset takes too long to import and a second instance of the `\Drupal\Core\Cron` service is run on another thread, since it could result in an attempt being made to import the same dataset twice. In such a case, a datastore could be polluted with duplicate data from the dataset file.  

## Steps To Reproduce
1. Create a new dataset with a large resource file (>= 100 MB).
2. Run cron using Drush (`drush cron`).
3. Run another instance of cron using Drush (`drush cron`).
4. Wait until both cron jobs have completed.
5. Observe duplicate rows in the imported dataset.

## Expected behavior
A single `datastore_import` queue job should only ever be run once.

clayliddell:
Since the Drupal Core cron service doesn't seem to handle jobs in the `datastore_import` queue properly, it should be removed in favor of using the `drush queue:run` command supplied by Drush.

A potential remedy may be supplied by [Drupal Core issue #2893933](https://www.drupal.org/project/drupal/issues/2893933), however it has yet to be merged.

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

- [ ] Create a new dataset with a large resource file (>= 100 MB).
- [ ] Run cron using Drush (`drush cron`).
- [ ] Run another instance of cron using Drush (`drush cron`).
- [ ] Wait until both cron jobs have completed.
- [ ] Ensure no duplicate rows were imported into the datastore for the created dataset.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,8 @@
         "ilbee/csv-response": "^1.0",
         "oomphinc/composer-installers-extender": "^2.0",
         "stolt/json-merge-patch": "^1.0",
-        "drupal/openapi_ui": "^1.0@RC"
+        "drupal/openapi_ui": "^1.0@RC",
+        "drush/drush": "^10.1"
     },
     "require-dev": {
         "getdkan/mock-chain": "^1.3.0",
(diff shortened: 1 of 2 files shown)


clayliddell:
Documentation is absolutely essential for this PR, since it will require a change in the way DKAN is set up on the hosting environment.

This may also warrant a major release due to this being a fairly large change in behavior.

PR #3643, opened 2021-08-30
Fix how the Drupal cron service processes `datastore_import` jobs.

clayliddell:
This is an alternative approach to issue #3639 (as opposed to fix outlined in PR #3640).

In this PR, a custom alternative cron service is used in place of `\Drupal\Core\Cron` which adds support for specification of a separate queue time-limit and queue item lease time.

I can think of two main benefit to this approach over #3640:
1. It maintains the single responsibility principle by allowing all queue jobs to be processed in the standard Drupal manner, rather than dividing the responsibility of processing queue jobs across more than one mechanism.
2. It does not require any additional setup on the environment level. As opposed to requiring some additional timed queue processing mechanism.

This approach may require additional maintenance if/when the Drupal Core cron queue processing mechanism is ever updated in Drupal Core, however #3640 would also require additional maintenance, albeit for a separate mechanism outside of Drupal entirely.

A side note: the `better_cron` module in this PR will no longer be necessary if/when [Drupal Core issue #2893933](https://www.drupal.org/project/drupal/issues/2893933) is merged.

## QA Steps

- [ ] Create a new dataset with a large resource file (>= 100 MB).
- [ ] Run cron using Drush (`drush cron`).
- [ ] Wait long enough for the queue time-limit to have expired, but for the item lease time to still be active for the `datastore_import` queue worker (1-2 minutes).
- [ ] Run another instance of cron using Drush (if there was only one job in the queue, the second instance should stop relatively quickly on account of not having any jobs to process).
- [ ] Wait until both cron jobs have completed.
- [ ] Ensure no duplicate rows were imported into the datastore for the created dataset.

### DIFF
--- a/modules/better_cron/better_cron.info.yml
+++ b/modules/better_cron/better_cron.info.yml
@@ -1,0 +1,6 @@
+name: Better Cron
+description: A Drupal Cron service implementation with support for separate queue time-limits and item lease times.
+type: module
+core: 8.x
+core_version_requirement: ^8 || ^9
+package: DKAN
--- a/modules/better_cron/better_cron.services.yml
+++ b/modules/better_cron/better_cron.services.yml
@@ -1,0 +1,14 @@
+services:
+  cron:
+    class: Drupal\better_cron\BetterCron
+    arguments: [
+      '@module_handler',
+      '@lock',
+      '@queue',
+      '@state',
+      '@account_switcher',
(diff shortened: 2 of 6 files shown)


clayliddell:
If this approach is accepted over #3640, the module should be broken out into a separate repo and maintained independently.

PR #3645, opened 2021-08-30
Fix datastore subscriber no triggering properties

dafeder:
We get some errors if the new "datastore triggering properties" configuration is not set at all.


thierrydallacroce:
Thank you, @dafeder. This PR does fix the issue in one of our custom tests. Merging.

dafeder:
We should probably be returning an empty array from get() there rather than needing to do this last-minute save, but this works for now.

PR #3653, opened 2021-09-03
API page caching with cache tags

### LINKED ISSUE #3635: Add cache tags to DKAN API responses
Include HTTP cache tags in API response headers and use purge module to invalidate upstream HTTP caches when the tagged resources change.

Drupal core uses the render array to aggregate cache tags from page components - we are serving JSON responses directly so we will need to determine the tags ourselves.

The general chain of entity references in DKAN 2 looks roughly like this:
dataset -> distribution -> resource

In general changes to entities will mean that entity and any entities to the _left_ (that were referencing the prior ID and now reference an updated ID) will need to have results invalidated. So, on any API responses we need to add tags going all the way down the chain to the _right_ so that changes will bubble up to the top/left. 

For instance, a dataset response will include tags for its distributions and resources. If a distribution changes, we will not invalidate the dataset tag directly, but will invalidate the distribution tag, which will invalidate and dataset responses that also have that tag. This is similar to how the Taxnomy Term-Node relationship is handled out of the box in Drupal.

Because we get a lot of tag and invalidation logic for free with the entity system, we will use default entity cache tags and modify our services and interfaces to make these more accessible.

## Implementation

Will require some trial and error, but thinking that refactoring our various WebServiceApi classes as actual Drupal controllers that return a [CacheableJsonResponse](https://api.drupal.org/api/drupal/core!lib!Drupal!Core!Cache!CacheableJsonResponse.php/class/CacheableJsonResponse/9.1.x) will allow us to more easily plug the whole api system into Drupal's cache framework. If that seems like too high of a LOE, we can just add some simple functionality to the JsonResponseTrait that manually adds tags to the `X-Drupal-Cache-Tags` header.

### LINKED ISSUE #3655: Trigger distribution invalidation on datastore import.
Now that we are using cache tags, we will potentially have a delay between an api response getting cached and then the datastore import completing. Upon import completion, all the related distribution tags should be invalidated.

dafeder:
* Adds page caching with node-based cache tags to the metastore, datastore, search and SQL controllers. 
* Invalidates the distribution cache tag when a datastore import is complete

Also, cleaned up some code in a few controllers, including fully renaming the metastore controller to start to be more consistent  with controller naming, be aware of possible conflicts.

## QA Steps

(better testing steps to come)

1. Try making various requests to endpoints from the four controllers above.
2. Check that on first request to a new URL, `X-Drupal-Cache: MISS` is in the headers. On a second request, `X-Drupal-Cache: HIT` should be present, and responses should be faster.
3. Check that requests with different GET parameters are cached separately, and that cachebusters work.
4. Check `SELECT cid, tags FROM cache_page` after a request and ensure that the expected tags are set for the URL you requested.
5. Edit a dataset, and check that the expected endpoints now respond with again `X-Drupal-Cache: MISS`
6. Create a new dataset, and request its metadata from the API before running the datastore_import queue. Ensure that it is cached.
7. Run the datastore_import queue and check again. Make sure that the dataset is no longer cached.
8. Replace the file on the above dataset and perform the same check again, before and after running the import queue.

## Known issues

* We now have quite a mess of code in different places building our API responses.
* Introduced a couple new hard-coded schema names (avoided as much as possible)


modules/metastore/src/MetastoreApiResponse.php
```diff
@@ -0,0 +1,133 @@
...
+      else {
+        throw new \InvalidArgumentException("Invalid cacheable dependency. " . print_r([$key => $item], TRUE));
+      }
+    }
+
+    if (isset($params)) {
+      $this->addContexts($cacheMetadata, $params);
+    }
+
+    $cacheMetadata->setCacheMaxAge($this->cacheMaxAge);
```
dafeder: No longer needed

grugnog:
Gave this a high level run through (including enabling purge and a setting up a purger that sends cache tag HTTP headers) and it seems to be working well for most of the endpoints listed.

- The metastore, datastore and SQL APIs all appear to be including both the dataset and the distribution node ID tags. I tested that both dataset and distribution node tags are invalidated on dataset edit (`dktl drush p:queue-browse` is helpful for this).
- The search API endponts are including cache_tags, but with a `node_list_data` tag - I tested that this is invalidated on node creation

A couple of things I think we do need to consider:

- I did discover that there aren't any cache tags for `api/1/datastore/imports` - this endpoint is called by the front-end and returns column names, so I think we would want it to be cached/invalidated.
- We also need to consider timing around datastore loads - if we don't do anything specific here the "empty datastore" page (if there is any request while the datastore is loading) will get cached indefinitely. So for this I think we may need invalidate the distribution node ID at the end of the queue job.
-  I think the same is the case for the resource URL - users could get a partially downloaded/copied file and that would get cached. This is more tricky as the file isn't served by Drupal - I think the fix for this is probably a separate ticket to leave datasets serving the most recent fully loaded distribution until the import queue job for a subsequent distribution marks it as ready. This would fix the resource URL issue as well as mean that public users don't see empty distributions while a new one is loading.

dafeder:
I've added tags to the imports API, see https://github.com/GetDKAN/dkan/pull/3653/files#diff-5d49dc64ecaa91891861d7b71afd6bb0aafed2530c553a172b53a2f31c6d2410R71. It only kicks in if you use a distiribution ID for now. This is ready for review, but I would say due to the issue you raise above it should be considered blocked by #3655

modules/metastore/tests/src/Unit/Reference/ReferenceLookupTest.php
```diff
@@ -0,0 +1,71 @@
...
+// use Drupal\metastore\Storage\NodeData;
+// use MockChain\Chain;
+// use MockChain\Options;
+// use PHPUnit\Framework\TestCase;
+// use Symfony\Component\DependencyInjection\ContainerInterface;
+
+// /**
+//  * @todo Finish these tests.
+//  */
+// class OrphanCheckerTest extends TestCase {
```
clayliddell: If these tests need to be added later, then I think a separate branch/PR should be created for them and referenced in this PR's ticket as opposed to just commenting out the progress that's been made.

clayliddell: Other than this, I think the code looks good.
This is more of a suggestion. I just want to make sure we don't get in the habit of pushing files which are never used and later forgotten to the code base.

dafeder: Yeah that's very fair, I'll split it off into its own PR draft.

PR #3654, opened 2021-09-07
Add irregular frequency option

janette:
## QA Steps

- [ ] Log in and navigate to `/node/add/data`
- [ ] Confirm that the options available for **Frequency** includes 'irregular'

### DIFF
--- a/schema/collections/dataset.json
+++ b/schema/collections/dataset.json
@@ -71,7 +71,7 @@
         "R/P4M",
         "R/P1W",
         "R/PT1H",
-        "irregular"
+        "Irregular"
       ],
       "enumNames": [
         "Decennial",
@@ -93,7 +93,7 @@
         "Three times a year",
         "Weekly",
         "Hourly",
-        "irregular"
+        "Irregular"
       ]
     },
     "describedBy": {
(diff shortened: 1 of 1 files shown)


schema/collections/dataset.json
```diff
@@ -70,7 +70,8 @@
         "R/P0.5M",
         "R/P4M",
         "R/P1W",
-        "R/PT1H"
+        "R/PT1H",
+        "irregular"
```
dafeder, line 74: Capitalize

dafeder (changes requested):

schema/collections/dataset.json
```diff
@@ -70,7 +70,8 @@
         "R/P0.5M",
         "R/P4M",
         "R/P1W",
-        "R/PT1H"
+        "R/PT1H",
+        "Irregular"
```
dafeder: As per https://resources.data.gov/resources/dcat-us/#accrualPeriodicity the actual value should be lowercase, we should just capitalize it in the display names. Will fix.

dafeder (changes requested):

PR #3662, opened 2021-09-13
Prevent premature removal of shared resources

### LINKED ISSUE #3647: Changing the resource URL on a dataset deletes the datastore for other datasets that used the same resource URL
## Describe the bug

Changing the resource URL on a dataset deletes the datastore for other datasets that used the same resource URL.

Just like we do for keyword references, check if other distributions using the resource reference before orphaning/deleting.

## Steps To Reproduce

- Create dataset A with resource URL https://people.sc.fsu.edu/~jburkardt/data/csv/ford_escort.csv
- Create dataset B with resource URL https://people.sc.fsu.edu/~jburkardt/data/csv/ford_escort.csv
- Run `dktl drush queue:run datastore_import`
- Observe that the datastore preview is visible on both dataset pages
- Edit dataset B and change the resource URL to https://people.sc.fsu.edu/~jburkardt/data/csv/tally_cab.csv
- Run `dktl drush queue:run resource_purger`
- Observe that dataset A (even though that was not edited) has had it's datastore removed

## Expected behavior

If a user edits one dataset other datasets are not changed.
If DKAN deduplicates resources (and datastores for these resources) behind the scenes this needs to be transparent to the user without unexpected side effects.

## Notes

When testing this we should test with and without identical distribution metadata (see #3648) - currently this reproduces in both cases.

Looking through the ResourcePurger service it seems like resources being shared by multiple distributions is not currently being taken into account.

clayliddell:
fixes #3647 

Test coverage exists: Yes
Documentation exists: N/A (since this is a bug fix)

## QA Steps

- [ ] Create dataset A with resource URL https://people.sc.fsu.edu/~jburkardt/data/csv/ford_escort.csv.
- [ ] Create dataset B with resource URL https://people.sc.fsu.edu/~jburkardt/data/csv/ford_escort.csv.
- [ ] Run cron.
- [ ] Observe that the datastore preview is visible on both dataset pages.
- [ ] Edit dataset B and change the resource URL to https://people.sc.fsu.edu/~jburkardt/data/csv/tally_cab.csv.
- [ ] Run cron.
- [ ] Observe dataset B is now showing the datastore preview for "tally_cab.csv".
- [ ] Observe dataset A is still showing the datastore preview for "ford_escort.csv".

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,8 +1,47 @@
-version: 2
+version: 2.1
+
+commands:
+  dktl:
+    steps:
+      - run:
+          name: Install/configure DKAN Tools (dktl)
+          command: |
+            if [ ! -d ~/dkan-tools ]; then git clone -b $DKTL_VERSION https://github.com/GetDKAN/dkan-tools.git ~/dkan-tools; fi
+            export PATH=~/dkan-tools/bin:$PATH
+            echo "export PATH=~/dkan-tools/bin:$PATH" >> $BASH_ENV
+  prepare_site:
+    parameters:
+      upgrade:
+        description: "If true, will upgrade a stable instance to the target commit"
+        default: false
+        type: boolean
+    steps:
+      - run:
+          name: Install Drupal
+          command: |
+            dktl dc up -d
+            sleep 10 && dktl install
+      - when:
+          condition: << parameters.upgrade >>
+          steps:
+            - run:
+                name: Switch to target commit and run DKAN upgrade
+                command: |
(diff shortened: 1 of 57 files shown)


dafeder (changes requested):
Code looks great and passes QA steps, but tests need to be extended to confirm behavior works as expected. Ideally both unit and functional tests but would be ok with just one of them for now. Perhaps the functional tests could be shoehorned into an existing test with minimal work.

dafeder (approved):

PR #3677, opened 2021-09-27
Update default search config to include modified date

dafeder:
## QA Steps
https://dkan-qa-3677.ci.civicactions.net/
- Go to the Datasets page
- Change the sort order to alphabetical, then back to "date"
- Confirm that the oldest dataset is displayed first (probably "Afghanistan Election Districts")

## Known issues

For this to work as expected out of the box we need to modify the front-end to:
1. Apply the date sort on initial page load, instead of needing the form element to be updated
2. Apply the descending order when using date, since that is most likely how people want it ordered (while still applying ascending order on alphabetical)

### DIFF
--- a/modules/metastore/modules/metastore_search/config/install/search_api.index.dkan.yml
+++ b/modules/metastore/modules/metastore_search/config/install/search_api.index.dkan.yml
@@ -1,3 +1,4 @@
+uuid: 55ab101c-3fc3-450b-b055-f3a8dc208437
 langcode: en
 status: true
 dependencies:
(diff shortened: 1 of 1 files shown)


modules/metastore/modules/metastore_search/config/install/search_api.index.dkan.yml
```diff
@@ -47,9 +52,9 @@ processor_settings:
       - description
       - keyword
       - publisher__name
+      - publisher__name
```
thierrydallacroce: Weird, duplicate `publisher__name`?

PR #3684, opened 2021-10-03
Translatable form and schemas.

dharizza:
Adding support for displaying form in different languages using a particular schema for said language, as part of this I:
- added support in the SchemaRetriever so we can get the schema for a particular language
- updated FormBuilder so that it gets the current language and retrieves the schema for it, if no schema is found, the default is used.

## QA Steps

- [ ] Go to /node/add/data the form should work as usual.
- [ ] Create a new schema for another language in schema/collections, for example dataset.es.json
- [ ] Enable language modules in your Drupal site and add one extra language, for example spanish
- [ ] Go to /en/node/add/data the form should look as usual
- [ ] Go to /es/node/add/data the form should be displaying the data as specified in the dataset.es.json file.

### DIFF
--- a/modules/json_form_widget/json_form_widget.services.yml
+++ b/modules/json_form_widget/json_form_widget.services.yml
@@ -1,7 +1,7 @@
 services:
   json_form.builder:
     class: \Drupal\json_form_widget\FormBuilder
-    arguments: ['@dkan.metastore.schema_retriever','@json_form.router', '@json_form.schema_ui_handler', '@logger.factory']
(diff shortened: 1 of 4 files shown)


dafeder:
@dharizza - are you still interested in working on this? Would you want to update the branch and get tests passing?

dafeder:
Tidying up the pull requests and I think I'm going to close this, especially as there hasn't been any movement on it in years and its a bit stale. 

I think there are a few ways we could support translating the schema/form elements but having a separate json-schema file seems problematic. Which one do you use for validation? How do we enforce that two language variants of the same schema are structurally the same and only differ in user-facing text strings? 

Will follow up with a new issue.

dafeder:
See #4426

PR #3687, opened 2021-10-05
Upgrade to latest dkan-tools

dafeder:
See GetDKAN/dkan-cli#9

* Brings the CLI tooling to php 7.4, and
* Fixes the certificate issue that has been causing all builds to fail
* Adds a `composer update` command in circleCI to avoid recurring build problems with [select2](https://www.drupal.org/project/select2)
* Fixes the xdebug commands in CircleCI for v3 and to avoid double coverage reporting
* Fixes some tests that broke after PHP upgrade (not clear why they worked before!)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,7 @@ jobs:
         type: boolean
     environment:
       TEST_RESULTS: /tmp/test-results
-      DKTL_VERSION: "4.2.3"
+      DKTL_VERSION: "4.2.4"
     steps:
       - checkout:
           path: dkan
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -97,17 +105,21 @@ jobs:
...
+                name: Run phpUnit tests
+                command: dktl dkan:test-phpunit --testsuite="DKAN Test Suite"
+      - unless:
+          condition: << parameters.upgrade >>
+          steps:
+            - run:
+                name: Run phpUnit tests
+                command: |
+                  dktl xdebug:start
+                  sed -i "/=debug/s/=.*/=coverage/" src/docker/etc/php/xdebug.ini
```
dafeder, line 121: Need to switch explicitly to "coverage" mode in xdebug3

PR #3689, opened 2021-10-07
Get datastore query maximum row limit from config

janette:
Allow site admins to set the maximum number of rows returned on a datastore query.

## QA Steps

- [ ] Visit /admin/dkan/sql-endpoint
- [ ] Set the value to something above 500
- [ ] Run a query with the limit set to the number you set in config
- [ ] Confirm the query works and is returning the expected number of rows

- [ ] Run a query with no limit specified
- [ ] Confirm it defaults to limit=500

- [ ] Run a query with the limit set to a number above the maximum value
- [ ] Confirm you get a JSON Schema validation failed error

## To Dos

- Add tests
- Move the config to live at `admin/dkan/datastore`
- Update documentation

### DIFF
--- a/modules/datastore/src/Controller/QueryController.php
+++ b/modules/datastore/src/Controller/QueryController.php
@@ -2,6 +2,7 @@
 
 namespace Drupal\datastore\Controller;
 
+use Drupal\Core\Config\ConfigFactory;
 use Drupal\common\DatasetInfo;
 use Drupal\datastore\Service\DatastoreQuery;
 use Symfony\Component\DependencyInjection\ContainerInterface;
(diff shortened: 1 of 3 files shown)


modules/datastore/src/Controller/QueryController.php
```diff
@@ -294,7 +305,8 @@ public function queryResource(string $identifier, bool $stream = FALSE) {
     }
     try {
       $payloadJson = RequestParamNormalizer::fixTypes($payloadJson, file_get_contents(__DIR__ . "/../../docs/query.json"));
-      $datastoreQuery = new DatastoreQuery($payloadJson);
+      $rows_limit = (int) $this->configFactory->get('datastore.settings')->get('rows_limit') ?: 500;
```
clayliddell, line 308: I think it would reducing confusion if we removed the default maximum of 500 specified here since it's already specified in the `$rows_limit` parameter in `Drupal\datastore\Service\DatastoreQuery::__construct()` (or vice versa, just ensure that it's occurring in one location).

clayliddell (changes requested):
Just one minor suggestion; the PR otherwise looks good to merge.

modules/datastore/src/Controller/QueryController.php
janette: sounds good

modules/datastore/src/Service/DatastoreQuery.php
```diff
@@ -14,9 +14,14 @@ class DatastoreQuery extends RootedJsonData {
    *
    * @param string $json
    *   JSON query string from API payload.
+   * @param int $rows_limit
+   *   Maxmimum rows of data to return.
    */
-  public function __construct(string $json) {
+  public function __construct(string $json, int $rows_limit = 500) {
```
clayliddell: See https://github.com/GetDKAN/dkan/pull/3689#pullrequestreview-774168989.
```suggestion
  public function __construct(string $json, int $rows_limit) {
```

dafeder:
For future reference, ab963b8097e2b1e458ad1abeb42ba5b5c16c5740 was dropped in #3686, because changing the signature requirements for a public method makes this no longer backwards compatible.

PR #3691, opened 2021-10-08
Allow management of data dictionaries in DKAN via API calls

clayliddell:
Adds a default data-dictionary schema to be used in future work to provide a native solution for column-level metadata, and for greater control over data types used in datastore imports.

- [x] Test coverage exists
- [x] Documentation exists

## QA Steps

Test that the new data-dictionary API endpoints are working by:
- [ ] Creating a new data-dictionary using the POST `/api/1/metastore/schemas/data-dictionary/items` endpoint
- [ ] Fetching a data-dictionary using the GET `/api/1/metastore/schemas/data-dictionary/items/{identifier}` endpoint
- [ ] Replacing a data-dictionary using the PUT `/api/1/metastore/schemas/data-dictionary/items/{identifier}` endpoint
- [ ] Updating a data-dictionary using the PATCH `/api/1/metastore/schemas/data-dictionary/items/{identifier}` endpoint
- [ ] Deleteing a data-dictionary using the DELETE `/api/1/metastore/schemas/data-dictionary/items/{identifier}` endpoint


clayliddell:
A part of this PR was adding complete integration testing for the Metastore API endpoints.
Unit tests are already in place for the underlying methods.

dafeder (approved):
Just a note - this is an important first step, but we'll have to think about if we want to have certain schemas be core-only, as I think the functionality we build on top of this will be less abstract-able and more coupled to the specific schema than, say, the dataset and datastore queries. At minimum, we'll need to tell people with custom schemas on their site (most of the sites we're aware of) to copy this into their schema folder, but also have any code that depends on it fail gracefully if a dictionary schema is found not to be present.

PR #3693, opened 2021-10-11
Remove max from row limit setting, combine forms

dafeder:
This PR consolidates the two datastore settings forms into one, and removes the hard limit imposed by the form on the row count.

## QA Steps

- [ ] The SQL Endpoint settings link is gone from the admin menu
- [ ] The row count limit on the datastore settings menu can be set to 20000
- [ ] If the row count field is left blank, it will be set to 500.


janette:
@dafeder  I get an error on the docs page
```
Fetch error
Bad Request /api/1/metastore/schemas/dataset/items/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/docs
```

janette:
Also am allowed to set the limit to anything on the `api/1/datastore/sql/?query=` endpoint

dafeder:
It should fail validation if the limit in the query is higher than the max setting

dafeder:
> @dafeder  I get an error on the docs page
> ```
> Fetch error
> Bad Request /api/1/metastore/schemas/dataset/items/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/docs
> ```

Ok I'll look into this.

clayliddell:
@janette I wasn't able to replicate the error you observed on my local environment with a Vanilla DKAN instance.

Also I was able to successfully complete the QA Instructions.

janette:
@dafeder @clayliddell  yes, after a fresh build everything is working as expected.

PR #3696, opened 2021-10-18
Clean up datastore controller and related classes

dafeder:
In preparation for new CSV pagination logic, organizing the datastore query controllers in a more straightforward way, to avoid bloat and code reuse. 

Code changes: 

- Introduction of an AbstractQueryController class, with the normal JSON queries and the streamed CSV queries both extending the same base class for consistency. SQL endpoint could later be brought in to follow this pattern, perhaps other datastore  variants in the future. Custom modules wanting to override or extend datastore behavior can also use this as a base.
- RequestParamNormalizer util class folded into AbstractQueryController
- Datastore routes consolidated to be easier to read
- CSV streaming code cleaned up to be more straightforward and easy to work with.
- The rows limit setting is now the query default as well as the maximum in the DatastoreQuery schema.

## QA Steps

- Set the rows limit to 1000 in admin/dkan/datastore-settings. Clear the cache.
- Confirm that https://dkan-qa-3696.ci.civicactions.net/api/1/datastore/query/schema reflects the limit setting in `['limit']` default and maximum
- Confirm that the limit has defaulted to the rows limit setting
- Other than this, the idea is this should not change anything for the user, for any exposed services or APIs. However, custom code should be checked as controllers and other classes have changed.


modules/datastore/src/Controller/AbstractQueryController.php
```diff
@@ -0,0 +1,301 @@
...
+use RootedData\RootedJsonData;
+use Drupal\Core\Config\ConfigFactoryInterface;
+use Drupal\datastore\Service;
+use Drupal\metastore\MetastoreApiResponse;
+use JsonSchema\Validator;
+use Symfony\Component\HttpFoundation\ParameterBag;
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * Class Api.
```
clayliddell:
```suggestion
 * Abstract Controller providing base functionality used to query datastores.
```

modules/datastore/src/Controller/QueryController.php
```diff
@@ -2,143 +2,17 @@
...
 use RootedData\RootedJsonData;
-use Drupal\common\Util\RequestParamNormalizer;
-use Drupal\Core\Config\ConfigFactoryInterface;
 use Ilbee\CSVResponse\CSVResponse as CsvResponse;
-use Drupal\datastore\Service;
-use Drupal\metastore\MetastoreApiResponse;
 use Symfony\Component\HttpFoundation\ParameterBag;
 
 /**
  * Class Api.
```
clayliddell:
```suggestion
 * Controller providing functionality used to perform paged datastore queries.
 *
 * This controllers streamed counterpart can be found at:
 * @see \Drupal\datastore\ControllerQueryDownloadController
```

modules/datastore/src/Controller/QueryDownloadController.php
```diff
@@ -0,0 +1,176 @@
...
+
+namespace Drupal\datastore\Controller;
+
+use Drupal\datastore\Service\DatastoreQuery;
+use Symfony\Component\HttpFoundation\StreamedResponse;
+use RootedData\RootedJsonData;
+use Symfony\Component\HttpFoundation\ParameterBag;
+
+/**
+ * Datastore CSV download methods.
```
clayliddell:
```suggestion
 * Controller providing functionality used to stream datastore queries.
 *
 * This controllers paged counterpart can be found at:
 * @see \Drupal\datastore\Controller\QueryController
```

modules/datastore/src/Controller/AbstractQueryController.php
```diff
@@ -0,0 +1,301 @@
...
+ *
+ * @package Drupal\datastore
+ */
+abstract class AbstractQueryController implements ContainerInjectionInterface {
```
clayliddell, line 23: If you have time later, I think we would benefit (in terms of extensibility) from migrating the shared functionality in this abstract controller into a "DatastoreQuery" service which could be injected into both concrete controllers.

modules/datastore/tests/src/Unit/Controller/QueryDownloadControllerTest.php
```diff
@@ -0,0 +1,233 @@
...
+
+    $csv = explode("\n", $this->buffer);
+    ob_get_clean();
+    $this->assertEquals('record_number,data', $csv[0]);
+    $this->assertEquals('1,data', $csv[1]);
+    $this->assertEquals('50,data', $csv[50]);
+    $this->assertEquals('501,data', $csv[501]);
+  }
+
+  public function testStreamedResourceQueryCsvSpecificColumns() {
```
clayliddell, line 142: I'm really confused by what case this test is covering. It seems as though it's doing the same thing as `QueryDownloadControllerTest::testStreamedQueryCsv()`, but only performing an assertion on the first row of the response.

clayliddell (commented):
@dafeder First off, good job on improve the overall organization of the datastore API query functionality. It definitely looks a lot better than it did before.
If we're able to find time at a later date, some of the methods in these controllers could definitely use some more refactoring, and additional documentation. If I were to walk away from this part of the project for a month, and then come back, it would take me some time to figure out how this all worked again.
I've made a few suggestions below. They're mostly related to documentation that I think would have helped me understand the responsibility of each of the controllers on my first read through. There's an additional suggestion for how we might re-organize the shared datastore query logic in the future (if we get a chance to re-visit this). And one question concerning a unit test I didn't quite understand.

modules/datastore/src/Controller/AbstractQueryController.php
dafeder: Agreed! I originally wanted to do this as part of this same PR, but that would be a potentially breaking change since it's a change to a service class with public methods that is used more widely. I'm trying to keep my PRs more granular. It's definitely something I want to do soon.

modules/datastore/tests/src/Unit/Controller/QueryDownloadControllerTest.php
dafeder: These tests were just moved to a new location, I didn't give each one a critical look. This is supposed to be testing that the header row of the CSV file will adapt if we pass a "properties" array, as opposed to just asking for everything. But the header row is actually the same in the end in this test, so you're right it's not testing anything really. I'll improve the test itself on another PR because it is relevant to the new streaming functionality I'm working on.

dafeder:
I tried committing your suggestions but they made codeclimate angry @clayliddell. I removed the @see lines; since the two classes are next to each other in the same dir, share the same parent class and have, IMO, pretty clear names I think we're ok without them.

modules/datastore/src/Controller/AbstractQueryController.php
clayliddell: Okay, sounds good to me!

clayliddell (approved):
Passed QA. Looks good to me!

PR #3699, opened 2021-10-19
Switch to sqlite for streaming tests

dafeder:
It's becoming increasingly difficult to test our datastore controllers with non-interactive mocks based on JSON files. This adds a simple example datastore table via sqlite into [memory](https://www.sqlite.org/inmemorydb.html) that we can perform actual queries against and get results.

The test data attempts to reproduce potential edge cases in real-world data: non-sequential row numbers, duplicate values, and a table length not divisible by an typical even row limit setting (such as 50, 500 or 10000).

## QA Steps

- No real way to QA this, just look at the test code and confirm we're still making equivilant assertions and that the test data is solid.


modules/datastore/src/Storage/SqliteDatabaseTable.php
```diff
@@ -0,0 +1,64 @@
...
+    }
+    return $schema ?? ['fields' => []];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function translateType(string $type, $info = NULL) {
+    // Clean up things like "int(10) unsigned".
+    $driver = $this->connection->driver() ?? 'mysql';
```
clayliddell: Shouldn't the default driver here be "sqlite"?
```suggestion
    $driver = $this->connection->driver() ?? 'sqlite';
```

modules/datastore/src/Storage/DatabaseTable.php
```diff
@@ -163,7 +163,7 @@ public function setSchema($schema) {
    *
    * @todo Note that this will breakZ on PostgresSQL
```
clayliddell: Just a slight pre-existing grammatical error.
```suggestion
   * @todo Note that this will break on PostgresSQL
```

modules/datastore/src/Storage/DatabaseTable.php
```diff
@@ -181,15 +181,15 @@ private function buildTableSchema($tableName, $fieldsInfo) {
...
    *   Type returned from the describe query.
-   * @param string|null $extra
-   *   MySQL "Extra" property for column.
+   * @param mixed $extra
+   *   Additional information for column.
    *
    * @return string
    *   Fritionless Table Schema compatible type.
    *
    * @see https://specs.frictionlessdata.io/table-schema
```
clayliddell: Slight pre-existing error in DocBlock return type and description.
```suggestion
   * @return array
   *   Drupal Schema array.
   *
   * @see https://api.drupal.org/api/drupal/core!lib!Drupal!Core!Database!database.api.php/group/schemaapi/9.2.x
```

modules/datastore/src/Storage/DatabaseTable.php
```diff
@@ -181,15 +181,15 @@ private function buildTableSchema($tableName, $fieldsInfo) {
...
    */
-  private function translateType(string $type, $extra = NULL) {
+  public function translateType(string $type, $extra = NULL) {
```
clayliddell: Is there intention to use this method elsewhere? If not, we should be safe to just use protected here.
```suggestion
  protected function translateType(string $type, $extra = NULL) {
```

modules/datastore/src/Storage/SqliteDatabaseTable.php
```diff
@@ -0,0 +1,64 @@
...
+      $schema['fields'][$name] = $this->translateType(strtolower($info->type), $info);
+      $schema['fields'][$name] = array_filter($schema['fields'][$name]);
+    }
+    return $schema ?? ['fields' => []];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function translateType(string $type, $info = NULL) {
```
clayliddell: See previous comment.

```suggestion
  protected function translateType(string $type, $info = NULL) {
```

modules/datastore/src/Storage/SqliteDatabaseTable.php
```diff
@@ -0,0 +1,64 @@
...
+      ->fetchAll();
+
+    $schema = $this->buildTableSchema($tableName, $fieldsInfo);
+    $this->setSchema($schema);
+  }
+
+  /**
+   * Get table schema.
+   *
+   * @todo Note that this will breakZ on PostgresSQL
```
clayliddell: I think it would be safe to remove this @todo since it's included in the parent class or just replace the DocBlock with `{@inheritdoc}`.

clayliddell (commented):
I like the idea here. I could see this helping out in other areas of testing as well. I've left a few suggestions below. Otherwise, looks good!

clayliddell (approved):
Looks good to me!

PR #3700, opened 2021-10-20
Change to keyset pagination for streaming large CSVs

### LINKED ISSUE #3646: Optimize large OFFSET queries on large datastore tables

dafeder:
Starting a cleaner PR to replace #3692. See that PR body for discussion of the performance gains from switching from OFFSET to WHERE for pagination. Briefly, initial tests of swapping the offset iterator for one that used a ">" condition based on the record_number column from the previous row of results gave us these very impressive benchmarks for streaming a 460mb CSV file locally:

|Limit|Offset|Duration|
|-|-|-|
|500|✔️|3:31:05|
|500|❌|2:04|
|20,000|✔️|6:52| 
|20,000|❌|1:18|

The iterator function in #3692 was only a few lines of code and could not accommodate many queries, particularly those with sorting. This new approach is based on ideas from [this discussion](https://stackoverflow.com/questions/56719611/generic-sql-predicate-to-use-for-keyset-pagination-on-multiple-fields) on efficient pagination with multiple ORDER BY clauses.

It introduces a new utility class called QueryIterator that adds record_number property, new sorts and conditions to iterate through multiple pages of results in the streaming response callback for downloading large CSVs. In addition to improved performance, this makes the streaming code more readable.

## Known issues

Because the query is broken up into many pages and simply iterates until no more results are returned, we cannot accept a limit in the query itself. This is not actually a regression, as the previous CSV streaming code overwrote the limit and streamed the full result set anyway. This could be corrected in a future improvement by storing the limit in a local variable before overwriting it in the DatastoreQuery object, keeping a running count of the number of rows returned, and stopping the iterator once that count matches the limit. This would require also changing the validation for the limit property in the DatastoreQuery schema, which currently has a maximum value of whatever the row limit is set to in the DKAN config.

Also, while in most cases the results returned by the QueryIterator should match those returned by the normal query endpoints exactly, there may be some differences with queries that sort by columns that do not have unique values, as the iterator will always sort by record_number in addition to any other sorts. The only way to correct this would be to append a record_number sort to all other queries as well, perhaps moving that particular function from the QueryIterator directly to the main datastore service.

### DIFF
--- a/modules/datastore/src/Controller/AbstractQueryController.php
+++ b/modules/datastore/src/Controller/AbstractQueryController.php
@@ -88,16 +88,19 @@ abstract class AbstractQueryController implements ContainerInjectionInterface {
    *   The json or CSV response.
    */
   public function query(Request $request) {
-    $payloadJson = static::getPayloadJson($request);
-
     try {
-      $datastoreQuery = new DatastoreQuery($payloadJson, $this->getRowsLimit());
+      $datastoreQuery = $this->buildDatastoreQuery($request);
     }
     catch (\Exception $e) {
       return $this->getResponseFromException($e, 400);
     }
-
-    $result = $this->datastoreService->runQuery($datastoreQuery);
+    try {
+      $result = $this->datastoreService->runQuery($datastoreQuery);
+    }
+    catch (\Exception $e) {
+      $code = (strpos($e->getMessage(), "Error retrieving") !== FALSE) ? 404 : 400;
+      return $this->getResponseFromException($e, $code);
+    }
 
     $dependencies = $this->extractMetastoreDependencies($datastoreQuery);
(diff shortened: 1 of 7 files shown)


dafeder:
Looks like we need to add a couple tests...

dafeder:
Closing for now in favor of #3703

PR #3702, opened 2021-10-25
Added conditional to prevent search_api from attempting to index an unpublished dataset.

ebenshap:
## QA Steps
At this url:
admin/config/workflow/workflows/manage/dkan_publishing
Change default moderation state to be 'Draft'

register and run a harvest. The result of this added code is that this warning message should not be printed out:
`[warning] Could not load the following items on index DKAN: ...`


jastraat:
I think this PR is out of date; it is not being using in PDC. We are using a different method to make sure draft revisions aren't being indexed by search api.

PR #3703, opened 2021-10-28
Unbuffered datastore queries for download

### LINKED ISSUE #3646: Optimize large OFFSET queries on large datastore tables

dafeder:
Another approach to #3646. Previously we'd been trying to optimize our CSV streaming by breaking up the query into multiple chunks, but adding any sorting to queries on columns other than the primary key does not scale with this. Sorting large tables will always be slow, and will need to be repeated on every iteration. This may only add a few seconds, but as tables get larger that time increases _and_ the number of times it is repeated increases.

A better solution may be to use [unbuffered queries](https://www.php.net/manual/en/mysqlinfo.concepts.buffering.php), to run the entire query just once and then stream the results directly from the database server. By default, Drupal and most other PHP/MySQL projects will always use buffered queries, meaning that the entire result set is passed to PHP immediately.

Here, we create a second database connection object based on default but adding the `PDO::MYSQL_ATTR_USE_BUFFERED_QUERY` attribute. This will be used by all datastore module database operations, which should be fine but we should look out for any unexpected side effects. 

## Known issues, misc

This adds to the pieces of DKAN that assume we are using MySQL as the underlying database. It would be good to have a fallback in case not, or to make it more obvious how one might add support for PostgreSQL or other PDO drivers. Most database backends have some equivalent to this cursor-based fetching, but it's not standardized in PDO.

## QA Steps

Coming... in general, just download some big CSVs and make sure they work!

### DIFF
--- a/modules/common/src/Storage/AbstractDatabaseTable.php
+++ b/modules/common/src/Storage/AbstractDatabaseTable.php
@@ -224,21 +224,27 @@ abstract class AbstractDatabaseTable implements DatabaseTableInterface {
    *   Query object.
    * @param string $alias
    *   (Optional) alias for primary table.
+   * @param bool $fetch
+   *   Fetch the rows if true, just return the result statement if not.
+   *
+   * @return array|\Drupal\Core\Database\StatementInterface
+   *   Array of results if $fetch is true, otherwise result of
+   *   Select::execute() (prepared Statement object or null).
    */
-  public function query(Query $query, string $alias = 't'): array {
+  public function query(Query $query, string $alias = 't', $fetch = TRUE) {
     $this->setTable();
     $query->collection = $this->getTableName();
     $selectFactory = new SelectFactory($this->connection, $alias);
     $db_query = $selectFactory->create($query);
 
     try {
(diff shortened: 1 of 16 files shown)


dafeder:
Testing notes: 

1. Tests have been copied over from the previous streaming CSV PR (#3700) and could probably be cleaned up to be more appropriate to what we're testing here.
2. We need quite a bit more test coverage for the database/query code

dafeder:
Note: This does seem to be working well, much faster than previous pagination-based approaches in all cases, but needs some more testing against real world data and queries.

grugnog (approved):

PR #3706, opened 2021-11-12
Catch case changes in file names when harvesting

janette:
## QA Steps

* copy the h1.json.txt file from this ticket (drop the .txt) to your _/src/site/files_ directory
 * run `dktl drush dkan:harvest:register '\{"identifier":"h1","extract":{"type":"\\Harvest\\ETL\\Extract\\DataJson","uri":"http://dkan.localtest.me/sites/default/files/h1.json"},"transforms":[],"load":\{"type":"\\Drupal\\harvest\\Load\\Dataset"}}'`
 * then `dktl drush dkan:harvest:run h1`
 * look at the harvested dataset in the edit form, notice the case of the filename
 * edit the h1.json file: change downloadURL to 

"http://demo.getdkan.org/sites/default/files/distribution/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/bike_lane.csv"

  * run the harvest again
  * look at the dataset in the edit form, confirm the filepath was updated

[h1.json.txt](https://github.com/GetDKAN/dkan/files/7530904/h1.json.txt)

### DIFF
--- a/modules/common/src/Storage/Query.php
+++ b/modules/common/src/Storage/Query.php
@@ -119,12 +119,14 @@ class Query implements
    *   Property to filter on.
    * @param string $value
    *   Property value to filter against.
+   * @param bool $case
+   *   Case sensitive filter?
    */
-  public function conditionByIsEqualTo(string $property, string $value) {
+  public function conditionByIsEqualTo(string $property, string $value, bool $case = FALSE) {
     $this->conditions[] = (object) [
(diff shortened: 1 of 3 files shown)


dafeder (approved):
This could be QA'ed with just a simple API call starting with one URL than patching with another, right? Not harvest-specific. I added a quick unit test.

PR #3707, opened 2021-11-13
More tests for datastore API

dafeder:
## QA Steps

Only tests changed here so just skim the new ones to make sure they make sense and are passing.


modules/datastore/tests/src/Unit/Controller/QueryControllerTest.php
```diff
@@ -57,16 +57,56 @@ public function testQueryJson() {
     $this->assertEquals(200, $result->getStatusCode());
   }
 
+    /**
+   * Do some things differently to throw exception on metastore retrieval.
+   */
```
clayliddell:
```suggestion
  /**
   * Do some things differently to throw exception on metastore retrieval.
   */
```

modules/datastore/tests/src/Unit/Controller/MockStorage.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Drupal\Tests\datastore\Unit\Controller;
+
+use Drupal\metastore\Exception\MissingObjectException;
+use Drupal\metastore\Storage\Data;
+
+/**
+ * Mock metastore controllery for certain datastore tests.
```
clayliddell:
```suggestion
 * Mock metastore controller for certain datastore tests.
```

clayliddell (approved):
Looks good to me!

I've added a couple of suggestions just to fix some comment typos, but otherwise approved.

PR #3714, opened 2021-11-25
Convert Datastore Dashboard Controller into Form

clayliddell:
- [x] Test coverage exists
- [ ] Documentation exists

## QA Steps

- [ ] Create a dataset via a harvest file.
- [ ] Ensure the DKAN instance you are working with has at least 11 datasets.
- [ ] Navigate to the datastore dashboard.
- [ ] Ensure only 10 datasets are showing up on the first page, and other datasets can be accessed via the pager.
- [ ] Ensure the UUID filter is working on the datastore dashboard.
- [ ] Ensure the harvest filter is working on the datastore dashboard.
- [ ] Ensure links for harvest IDs on the harvest dashboard navigate to a pre-filtered datastore dashboard.


modules/datastore/src/Form/DashboardForm.php
```diff
@@ -0,0 +1,440 @@
...
+      $rows = array_merge($rows, $datasetRow);
+    }
+
+    return $rows;
+  }
+
+  /**
+   * Fetch the status of all harvests.
+   *
+   * @return \Generator
```
dafeder: @clayliddell idk if upgrading codeclimate will fix this, but really this is not a very useful piece of documentation anyway - it's technically correct but doesn't really tell you what is being yielded. I'd say just remove this line.

dafeder (changes requested):
Let's override codeclimate on the buildFilters() method -- sometimes Drupal's just gonna Drupal -- but remove that `@return`

modules/datastore/src/Form/DashboardForm.php
```diff
@@ -0,0 +1,440 @@
...
+   *   Dataset harvest statuses.
```
clayliddell:
```suggestion
```

modules/datastore/src/Form/DashboardForm.php
clayliddell: Fair enough. Resolved in https://github.com/GetDKAN/dkan/pull/3714#discussion_r760346155.

modules/datastore/src/Form/DashboardForm.php
```diff
@@ -0,0 +1,437 @@
...
+
+  /**
+   * Build datasets import status table filters.
+   *
+   * @param string[] $filters
+   *   Dataset filters.
+   *
+   * @return array[]
+   *   Table filters render array.
+   */
```
clayliddell, line 180:
```suggestion
   *
   * @SuppressWarnings(PHPMD.ExcessiveMethodLength)
   */
```

dafeder: hmm neat trick. ugly but effective!

dafeder: heh nevermind, does not look like CC catches that

clayliddell: Yeah, I've tried everything I can think of. I thought this particular warning was coming from PHPMD, but I guess not...

> Let's override codeclimate on the buildFilters() method -- sometimes Drupal's just gonna Drupal...

How I can do this?

PR #3717, opened 2021-11-29
Update to PHP_CodeSniffer version 3.6 via CC beta channel

clayliddell:
I created this PR because CodeClimate's stable channel is still using PHP_CodeSniffer version 2, which doesn't have support for many features added in PHP 7+ (which is why we've been seeing quite a few false positives for non-issues surrounding these new features). If we decide to merge this PR, we'll need to do some clean up around the code base to take care of the list of new issues CodeClimate has detected.


dafeder:
@clayliddell I'm into it, let's keep this open. Can you try running phpcbf and seeing how much of this can be fixed automatically?

dafeder (approved):
Heroic effort on those docblocks @clayliddell!

PR #3718, opened 2021-11-30
Add 'hidden' state to dkan_publishing workflow

tiffneybare:
fixes WCMS-4889

Adding logic that looks at the moderation state and if the moderation state equals publish_without_search_indexing not to index the dataset in the dkan search API. 

## QA Steps

1. Edit an existing dataset or create a new dataset.
  a. Select the 'Publish without search indexing' moderation state at the bottom of the form.
  b. Click Save.
2. Search for the dataset you have 'hidden'.
  a. The dataset should not show up in the search results.
3. Navigate to the hidden dataset.
  a. It should be visible.
4. Unhide & publish the hidden dataset.
5. Search for the dataset.
  a. It should show up in the search results.


clayliddell:
@dafeder Are the upgrade tests not importing config changes? That's the only reason I can think of why they might be failing...

Update: Response from @dafeder on Slack:

> @clayliddell upgrade tests just run updatedb. If config change you need to write an update hook

modules/metastore/modules/metastore_search/metastore_search.module
```diff
@@ -1,58 +1,66 @@
...
-
-  $storage = \Drupal::service('entity_type.manager')
-    ->getStorage('search_api_index');
-
-  /* @var $index \Drupal\search_api\Entity\Index */
-  $index = $storage->load('dkan');
-
-  $index->{$method}('dkan_dataset', [$entity->uuid()]);
+function _is_dataset(EntityInterface $entity): bool {
+  return $entity->getEntityTypeId() === 'node' && $entity->bundle() === 'data' && (new Data($entity))->getDataType() === 'dataset';
```
dafeder: @clayliddell isn't the third condition here all we need? We could wrap the `new Data()` call in a try/catch.

modules/metastore/modules/metastore_search/metastore_search.module
```diff
@@ -1,58 +1,66 @@
...
-  $method = "trackItems{$operation}";
-
-  $storage = \Drupal::service('entity_type.manager')
-    ->getStorage('search_api_index');
-
-  /* @var $index \Drupal\search_api\Entity\Index */
-  $index = $storage->load('dkan');
-
-  $index->{$method}('dkan_dataset', [$entity->uuid()]);
+function _is_dataset(EntityInterface $entity): bool {
```
dafeder: I think it's best not to have a function name this generic in the global namespace... we could easily end up defining a function with the same name in any other dkan module. can we rename to `_dkan_metastore_search_is_dataset()`?

modules/metastore/src/Controller/MetastoreController.php
```diff
@@ -122,9 +122,16 @@ public function getAll(string $schema_id, Request $request) {
...
    *   The json response.
+   *
+   * @throws \InvalidArgumentException
+   *   When an unpublished or invalid resource is requested.
    */
   public function get(string $schema_id, string $identifier, Request $request) {
     try {
+      if (!$this->service->isPublished($schema_id, $identifier)) {
+        throw new \InvalidArgumentException("Error retrieving published dataset: {$schema_id} {$identifier} not found.");
+      }
```
dafeder: Not clear on why this is necessary; the [service get method](https://github.com/GetDKAN/dkan/blob/387959f740b9734fcc01808d612ba16b5591042a/modules/metastore/src/Service.php#L241) will only [retrieve published datasets](https://github.com/GetDKAN/dkan/blob/387959f740b9734fcc01808d612ba16b5591042a/modules/metastore/src/Storage/Data.php#L187). 

If it is necessary, I don't think `InvalidArgumentException` is appropriate here. The arguments are valid. You could use `Drupal\metastore\Exception\MissingObjectException`.

dafeder (changes requested):
Couple questions in the comments -- looks very good but I'd like to check in and make sure I understand the plugin/processor architecture here and why it was necessary.

modules/metastore/modules/metastore_search/metastore_search.module
clayliddell: Yeah, I agree. It's a bit less readable, but it at least guarantees consolidation of validation logic.

modules/metastore/modules/metastore_search/metastore_search.module
clayliddell: Fair enough.

modules/metastore/src/Controller/MetastoreController.php
clayliddell: I think you're right. This was an early change that was made which can be removed due to later refactoring.

dafeder (approved):

modules/metastore/config/install/workflows.workflow.dkan_publishing.yml
```diff
@@ -28,6 +28,11 @@ type_settings:
       default_revision: true
       label: Orphaned
       weight: 2
+    hidden:
+      published: true
+      default_revision: true
+      label: 'Publish without search indexing'
```
dafeder:
```suggestion
      label: 'Published (hidden)'
```
More consistent state labels

modules/metastore/metastore.install
```diff
@@ -60,3 +60,49 @@ function metastore_update_8004() {
...
+
+/**
+ * Add hidden state and transitions to dkan publishing workflow config.
+ */
+function metastore_update_8005() {
+  $workflow_settings = \Drupal::service('config.factory')->getEditable('workflows.workflow.dkan_publishing');
+
+  $states = $workflow_settings->get('type_settings.states');
+  $states['hidden'] = [
+    'label' => 'Publish without search indexing',
```
dafeder:
```suggestion
    'label' => 'Published (hidden)',
```
More consistent state labels

dafeder (changes requested):
Sorry for a second round, this came up after more careful QA. I think the naming needs to be more consistent with the other states, and I think "search indexing" is a confusing term for a non-Drupal-expert. Any name is still potentially confusing though and we should figure out a way to inject more contextual help in the form later on.

dafeder:
Applied suggestions after 👍🏽 from @clayliddell , will merge if tests still pass.

dafeder (approved):

PR #3719, opened 2021-11-30
Allow newline chars in CSV headers in MySQL Import service

clayliddell:
fixes WCMS-7337

- [x] Test coverage exists
- [x] Documentation exists

## QA Steps

- [ ] Create a dataset with a CSV file containing newlines in it's headers.
- [ ] Ensure the dataset imports correctly.


modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
```diff
@@ -86,38 +97,32 @@ protected function runIt() {
       return $this->setResultError(sprintf('Unable to resolve file name "%s" for resource with identifier "%s" and version "%s".', $file_path, $this->resource->getIdentifier(), $this->resource->getVersion()));
     }
 
-    // Read the header line from the CSV file.
+    // Read the columns and EOL character sequence from the CSV file.
     try {
-      $header_line = $this->getFirstLineFromFile($file_path);
+      ['columns' => $columns, 'lines' => $column_lines] = $this->getColsFromFile($file_path);
```
dafeder:
```suggestion
      [$columns, $column_lines] = $this->getColsFromFile($file_path);
```
Can we simplify this?

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
```diff
@@ -153,16 +159,22 @@ protected function getFirstLineFromFile(string $file_path): string {
...
-    // Ensure the first line of the resource file was successfully read.
-    if (!isset($header_line) || $header_line === FALSE) {
-      throw new FileException(sprintf('Failed to read header line from resource file "%s".', $file_path));
+    // Ensure the columns of the resource file were successfully read.
+    if (!isset($columns) || $columns === FALSE) {
+      throw new FileException(sprintf('Failed to read columns from resource file "%s".', $file_path));
     }
 
-    return $header_line;
+    return ['columns' => $columns, 'lines' => $column_lines];
```
dafeder:
```suggestion
    return [$columns, $column_lines];
```
Can we simplify this? Maybe this is a common construction but I think it would be a bit clearer, if we have to return a list, to just use a simple array,. that's what i'm more used to seeing. (Not going to die on this hill though if you feel strongly this is better)

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
```diff
@@ -86,38 +97,32 @@ protected function runIt() {
...
-
     // Call `count` on database table in order to ensure a database table has
     // been created for the datastore.
     // @todo Find a better way to ensure creation of datastore tables.
     $this->dataStorage->count();
     // Construct and execute a SQL import statement using the information
     // gathered from the CSV file being imported.
     $this->getDatabaseConnectionCapableOfDataLoad()->query(
-      $this->getSqlStatement($file_path, $this->dataStorage->getTableName(), $headers, $eol));
+      $this->getSqlStatement($file_path, $this->dataStorage->getTableName(), $headers, $eol, $header_line_count));
```
dafeder, line 125:
```suggestion
      $this->getSqlStatement($file_path, $headers, $eol, $header_line_count));
```
CodeClimate's complaining about too many arguments, and this can be done inside the method.

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
```diff
@@ -299,19 +311,21 @@ protected function truncateHeader(string $column): string {
...
    *   End Of Line character for file importation.
+   * @param int $header_line_count
+   *   Number of lines occupied by the csv header row.
    *
    * @return string
    *   Generated SQL file import statement.
    */
-  private function getSqlStatement(string $file_path, string $tablename, array $headers, string $eol): string {
+  protected function getSqlStatement(string $file_path, string $tablename, array $headers, string $eol, int $header_line_count): string {
     return implode(' ', [
```
dafeder, line 321:
```suggestion
  protected function getSqlStatement(string $file_path, array $headers, string $eol, int $header_line_count): string {
    $tablename = $this->dataStorage->getTableName();
    return implode(' ', [
```
CodeClimate is complaining about number of args, and there's no reason we can't do this here.

dafeder (changes requested):
Couple suggestions on code standards

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
clayliddell: I can see the benefit of it being either way. I prefer the verbosity, but if it's bothering you, I don't mind to change it.

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
clayliddell: I would prefer not to do this, since it would make the `getSqlStatement()` method no longer be deterministic (pure/static), and would further complicate the already convoluted MySQL service mock in `MysqlImportTest::getMysqlImporter()`.

I understand the idea behind CodeClimate's 4 parameter limit, but I think string builder functions (like this one) and constructor methods are really an exception.

If this is really bothering you, I'd be open to possibly using a parameter object to clean up the method signature a bit.

dafeder (approved):

PR #3724, opened 2021-12-09
Tweak the admin links and help text.

janette:
Help text updated for the following fields:
- title
- description
- distribution
- dist. title
- dist. description
- dist. format
- dist. download url
- dist. access url
- dist. data standard
    
Titles updated for the following fields:
- dist. title => File Title
- dist. desc -> File Description
- dist. format => File Format

menu


dafeder (approved):
I think there are more improvements to be made on the terminology here, "file" vs "resource" vs "distribution", but this is definitely an improvement!

PR #3728, opened 2021-12-17
Remove unavailable class typehints from project

clayliddell:
- [X] Test coverage exists

## QA Steps

- [ ] Ensure the CircleCI pipelines are passing without warnings.


dafeder (approved):
This feels a little strange but I can't think of a scenario where this actually causes a problem, best way to silence that error for now!

Issue #3721, opened 2021-12-02
File usage on uploaded files is not captured correctly


janette:
Currently the [file usage](https://github.com/GetDKAN/dkan/blob/2.x/modules/json_form_widget/json_form_widget.module#L74) is tied to a non-existent entity type (json_form_widget) and the users id.

1. This causes an error when clicking on the usage link in /admin/content/files - See error below.

2. When a dataset is deleted, it's distributions are set to unpublished. If an administrator deletes these distribution nodes, the usage is not updated to 0, the original usage remains, making it impossible to delete the files properly from the server.

To Dos:
1. Uploaded file usage should be assigned to the entity type "node", and distribution node ID rather than "json_form_widget" and user ID.
2. To clean up existing sites, we would need a hook to fix the existing usage records.
3. There should also be a config setting to automate deletion of files when the usage is at 0.

```
Drupal\Component\Plugin\Exception\PluginNotFoundException: The "json_form_widget" entity type does not exist. in Drupal\Core\Entity\EntityTypeManager->getDefinition() (line 143 of core/lib/Drupal/Core/Entity/EntityTypeManager.php).

Drupal\Core\Entity\EntityTypeManager->getHandler('json_form_widget', 'storage') (Line: 196)
Drupal\Core\Entity\EntityTypeManager->getStorage('json_form_widget') (Line: 139)
Drupal\views\Plugin\views\field\EntityLabel->preRender(Array) (Line: 1503)
Drupal\views\ViewExecutable->render() (Line: 183)
Drupal\views\Plugin\views\display\Page->execute() (Line: 1630)
Drupal\views\ViewExecutable->executeDisplay('page_2', Array) (Line: 81)
Drupal\views\Element\View::preRenderViewElement(Array)
call_user_func_array(Array, Array) (Line: 101)
Drupal\Core\Render\Renderer->doTrustedCallback(Array, Array, 'Render #pre_render callbacks must be methods of a class that implements \Drupal\Core\Security\TrustedCallbackInterface or be an anonymous function. The callback was %s. See https://www.drupal.org/node/2966725', 'exception', 'Drupal\Core\Render\Element\RenderCallbackInterface') (Line: 786)
Drupal\Core\Render\Renderer->doCallback('#pre_render', Array, Array) (Line: 377)
Drupal\Core\Render\Renderer->doRender(Array, ) (Line: 201)
Drupal\Core\Render\Renderer->render(Array, ) (Line: 241)
[...]

PR #3730, opened 2021-12-22
New file usage logic for json_form_widget

dafeder:
File usage for uploaded files on the form was broken. This PR:

* Sets correct entity name and ID
* Removes DKAN-specific logic (hope is to spin this module off some day)
* Adds a hook_entity_delete to clean up file usage if the associated node is deleted

## QA Steps

- [ ] Create a new dataset via the UI and upload a CSV file
- [ ] Check that the new file appears as expected in admin/content/files, used in 1 place
- [ ] Click on "1 place" and confirm that the correct dataset node is referenced
- [ ] Delete the new dataset
- [ ] Clear cache
- [ ] Confirm that admin/content/files now lists "0 places" for the file.

### Delete hook

The new delete hook could be expensive in certain scenarios -- either sites with a lot of content types and functionality beyond DKAN, or very large catalogs. If we see issues in any implementations or from the community we could add a configuration setting to make the delete behavior optional.

### Orphans

Most of the point of this was to make orphaned files go away after deleting a dataset.

If this new behavior is used in combination with setting `file.settings:make_unused_managed_files_temporary` to `true`, files will be set to "temporary" status as well once their parent dataset is deleted. This will cause them to be deleted after 6 hours (or whatever you have set in `system.file:temporary_maximum_age`. 

But, maybe we shouldn't do or recommend this for now:

* This setting is [considered dangerous](https://www.drupal.org/node/2891902). If DKAN is being used, for instance, on a site that also has a media gallery, it would definitely not be advisable to enable this setting. 
* This file usage functionality is new and could have undiscovered bugs that prevent proper file usage from being recorded; if usage were accidentally set to 0 but a file were still in use it could be deleted unexpectedly.
* For now, file entities are only created for files uploaded through the form. So we can imagine a scenario where a file is uploaded, and then a second dataset created that points to the same file via URL. That will not get recorded as a usage, so if we delete the first dataset, the file will be orphaned and deleted even though it's still in use by the second dataset. This is obviously an edge case but it does give me some pause.

I'd recommend we add this for now and keep our eye on the file usage. We may want to install [file_delete](https://www.drupal.org/project/file_delete) so that we can manually delete files once we see they are unused. In the future we can experiment with the `make_unused_managed_files_temporary` setting or even adding our own logic to the module that sets the unused json_form_widget files to temporary automatically.

### Tests

My understanding is that we still do not have a reliable way to test file uploads with cypress. Also all the code here is in the .module file; I don't actually know how to test this properly.

### DIFF
--- a/modules/json_form_widget/json_form_widget.module
+++ b/modules/json_form_widget/json_form_widget.module
@@ -2,9 +2,16 @@
 
 /**
  * @file
+ * Defines a multi-field form element based on a JSON Schema.
  */
 
+use Drupal\Core\Entity\EntityFormInterface;
+use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\file\Entity\File;
+use Drupal\file\FileInterface;
+use Drupal\json_form_widget\Plugin\Field\FieldWidget\JsonFormWidget;
+
 
 /**
  * Submit handler for the "add-one-more" button.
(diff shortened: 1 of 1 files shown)


janette (approved):
Nice

stefan-korn:
Questions about this change: 

Is it supposed to solve #3721 too?

The file usage is tied to the dataset entity, while the file is actually tied to a distribution. Deleting a dataset, does not delete the related distributions. Are the distributions with deleting the dataset to be considered obsolete or orphaned and therefore it is okay to remove the file usage although the distribution is still exisiting?

dafeder:
@stefan-korn yes, a distribution with no datasets referencing it is considered to be orphaned and unneeded. I agree there is an argument for tying file usage to the distribution not the dataset, but the usage is really more just a byproduct of using the form and not used anywhere in DKAN's business logic. Since it's actually the dataset form being used it gets tied to that. Because files referenced using the API or a harvest are not even given IDs or usage stats, we have mostly left this alone.

stefan-korn:
@dafeder : Thanks for the explanation. I am only a bit wary of file usage, because I have had situations (not DKAN in particular, but Drupal in general) where file usage was preventing expected file deletion through entity deletion or vice versa. And this can lead to an "outcry" with pointing fingers and the Drupal users often do not get the difference and implications between file and entity.

PR #3731, opened 2021-12-22
Suport GROUP BY clause in datastore queries

dafeder:
Adds support for SQL `GROUP BY ` clauses in datastore queries via new `groups` property. For instance, for a table of objects with a column called `color`, we could ask for the count of every color with this query:

```json
{
    "properties": [
        "color",
        {
            "alias": "count",
            "expression": {
                "operator": "count",
                "operands": [
                    "record_number"
                ]
            }
        }
    ],
    "groupings": ["color"]
}
```

## QA Steps

1. Create a dataset using the following csv file: https://gist.githubusercontent.com/clayliddell/7a7ead939be578d5303ffcfc60247e12/raw/bd2768564bc3a0a592f25312bc606beceb10c7d1/colors.csv.
2. Perform a request against the /api/1/datastore/query/{datasetId}/{index} endpoint with the created dataset ID and the above JSON payload.
3. Ensure the API response contains the following result:
```json
{
  "results": [
    {
      "color": "blue",
      "count": "2"
    },
    {
      "color": "green",
      "count": "3"
    },
    {
      "color": "red",
      "count": "1"
    },
    {
      "color": "yellow",
      "count": "1"
    }
  ]
}
```

### DIFF
--- a/modules/datastore/src/Storage/QueryFactory.php
+++ b/modules/datastore/src/Storage/QueryFactory.php
@@ -92,7 +92,6 @@ class QueryFactory {
     }
     $props = $this->extractPropertyNames($this->datastoreQuery->{"$.properties"} ?? []);
     if ($ungrouped = array_diff($props, $groups)) {
-      \Drupal::logger('dkan')->notice('<pre>' . print_r($ungrouped, TRUE) . '</pre>');
       throw new \Exception('Un-grouped properties found in aggregate query: ' . implode(', ', $ungrouped));
     }
     $query->groupby = $groups;
--- a/modules/datastore/src/Storage/QueryFactory.php
+++ b/modules/datastore/src/Storage/QueryFactory.php
@@ -86,11 +86,11 @@ class QueryFactory {
    *   When ungrouped properties are found in the datastore query.
    */
   private function populateQueryGroupBy(Query $query): void {
-    $groups = $this->extractPropertyNames($this->datastoreQuery->{"$.groups"});
+    $groups = $this->extractPropertyNames($this->datastoreQuery->{"$.groups"} ?? []);
     if (empty($groups)) {
       return;
     }
-    $props = $this->extractPropertyNames($this->datastoreQuery->{"$.properties"});
+    $props = $this->extractPropertyNames($this->datastoreQuery->{"$.properties"} ?? []);
     if ($ungrouped = array_diff($props, $groups)) {
       \Drupal::logger('dkan')->notice('<pre>' . print_r($ungrouped, TRUE) . '</pre>');
       throw new \Exception('Un-grouped properties found in aggregate query: ' . implode(', ', $ungrouped));
--- a/modules/datastore/src/Storage/QueryFactory.php
+++ b/modules/datastore/src/Storage/QueryFactory.php
@@ -77,7 +77,7 @@ class QueryFactory {
   }
 
   /**
-   * Helper function for adding datastore groups to the given query.
+   * Helper function for adding group by clauses to the given query.
    *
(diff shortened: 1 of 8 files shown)


modules/datastore/src/Storage/QueryFactory.php
```diff
@@ -75,6 +76,68 @@ public function populateQuery(): Query {
...
+      if (is_array($prop)) {
+        $prop = (object) $prop;
+      }
+      if (is_object($prop)) {
+        if (isset($prop->property)) {
+          return $prop->property;
+        }
+        // If this property object is an expression, ignore it.
+        if (isset($prop->expression)) {
+          return NULL;
```
dafeder, line 133: In this case, wouldn't we want to return the expression's alias, not `null`?

modules/datastore/src/Storage/QueryFactory.php
```diff
@@ -75,6 +76,68 @@ public function populateQuery(): Query {
...
+        }
+        throw new \Exception('Invalid property object is neither a property or expression.');
+      }
+      throw new \Exception('Invalid type for resource property, expected string or object, found "' . gettype($prop) . '".');
```
dafeder: I don't know that these exceptions are necessary, but if you think we need them I think it would be best practice to provide test coverage.

modules/datastore/tests/data/query/groupByQuery.json
```diff
@@ -0,0 +1,40 @@
...
+    ],
+    "conditions": [
+        {
+            "resource": "t",
+            "property": "filterable",
+            "value": "value",
+            "operator": "="
+        }
+    ],
+    "groups": [
```
dafeder: @clayliddell the more I think about it the more I think, semantically, "groupings" is the right word to use here, not "groups." Would you be open to changing this?

dafeder (commented):
I think you may need to branch off this and push as a new PR for me to actually approve and merge this!

modules/datastore/src/Storage/QueryFactory.php
dafeder: Actually, looks like we fail codeclimate coverage check unless we add that coverage.

dafeder:
Pushing some new commits to address the comments above - I don't believe the two exceptions are necessary as we validate the properties array elsewhere.

modules/datastore/src/Storage/QueryFactory.php
clayliddell: I have mixed feelings about this. I think it's probably ok to remove the exceptions for now, but the `QueryFactory` service should really be refactored such that it's not tightly coupled to schema validation. That way these exceptions could be reached and tested.

modules/datastore/src/Storage/QueryFactory.php
clayliddell: The name of this method is a bit confusing. I'm trying to extract "property properties" not "expression properties" :sweat_smile:

modules/datastore/tests/data/query/groupByQuery.json
clayliddell: Yeah, I like that better!

clayliddell (approved):

modules/datastore/src/Storage/QueryFactory.php
dafeder: @clayliddell currently these conditions should still cause an exception elsewhere in the system so there's no plausible way it would get this far, making this difficult to test plausibly. Also I am hoping we can remove QueryFactory completely on the next round of refactoring so I don't think it's worth investing more time in developing tests for this.

PR #3733, opened 2021-12-23
Consolidate metastore retrieve/range methods

dafeder:
Fixes multiple inconsistencies and limitations in the metastore service retrieve methods. Makes it easier to use the metastore service and underlying storage classes to retrieve lists of metadata items or identifiers and either include or exclude unpublished items.

## QA Steps

1. Install demo
2. Install and enable simple_sitemap
3. Run cron. Confirm you see all datasets in /sitemap.xml, including `cedcd327-4e5d-43f9-8eb1-c11850fa7c55`
4. Unpublish the "Florida Bike Lanes" dataset
5. Run cron (or regenerate the sitemap in the admin UI)
6. Confirm `cedcd327-4e5d-43f9-8eb1-c11850fa7c55` is now excluded from /sitemap.xml
7. Confirm `cedcd327-4e5d-43f9-8eb1-c11850fa7c55` is still listed in the Datastore Import Status page as "unpublished"

## Note

* This PR changes several public methods around retrieval, and also changes the signature for the MetastoreStorageInterface::publish method (returns bool instead of string). 
* For this reason, review carefully and make sure you agree with the change, and be sure to make the next release after merge a point, bumping us to (as of this writing) 2.13.

### DIFF
--- a/modules/metastore/src/Service.php
+++ b/modules/metastore/src/Service.php
@@ -118,12 +118,14 @@ class Service implements ContainerInjectionInterface {
    *
    * @param string $schema_id
    *   The schema ID to be counted.
+   * @param bool $unpublished
+   *   Whether to include unpublished items.
    *
    * @return int
    *   Object count.
    */
-  public function count(string $schema_id): int {
-    return $this->getStorage($schema_id)->count();
+  public function count(string $schema_id, bool $unpublished = FALSE): int {
+    return $this->getStorage($schema_id)->count($unpublished);
   }
 
   /**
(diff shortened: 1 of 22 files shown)


modules/metastore/src/Storage/Data.php
```diff
@@ -61,6 +67,20 @@ abstract class Data implements MetastoreStorageInterface {
...
 
+  /**
+   * The entity field or property used to store JSON metadata.
+   *
+   * @var string
+   */
+  protected $metadataField;
+
+  /**
+   * The entity field or property used to store the schema ID (e.g. "dataset).
```
clayliddell:
```suggestion
   * The entity field or property used to store the schema ID (e.g. "dataset").
```

modules/metastore/src/Service.php
```diff
@@ -106,7 +106,7 @@ public function getSchema($identifier) {
    * @return \Drupal\metastore\Storage\MetastoreStorageInterface
    *   Entity storage.
    */
-  private function getStorage(string $schema_id): MetastoreStorageInterface {
+  public function getStorage(string $schema_id): MetastoreStorageInterface {
```
clayliddell: I don't think we need to change the access level of this method. It seems like we could just call `DataFactory::getInstance()` directly in the tests for the NodeData class.

modules/metastore/src/Service.php
```diff
@@ -137,54 +139,33 @@ public function count(string $schema_id): int {
...
    *   Number of items to retrieve.
+   * @param bool $unpublished
+   *   Whether to include unpublished items. Should default to false.
    *
    * @return array
    *   Array of RootedJsonData objects.
    */
-  public function getRange(string $schema_id, int $start, int $length):array {
-    $jsonStringsArray = $this->getStorage($schema_id)->retrieveRange($start, $length);
+  public function getAll(string $schema_id, ?int $start = NULL, ?int $length = NULL, $unpublished = FALSE):array {
```
clayliddell, line 167: Slight nitpick: `getAll()` sounds weird to me since this no longer gets all of the metastore items for the given schema. How does `getMultiple()` sound?

clayliddell (commented):
Awesome! The metastore service is looking a lot cleaner with these revisions. I left a couple comments below for discussion.

modules/metastore/src/Storage/MetastoreEntityStorageInterface.php
```diff
@@ -7,28 +7,4 @@
  */
 interface MetastoreEntityStorageInterface extends MetastoreStorageInterface {
```
clayliddell, line 8: Can this file be deleted now? It doesn't look like the interface is being used anywhere else.

dafeder: Maybe. Probably. This was one of several things I considered doing in this PR but decided to hold back on to minimize the disruption in one PR. I think when we're all back on this full time we should have a less rushed conversation about further refactor/consolidation of the classes and interfaces in the metastore module, but I'd like to hold off on deleting this completely until we can do that.

modules/metastore/src/Service.php
dafeder: You are correct. Changing.

modules/metastore/src/Service.php
dafeder: I had the same thought, but ultimately decided to punt on it for similar reasons I decided against removing that interface. I'd like to think more carefully and in a group about the best naming for all these methods, and also I think public methods on a service should be the ones we want to be the most cautious about renaming or removing.

Think about it this way: we are still getting "all" the datasets. We may want to paginate the results in certain contexts for performance/scaling reasons, and "all" is ambiguous because it could mean all the public datasets or all the datasets including unpublished, but there is no further filtering going on here.

I'd propose we keep it getAll at least for this week, and have that conversation or conversations about further metastore refactoring next week and just jump to 2.14 if we decide renaming getAll is still the best course of action.

clayliddell: Ok, fair enough. This would be considered a BC-breaking change if someone were making use of the metastore service in some custom code. I agree that we need to have a larger conversation around refactoring efforts.

clayliddell (approved):

PR #3737, opened 2022-01-13
Bulk unpublish/archive harvested datasets

dafeder:
Adds a new drush command, `dkan:harvest:archive`. Provided a harvest ID, it will archive all datasets connected to that harvest.

Note that connected items, like distributions, will not be touched. A separate PR will sync distributions workflow state to their parent datasets.

There is also a `dkan:harvest:publish` command that will do the inverse.

## QA Steps

This harvest from demo.getdkan.com, so don't run these exact steps if you have the QA content.

- `drush dkan:harvest:register '{"identifier":"example","extract":{"type":"\\Harvest\\ETL\\Extract\\DataJson","uri":"https://demo.getdkan.org/data.json"},"transforms":[],"load":{"type":"\\Drupal\\harvest\\Load\\Dataset"}}'``
- `drush dkan:harvest:run example`
- Confirm that you have some published datasets.
- `drush dkan:harvest:archive example`
- Confirm that datasets are now unpublished.
- `drush dkan:harvest:publish example`
- Confirm that datasets are now published.

## Note

I don't think there's a way to keep the method count under 20 on that one file CodeClimate is complaining about. I will be trying to break a lot of this stuff up in the coming weeks to refactor the metastore into some more manageable classes; lets revisit after that?


dafeder:
Added a publish command. I think the redundant code complaint from codeclimate is ok for now; command classes are kind of their own thing (should clean up at some point through).

janette (approved):

janette:
Keep this branch until the next tag is cut, **2.13.2**

PR #3739, opened 2022-01-18
Doc cleanup

dafeder:
Was originally going to add documentation for the hidden state (#3718) but realized there is not really a good place to add it and docs need a general workflow section (along with a lot of other stuff!). But, noticed some places the docs could be cleaned up:

1. The readme files of two newer modules were accidentally being picked up and placed in the TOC root. We may want to figure out a way to incorporate these later but for now this seems out of place, so these were hidden.
2. The version number was stuck at 2.0.0. I replaced this in the doxyfile to be a variable called `DKAN_VERSION`.

To generate the docs with a arbitrary version number, you could run..

```sh
DKAN_VERSION=`v2.1.0` doxygen
```
from the DKAN root dir. To supply the current branch, you could use:

```sh
DKAN_VERSION=`git rev-parse --abbrev-ref HEAD` doxygen
```

The demo site script has already recieved a PR to use this: GetDKAN/jenkins_release_n_update_builds#2

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -38,7 +38,7 @@ PROJECT_NAME           = "DKAN"
 # could be handy for archiving the generated documentation or if some version
 # control system is used.
 
-PROJECT_NUMBER         = v2.0.0
+PROJECT_NUMBER         = $(DKAN_VERSION)
 
 # Using the PROJECT_BRIEF tag one can provide an optional one line description
 # for a project that appears at the top of each page and should give viewer a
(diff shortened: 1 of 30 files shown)


clayliddell:
@dafeder Could we create a follow up ticket or add a note on an existing ticket in Jira to incorporate the new modules (`dkan_js_frontend` and `datastore_mysql_import`)?

Otherwise this looks good.

dafeder:
@janette can you take a look at this?

janette:
The arbitrary version did not work for me. 
The screen shot is from using: 
```DKAN_VERSION=`git rev-parse --abbrev-ref HEAD` doxygen```

Is having some modules in the TOC root what we were expecting?
Or are they supposed to show up under 'Modules'
docs
?

janette (approved):

PR #3742, opened 2022-01-21
Add 'hidden' state to dkan_publishing workflow (round 2)

clayliddell:
fixes WCMS-4889

The original PR (#3718) for this was reverted. This PR fixes some issues found related to unpublished nodes being visible from the frontend. 

- [X] Test coverage exists
- [X] Documentation exists

### Hidden State
1. Edit an existing dataset or create a new dataset.
  a. Select the "Publish (Hidden)" moderation state at the bottom of the form.
  b. Click Save.
2. Search for the hidden dataset.
  a. The dataset should not show up in the search results.
3. Navigate to the hidden dataset.
  a. It should be visible.
4. Unhide & publish the hidden dataset.
5. Search for the dataset.
  a. It should show up in the search results.

###  Archived State
1. Edit an existing dataset or create a new dataset.
  a. Select the "Archive" moderation state at the bottom of the form.
  b. Click Save.
2. Search for the archived dataset.
  a. The dataset should not show up in the search results.
3. Navigate to the archived dataset.
  a. It should not be visible.
4. Unarchive & publish the archived dataset.
5. Search for the dataset.
  a. It should show up in the search results.


dafeder (approved):

cypress/integration/10_workflow_transitions.spec.js
```diff
@@ -0,0 +1,163 @@
...
+
+      // Ensure dataset is hidden from catalog
+      dkan.searchMetastore({fulltext: dataset_title, facets: ''}).then((response) => {
+        expect(response.status).to.eq(200)
+        expect(response.body.results).to.be.empty
+      })
+
+      // Ensure dataset details are not available when directly visited
+      cy.request({
+        url: '/api/1/metastore/schemas/dataset/items/' + datasetId,
```
dgading, line 20: Using `?show-reference-ids` shouldn't change how the metastore calls work and possibly show something, correct? I use this param in the frontend on most of my calls.

dafeder: No, that parameter is checked and acted on after the item has already been retrieved from the metastore; it's outside of the logic being tested here.

dafeder: See https://github.com/GetDKAN/dkan/blob/2.x/modules/metastore/src/Controller/MetastoreController.php#L129 - we would have already thrown an exception if the get() failed

PR #3743, opened 2022-01-21
Harvest drush options

dafeder:
Adds some arguments to the `drush dkan:harvest:register` command so you don't have to use JSON.

JSON argument still works. But now you can create a standard Harvest plan by just passing `--identifier` and `--extract-uri`. All JSON properties do have an equivalent CLI option though that can be passed.

Also adds a JSON schema for harvest plans, but it's not being used for anything yet.


janette (changes requested):
@dafeder can you make the --extract-uri required as well. And maybe the schema should include the titles and required properties to match the one in the harvest repo https://github.com/GetDKAN/harvest/blob/master/schema/schema.json

dafeder:
> And maybe the schema should include the titles and required properties to match the one in the harvest repo 

Hah I totally missed that this existed. My schema is redundant, I'll just remove it.

modules/harvest/src/Commands/HarvestCommands.php
```diff
@@ -64,20 +66,37 @@ function ($id) {
...
+   *
+   * @option identifier Identifier
+   * @option extract-type Extract type
+   * @option extract-uri Extract URI
+   * @option transform A transform class to apply. You may pass multiple transforms.
+   * @option load-type Load class
    *
    * @command dkan:harvest:register
+   *
    * @usage dkan-harvest:register '{"identifier":"example","extract":{"type":"\\Harvest\\ETL\\Extract\\DataJson","uri":"https://source/data.json"},"transforms":[],"load":{"type":"\\Drupal\\harvest\\Load\\Dataset"}}'
```
janette: update to `@usage dkan:harvest:register`

dafeder:
@janette see updates

janette (approved):

PR #3745, opened 2022-01-29
Prepend paths in api docs, based on $settings['dkan_api_base']

thierrydallacroce:
Based on a user-defined `$settings['dkan_api_base']` value (i.e. `/provider-data`) in a `settings*.php` file, prepend that value to the API paths documentation.

- [x] Test coverage exists

### DIFF
--- a/modules/common/src/DkanApiDocsGenerator.php
+++ b/modules/common/src/DkanApiDocsGenerator.php
@@ -59,8 +59,9 @@ class DkanApiDocsGenerator {
       $spec = array_merge_recursive($spec, $pluginSpec);
     }
 
-    // Add 'dkan_api_base' setting to prefix the API path if your routing does not start
-    // with your site's root URL (e.g. "data" for API paths to use "/data/api/1").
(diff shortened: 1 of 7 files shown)


thierrydallacroce:
@dafeder, this should be good to go. Let me know if you have any concerns or questions.
Thanks in advance.

modules/common/src/DkanApiDocsGenerator.php
```diff
@@ -40,6 +59,10 @@ public function buildSpec(array $plugins = []) {
       $spec = array_merge_recursive($spec, $pluginSpec);
     }
```
dafeder, line 61:
```suggestion

    // Add 'dkan_api_base' setting to prefix the API path if your routing does not start
    // with your site's root URL (e.g. "data" for API paths to use "/data/api/1").
```

dafeder: @thierrydallacroce let's add a bit of explanation?

dafeder (changes requested):
Looks great just one inline documentation request

dafeder (approved):

PR #3747, opened 2022-02-01
Metastore API integration tests

dafeder:
New PHPUnit-based integration tests for API based on network requests. They live under Functional for compatibility with [Drupal test traits](https://gitlab.com/weitzman/drupal-test-traits) but are more properly understood conceptually as a combination of integration tests and [contract tests](https://pactflow.io/blog/what-is-contract-testing/).

Hopefully this pattern can continue to be implemented in the rest of the API, and we can share as much code as possible in base classes or even automate aspects of test creation through OpenAPI documentation.


dafeder:
This was included in #3748

PR #3748, opened 2022-02-03
Add revisions API for changing moderation state

dafeder:
Adds a new endpoint to the metastore API. After any /api/1/metastore/schemas/{schema-id}/items/{item-id}, we can add /revisions and GET a list of basic metadata on each revision of the metastore item. We can also POST a new revision with a moderation state and a log message, which will have the effect of moving the item (usually a dataset) to that new moderation state. Finally, to correct follow the REST pattern we have /revisions/{revision-id} which will just return the single revision object.

## QA Steps

- Start with a site that includes the sample content.
- Request `GET /api/1/metastore/schemas/dataset/items/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/revisions`
- Confirm there is at least one revision, and the first is published "true" and state "published"
- Request:
```http
POST /api/1/metastore/schemas/dataset/items/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/revisions
Authorization: Basic user:password

{
    "state": "hidden",
    "message": "Testing state change"
}
```
- Confirm success. Request the GET endpoint again, confirm current state is now "hidden".
- Confirm dataset behaves properly as a hidden dataset.
- Repeat test with "archived" and then back to "published"

## Known issues

- Our documentation is much too bloated as we currently render a separate path for every schema. We need to trim it down, in my opinion, before adding three more items to that section of the docs. A future PR will clean this up and add the revisions information.
- This PR introduces a new pattern for API tests using guzzle to test the endpoints directly, rather than calling PHP methods. This pattern will hopefully evolve and introduce more automation as it covers more of the API. The goal would be to test every possible iteration of requests and responses documented in OpenAPI spec with a minimum of code. Unfortunately, for now it is not integrating with the code coverage reporter in PHPUnit/Xdebug. So, for now I am excluding the new revisions controller from coverage.


clayliddell (approved):
Code lgtm & passed QA.

PR #3751, opened 2022-02-16
Simplify metastore docs

dafeder:
In the metastore openapi spec, we've been caught between two problems:

1. Treating the get/post/patch/put endpoints for specific schemas as separate paths results in a very long spec. And now with revisions we have three additional paths, which would get replicated with each schema.
2. Treating each verb as a single path with a {schemaId} parameter makes the documentation confusing -- most people are looking for information on how to work with datasets specifically. We can't specify request schemas in the spec when the schema itself itself is dynamic.

This PR attempts a hybrid approach:

1. Revision paths are added once, with parameterized schemaID, as the actual request schema does not changes between datasets, distributions etc.
2. We show only dataset paths for get/post/patch/put, but add some text explaining that the same endpoints work for all schemas.

Some additional cleanup and corrections have been added to metastore and datastore specs.


clayliddell (approved):
lgtm

PR #3758, opened 2022-02-28
Multiple sorts in search

dafeder:
In some cases we may want to sort by multiple fields in the search results. This gives the API the ability to receive multiple values for both sort and sort-order.

## QA Steps

Because the search API returns full datasets it can be hard to read the results of test requests. These QA steps will work on the command line if you have the [jq](https://stedolan.github.io/jq/) utility installed in your local shell.

1. In a local instance with sample content or the [QA site](https://dkan-qa-3758.ci.civicactions.net/), create a new dataset called with the title "Afghanistan Election Districts".
2. `curl "[site-url]/api/1/search?sort[0]=title&sort[1]=modified&sort-order[0]=asc&sort-order[1]=asc" | jq '.results[] | {"title","identifier","modified"}'`
3. Confirm that there are two "Afghanistan Election Districts" datasets, at the top of the search results, and that the sample content one, with modified date `2012-10-30`, is first.
4. `curl "[site-url]/api/1/search?sort[0]=title&sort[1]=modified&sort-order[0]=asc&sort-order[1]=desc" | jq '.results[] | {"title","identifier","modified"}'`
5. Now confirm that the two Afghanistan datasets are still on top, but the sample dataset should be second.

## Known issues

I don't believe the search service is being properly tested by our unit test suite, even though it registers as having 100% coverage. I would like to do a second PR with new tests and consolidating the QueryBuilderTrait into Search.

### DIFF
--- a/modules/metastore/modules/metastore_search/docs/openapi_spec.json
+++ b/modules/metastore/modules/metastore_search/docs/openapi_spec.json
@@ -72,7 +72,7 @@
                     {
                         "name": "sort",
                         "in": "query",
-                        "description": "Which property to sort results on.",
+                        "description": "Which property to sort results on. Use array format to sort on multiple properties, e.g. sort[0]=field1&sort[1]=field2",
                         "schema": {
                             "type": "string"
                         },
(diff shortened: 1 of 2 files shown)


janette (approved):
Nice! Works great, I needed to add `-g` to the curl command to avoid a `bad range in URL position 44` error

PR #3760, opened 2022-03-04
Switch to comma-separated values for search API sorts

dafeder:
Slight redux of #3758 adding support for comma-separated multi-value (array) parameters `sort` and `sort-order`. This is more consistent with the other parameters in the DKAN search API.

## QA Steps

1. In the [QA site](https://dkan-qa-3760.ci.civicactions.net/), create a new dataset called with the title "Afghanistan Election Districts".
2. Retrieve https://dkan-qa-3760.ci.civicactions.net/api/1/search?sort=title,modified&sort-order=asc,asc
3. Confirm that there are two "Afghanistan Election Districts" datasets, at the top of the search results, and that the sample content one, with modified date `2012-10-30`, is first.
4. Retrieve https://dkan-qa-3760.ci.civicactions.net/api/1/search?sort=title,modified&sort-order=asc,desc
5. Now confirm that the two Afghanistan datasets are still on top, but the sample dataset should be second.

## Note

While no longer documented, the "deepObject" syntax described in #3758 will actually still work.

### DIFF
--- a/modules/metastore/modules/metastore_search/docs/openapi_spec.json
+++ b/modules/metastore/modules/metastore_search/docs/openapi_spec.json
@@ -72,23 +72,30 @@
                     {
                         "name": "sort",
                         "in": "query",
-                        "description": "Which property to sort results on. Use array format to sort on multiple properties, e.g. sort[0]=field1&sort[1]=field2",
+                        "description": "Which property to sort results on.",
                         "schema": {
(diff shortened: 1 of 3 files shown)


dgading (approved):
From a frontend testing perspective this is great. Works as I would expect when making calls through the React app.

PR #3764, opened 2022-03-14
Build datastore alter queue job

clayliddell:
fixes WCMS-8693

- [x] Test coverage exists
- [x] Documentation exists

## QA Steps

- [x] Create a data-dictionary using the API:
```bash
AUTHCOOKIE='your auth cookie';
curl 'http://2x.localtest.me/api/1/metastore/schemas/data-dictionary/items' -X POST \
-H 'Accept: application/json' \
-H 'Accept-Encoding: gzip, deflate' \
-H 'Cookie: $AUTHCOOKIE' \
--data-raw '
{
  "identifier": "data-dictionary",
  "title": "Test",
  "data": {
    "fields": [
      {
        "name": "a",
        "title": "Field A",
        "type": "number",
        "format": "default"
      },
      {
        "name": "b",
        "title": "Field B",
        "type": "date",
        "format": "%-m/%d/%y"
      },
      {
        "name": "c",
        "title": "Field C",
        "type": "integer",
        "format": "default"
      }
    ]
  }
}'
```
- [ ] Enable global data-dictionary setting, and set global data-dictionary ID.
```php
drush php
...
use \Drupal\metastore\DataDictionary\DataDictionaryDiscoveryInterface;
$data_dict_id = 'data-dictionary';
$config = \Drupal::configFactory()->getEditable('metastore.settings');
$config->set('data_dictionary_sitewide', $data_dict_id);
$config->set('data_dictionary_mode', DataDictionaryDiscoveryInterface::MODE_SITEWIDE);
$config->save();
exit
```
- [ ] Create a new dataset with the following CSV file: https://gist.githubusercontent.com/clayliddell/d4ab207343964a43a7f8c9feae6a790e/raw/250d4ce52f428d8cc1e7c368bce429d9a46fa9db/datastore_alter_queue_test.csv
- [ ] Run cron twice.
- [ ] Navigate to the dataset page.
- [ ] Verify that the dataset preview table loads correctly.
- [ ] Verify that the data types in the http://2x.localtest.me/api/1/datastore/imports/{dataset_id} request.


modules/datastore/src/DataDictionary/AlterTableQuery/MySQLQuery.php
```diff
@@ -0,0 +1,162 @@
...
+namespace Drupal\datastore\DataDictionary\AlterTableQuery;
+
+use Drupal\Core\Database\Connection;
+use Drupal\datastore\DataDictionary\AlterTableQueryInterface;
+use Drupal\datastore\DataDictionary\FrictionlessDateFormatConverterInterface;
+
+/**
+ * MySQL table alter query.
+ */
+class MySQLQuery implements AlterTableQueryInterface {
```
clayliddell, line 16: We need to sanitize user input to protect against SQL injection in this class.

modules/datastore/src/DataDictionary/FrictionlessDateFormatConverterBase.php
```diff
@@ -0,0 +1,88 @@
...
+ * | Year as a numeric, 4-digit value           | %Y               | ?         |
+ * | Year as a numeric, 2-digit value           | %y               | ?         |
+ * | A literal '%' character                    | %%               | ?         |
+ * | Locale date/time (e.g. 09/08/13 07:06:05)  | %c               | ?         |
+ * | Locale’s date (e.g. 09/08/13)              | %x               | ?         |
+ * | Locale’s time (e.g. 07:06:05)              | %X               | ?         |
+ * | UTC offset ±HHMM[SS[.ffffff]] (e.g. +0000) | %z               | ?         |
+ * | Time zone name (e.g. UTC)                  | %Z               | ?         |
+ */
+abstract class FrictionlessDateFormatConverterBase implements FrictionlessDateFormatConverterInterface {
```
clayliddell: I'm planning to completely refactor this class to reduce cognitive complexity and break it down into smaller, easier to understand, swapable pieces.

modules/datastore/src/DataDictionary/DateFormat/Parser.php
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * Break up supplied date format string into lexemes.
+   *
+   * @param string $input_format
+   *   Date format string to lex.
+   *
+   * @return \Drupal\datastore\DataDictionary\FrictionlessDateFormatConverter\TokenInterface[]
+   *   Token list.
+   */
+  protected function lex(string $input_format): array {
```
clayliddell: Even though this method has a high "cognitive complexity", I think that breaking it up into smaller methods would only decrease the readability of the `lex()` method. I've documented it as extensively as I know how, but this is one of those algorithms which is hard to break up, and difficult to understand the parts of in isolation. As such, I recommend this method be left composite.

dafeder: @clayliddell I agree, I think, about the cognitive complexity being acceptable here but I think we should think about isolating esoteric, self-contained/reusable things like parsers from our basic business logic. And if possible, avoid implementing and having to maintain them ourselves. Did you look into this library, which seems to provide strptime parsing for current PHP versions? 

https://packagist.org/packages/php81_bc/strftime

If this one is missing critical functionality I'd suggest we break off your date parser into its own library.

modules/datastore/src/DataDictionary/DateFormat/Converter.php
```diff
@@ -0,0 +1,41 @@
...
+ * Converts between date formats.
+ *
+ * The date format parser given is used to parse the input format supplied to
+ * `::convert()`, and the date format compiler is then used to generate the
+ * converted output format.
+ *
+ * The internal language used by ASTs which all parsers are supposed to convert
+ * input format's to is the Frictionless date format.
+ */
+class Converter implements DateFormatConverterInterface {
```
clayliddell: When discussing this PR on Slack, @dafeder suggested we break off the date format converter into it's own separate library, since it's a bit esoteric and isolated. I think this is a good idea and certainly worth pursuing.

composer.json
```diff
@@ -30,7 +30,8 @@
         "npm-asset/select2": "^4.0",
         "oomphinc/composer-installers-extender": "^2.0",
         "ramsey/uuid" : "^3.8.0",
-        "stolt/json-merge-patch": "^1.0"
+        "stolt/json-merge-patch": "^1.0",
+        "getdkan/pdlt": "0.1.2-alpha"
```
dafeder: Cut stable release

modules/datastore/src/Service/Import.php
```diff
@@ -25,6 +26,13 @@ class Import {
 
   const DEFAULT_TIMELIMIT = 50;
 
+  /**
+   * Default data dictionary ID.
+   *
+   * @var string
+   */
+  protected const DEFAULT_DICTIONARY_ID = 'string';
```
dafeder: Remove this

modules/datastore/src/Service/Import.php
```diff
@@ -59,6 +81,8 @@ class Import {
    * Constructor.
    */
   public function __construct(Resource $resource, JobStoreFactory $jobStoreFactory, DatabaseTableFactory $databaseTableFactory) {
+    $this->resourceId = $resource->getIdentifier();
+    $this->resourceVersion = $resource->getVersion();
     $this->initializeResource($resource);
```
dafeder: Let's either differentiate between the two resources objects here through method and variable names or add more docs to this class that explain what's going on.

modules/datastore/src/Plugin/QueueWorker/DictionaryEnforcer.php
```diff
@@ -0,0 +1,136 @@
...
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function processItem($data) {
+    $dict_id = $this->dataDictionaryDiscovery->dictionaryIdFromResource($data->resource->id, $data->resource->version);
+    if (!isset($dict_id)) {
+      throw new \UnexpectedValueException(sprintf('No data-dictionary found for resource with id "%s" and version "%s".', $data->resource->id, $data->resource->version));
```
dafeder: Let's log this but not throw an exception so that we don't just keep requeing forever

modules/datastore/src/DataDictionary/AlterTableQuery/MySQLQuery.php
```diff
@@ -0,0 +1,220 @@
...
+   *   MySQL table to get type for.
+   *
+   * @return string
+   *   MySQL data type.
+   */
+  protected function getType(string $frictionless_type, string $column, string $table): string {
+    // Build the MySQL type argument list for the given Frictionless data type.
+    $args = $this->buildTypeArgs($frictionless_type, $column, $table);
+
+    // Build full MySQL type for the given Frictionless data type.
```
dafeder: Be more precise with naming. The name of the standard we're using is "Table Schema" from the "Frictionless Data" project/ecosystem/whatever  -- using "frictionless" on its own is too general because there are many specs in the Frictionless Data project, but maybe "table schema" is too generic and could be confused with Drupal's own table schema. Figure out a way to describe more clearly without being too clunky

dafeder (changes requested):

clayliddell:
The "Unused use statement" warning in CodeClimate is inaccurate. The use statement in question is being used.

modules/datastore/src/Service/Import.php
```diff
@@ -161,13 +121,7 @@ public function import() {
...
         // Queue the imported resource for data-dictionary enforcement.
         $dictionary_enforcer_queue = \Drupal::service('queue')->get('dictionary_enforcer');
-        $dictionary_enforcer_queue->createItem((object) [
-          'datastore_table' => $this->getStorage()->getTableName(),
-          'resource' => (object) [
-            'id' => $this->getResourceId(),
-            'version' => $this->getResourceVersion(),
-          ],
-        ]);
+        $dictionary_enforcer_queue->createItem($resource);
```
clayliddell, line 124: It would be better to create a queue item for the specific resource version, and check in the enforcer queue that the version matches.

Don't throw exception, or log error or warning. Better to instead just log it as a notice.

modules/datastore/src/Service/Import.php
```diff
@@ -96,46 +88,12 @@ public function setImporterClass($className) {
...
-
-  /**
-   * Getter.
-   *
-   * @return \Dkan\Datastore\Resource
-   *   Resource.
-   *
-   * @codeCoverageIgnore
-   */
-  protected function getResource() : DatastoreResource {
```
clayliddell, line 96: We can get rid of this method and just access the property directly.

modules/datastore/tests/src/Unit/Plugin/QueueWorker/DictionaryEnforcerTest.php
```diff
@@ -21,25 +26,26 @@
  */
 class DictionaryEnforcerTest extends TestCase {
```
clayliddell, line 28: Create a functional test for the entire import and data-dictionary application workflow.

Possibly API driven. Write this one in PHPUnit.

modules/datastore/src/Service/Import.php
clayliddell: I'm going to hold off on removing this method since it looks like there's quite a few tests which are currently dependent on it.

modules/datastore/src/Service/Import.php
```diff
@@ -21,9 +20,19 @@ class Import {
   use LoggerTrait;
   use EventDispatcherTrait;
 
-  const EVENT_CONFIGURE_PARSER = 'dkan_datastore_import_configure_parser';
+  /**
+   * Event name used when configuring the parser during import.
+   *
+   * @var string
+   */
+  protected const EVENT_CONFIGURE_PARSER = 'dkan_datastore_import_configure_parser';
```
clayliddell:
```suggestion
  public const EVENT_CONFIGURE_PARSER = 'dkan_datastore_import_configure_parser';
```
Event constants should be public.

dafeder:
@clayliddell FYI on initial test I got an error for your sample JSON, but it was because you had "int" instead of "integer"...

clayliddell:
> @clayliddell FYI on initial test I got an error for your sample JSON, but it was because you had "int" instead of "integer"...

Whoops, updated.

dafeder (dismissed):
I removed the "identifier" from the dictionary JSON (after submitting it was assigned identifier `abfc5cba-5eeb-5c6d-bddb-2270baefe334`). Also, used a simpler set of drush commands for config:
```
dktl drush config:set metastore.settings data_dictionary_sitewide 'abfc5cba-5eeb-5c6d-bddb-2270baefe334'
dktl drush config:set metastore.settings data_dictionary_mode 'sitewide'
```

After running the first `drush cron`, I got the following error:

```
[error]  Invalid format provided; token "%-y" not found in grammar "PDLT\Grammar\Strptime".
```
And no `dictionary_enforcer` queue job was created.

clayliddell:
@dafeder Your error appears to be a valid one. `%-y` doesn't appear to be a universally accepted directive according to the Python documentation linked on the Frictionless site ([here](https://docs.python.org/2/library/datetime.html#strftime-strptime-behavior)).

composer.json
```diff
@@ -30,7 +30,8 @@
         "npm-asset/select2": "^4.0",
         "oomphinc/composer-installers-extender": "^2.0",
         "ramsey/uuid" : "^3.8.0",
-        "stolt/json-merge-patch": "^1.0"
+        "stolt/json-merge-patch": "^1.0",
+        "getdkan/pdlt": "0.1.3"
```
clayliddell:
```suggestion
        "getdkan/pdlt": "~0.1"
```

docs_assets/pages/3.1_metastore.md
```diff
@@ -24,3 +24,21 @@ To change the schema being used, copy the contents of the `schema` directory fro
 
 @warning
   Warning: The schema is actively used by the catalog to verify the validity of the data. Making changes to the schema after data is present in the catalog should be done with care as non-backward-compatible changes to the schema could cause issues. Look at Drupal::metastore::SchemaRetriever::findSchemaDirectory() for context.
+
+## Data Dictionaries
```
janette, line 28: @clayliddell Will all three options require that the **Datastore MySQL Import** module be enabled?

clayliddell: No, the datastore mysql import module is not required.

cypress.json
```diff
@@ -2,10 +2,9 @@
   "apiUri": "api/1",
   "env": {
     "TEST_USER_CREDENTIALS": {
-      "user": "testadmin",
-      "pass": "testadmin"
+      "user": "testapiuser",
+      "pass": "testapiuser"
```
clayliddell:
```suggestion
      "user": "testadmin",
      "pass": "testadmin"
```

janette (approved):

janette (approved):

PR #3766, opened 2022-03-15
Add moderation state to admin dataset view. Add hide action to VBO.

janette:
- Visit the /admin/dkan/datasets screen
- Confirm you can see the moderation state of the datasets in the view table
- Confirm you can also use bulk operations to change multiple datasets to the hidden state
- Review documentation updates to Glossary


modules/metastore/modules/metastore_admin/src/Plugin/Action/HideCurrentRevisionAction.php
```diff
@@ -0,0 +1,63 @@
...
+ * If type is left empty, action will be selectable for all
+ * entity types.
+ *
+ * @Action(
+ *   id = "hide_current_revision_action",
+ *   label = @Translation("Hide Current Revision"),
+ *   type = "node",
+ *   confirm = TRUE,
+ * )
+ */
```
dafeder, line 26:
```suggestion
 * )
 *
 * @codeCoverageIgnore
 */
 Let's try not to get in trouble with code climate :)
```

dafeder (changes requested):

dafeder:
@janette when trying the "hide" action I get a server error. Message is
```
Error: Call to undefined method Drupal\moderated_content_bulk_publish\AdminModeration::hide() in Drupal\metastore_admin\Plugin\Action\HideCurrentRevisionAction->execute() (line 41 of /var/www/dkan/modules/metastore/modules/metastore_admin/src/Plugin/Action/HideCurrentRevisionAction.php) 
```

janette:
**Remaining task:** 
Create a code coverage test for https://github.com/GetDKAN/dkan/blob/WCMS-8695/modules/metastore/modules/metastore_admin/src/Plugin/Action/HideCurrentRevisionAction.php


The new VBO option is tested in two Cypress tests but we need coverage that codeclimate will recognize.

dafeder (approved):

PR #3769, opened 2022-03-21
Remove bulk moderation modals in Cypress tests

dafeder:
The moderated_content_bulk_publish module has helped with the functionality around moderation states, but it also has introduced a lot of confirmation dialogs that are breaking our tests. Hopefully these modals will be optional soon - see https://www.drupal.org/project/moderated_content_bulk_publish/issues/3185848

In the meantime, this PR adds the extra click needed to our cypress tests.

## QA Steps

- [ ] Tests should not fail on form submit


dgading (approved):
Cypress tests look good.

PR #3771, opened 2022-03-21
Upgrade CircleCI to Ubuntu 20.04

dafeder:
Our current executor image on CircleCI has been deprecated. See: https://circleci.com/blog/ubuntu-14-16-image-deprecation/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,7 +80,7 @@ commands:
 jobs:
   phpunit:
     machine:
-      image: circleci/classic:latest
+      image: ubuntu-2004:current
     parameters:
       upgrade:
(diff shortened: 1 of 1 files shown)


clayliddell (approved):

clayliddell:
CircleCI recommends using convenience images over machine images due to them being prebuilt, more stable, and spinning up faster (see https://circleci.com/docs/2.0/circleci-images/#next-generation-convenience-images). I'd like to recommend we use [cimg/base:current](https://circleci.com/developer/images/image/cimg/base) instead, since it's the convenience image equivalent of [ubuntu-2004](https://circleci.com/developer/machine/image/ubuntu-2004) and once Ubuntu 22.04 is consider stable, it should automatically switch over.

PR #3779, opened 2022-04-28
Update test users

janette:
REF: WCMS-6450
Companion PR to: https://github.com/GetDKAN/dkan-tools/pull/328

- [ ] Test coverage exists
- [ ] Documentation exists

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -88,7 +88,7 @@ jobs:
         type: boolean
     environment:
       TEST_RESULTS: /tmp/test-results
-      DKTL_VERSION: "4.2.9"
+      DKTL_VERSION: "dev-adding_testing_accounts_module_to_testing_command"
     steps:
       - prepare_build:
           upgrade: << parameters.upgrade >>
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -88,7 +88,7 @@ jobs:
         type: boolean
     environment:
       TEST_RESULTS: /tmp/test-results
-      DKTL_VERSION: "adding_testing_accounts_module_to_testing_command"
+      DKTL_VERSION: "4.2.10-RC"
     steps:
       - prepare_build:
           upgrade: << parameters.upgrade >>
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -88,7 +88,7 @@ jobs:
         type: boolean
     environment:
       TEST_RESULTS: /tmp/test-results
-      DKTL_VERSION: "dev-adding_testing_accounts_module_to_testing_command"
+      DKTL_VERSION: "adding_testing_accounts_module_to_testing_command"
     steps:
       - prepare_build:
           upgrade: << parameters.upgrade >>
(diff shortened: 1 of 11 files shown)


.circleci/config.yml
```diff
@@ -88,7 +88,7 @@ jobs:
         type: boolean
     environment:
       TEST_RESULTS: /tmp/test-results
-      DKTL_VERSION: "4.2.9"
```
clayliddell, line 91: What's with the "RC"? "Release Candidate" maybe?

.codeclimate.yml
```diff
@@ -1,8 +1,8 @@
 version: 2
 plugins:
   phpcodesniffer:
-    enabled: true
-    channel: beta
```
clayliddell: We should revert these changes since it doesn't look like they solved our CodeClimate woes.

cypress/integration/09_admin_links.spec.js
```diff
@@ -1,7 +1,8 @@
 context('Administration pages', () => {
   let baseurl = Cypress.config().baseUrl;
   beforeEach(() => {
-      cy.drupalLogin('testeditor', 'testeditor')
+    const user_credentials = Cypress.env('TEST_USER_CREDENTIALS')
+    cy.drupalLogin(user_credentials.user, user_credentials.pass)
```
clayliddell, line 5: Nice! I like this. It's a lot cleaner.

cypress/integration/11_admin_dataset_file_upload.spec.js
```diff
@@ -82,8 +85,8 @@ context('Admin dataset file upload', () => {
 
     before(() => {
       const selectorDist = '#edit-field-json-metadata-0-value-distribution-distribution-0-distribution-downloadurl-upload'
-
```
clayliddell, line 85: Could we add this const out into the context since it's used in two separate places in the context?

cypress/integration/11_admin_dataset_file_upload.spec.js
```diff
@@ -126,7 +129,8 @@ context('Admin dataset file upload', () => {
     })
 
     beforeEach(() => {
-      cy.drupalLogin('testeditor', 'testeditor')
```
clayliddell, line 129: See previous comment.

clayliddell (changes requested):
Mostly looks good. I left a bit of feedback, some questions, and some comments.

clayliddell:
Since only 1 line of PHP was modified in this PR, I'm not to worried about PHP CS on this :smiley:

.circleci/config.yml
janette: @clayliddell yes I am testing the changes on [DKAN-Tools](https://github.com/GetDKAN/dkan-tools/pull/328)

.codeclimate.yml
janette: Agree

clayliddell (approved):

PR #3782, opened 2022-05-10
Fix 2 operationIds mistakenly switched in datastore/docs/openapi_spec…

thierrydallacroce:
This should fix the identifiers of 2 operations in the datastore's `openapi_spec.json`. The two GETs should now more closely match their related POST and path.

Screen Shot 2022-05-10 at 1 02 04 PM


dafeder (approved):
Looks good

PR #3783, opened 2022-05-10
Fix dataset-specific docs endpoint

thierrydallacroce:
When the dataset-specific API docs is generated, it's created from the full spec, whose paths were already prefixed with `$settings['dkan_api_base']`. As a result, `$fullSpec['paths']['/api/1/...']` is not found, resulting in breaking the dataset-specific API docs.

This PR attempts to correct that.

Since the prefix is now needed in more than one function, I made it a class property, initialized in the constructor. This allows the removal of the settings property.

### DIFF
--- a/modules/metastore/tests/src/Unit/DatasetApiDocsTest.php
+++ b/modules/metastore/tests/src/Unit/DatasetApiDocsTest.php
@@ -1,0 +1,177 @@
+<?php
+
+namespace Drupal\Tests\metastore\Unit;
+
+use Drupal\Core\Site\Settings;
+use Drupal\common\DkanApiDocsGenerator;
+use Drupal\common\Plugin\DkanApiDocsBase;
+use Drupal\common\Plugin\DkanApiDocsPluginManager;
+use Drupal\metastore\DatasetApiDocs;
+use Drupal\metastore\SchemaRetriever;
+use Drupal\metastore\Service;
+use Drupal\metastore\Storage\DataFactory;
+use Drupal\metastore\ValidMetadataFactory;
+use MockChain\Chain;
+use PHPUnit\Framework\TestCase;
+use RootedData\RootedJsonData;
+
+/**
+ * Unit tests for DatasetApiDocs.
+ */
+class DatasetApiDocsTest extends TestCase {
+
+  /**
+   *
+   */
+  public function testBuildDatasetApiDocsWithDkanApiBase() {
+
+    $dkanApiPath = '/foo/bar';
+
+    $settings = new Settings(['dkan_api_base' => $dkanApiPath]);
+
+    $generator = new DkanApiDocsGenerator(
+      $this->getManagerChain()->getMock(),
+      $settings
+    );
+
+    $service = $this->getServiceChain()->getMock();
+
+    $datasetDoc = new DatasetApiDocs($generator, $service, $settings);
+
+    $spec = $datasetDoc->getDatasetSpecific('123');
+
+    $this->assertTrue(is_array($spec['paths']));
+
+    $paths = array_keys($spec['paths']);
+    $expected_paths = [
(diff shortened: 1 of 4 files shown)


dafeder:
@thierrydallacroce does this definitely fix the issue for you? I've updated this and will merge if all the docs still work as expected on a "regular" site and you confirm it works on a site with a non-standard base path.

thierrydallacroce:
@dafeder, yes I am using this fix in a custom branch based off 2.13.8, and it's working as intended.

thierrydallacroce:
@dafeder, I updated this branch.

thierrydallacroce:
@dafeder, updated on top of the current `2.x`

modules/metastore/tests/src/Unit/DatasetApiDocsTest.php
```diff
@@ -0,0 +1,177 @@
...
+
+    $datasetDoc = new DatasetApiDocs($generator, $service, $settings);
+
+    $spec = $datasetDoc->getDatasetSpecific('123');
+
+    $this->assertTrue(is_array($spec['paths']));
+
+    $paths = array_keys($spec['paths']);
+    $expected_paths = [
+      $dkanApiPath . '/api/1/metastore/schemas/dataset/items/123',
```
dafeder, line 47: Sorry maybe I'm missing something... if I understand the PR correctly wouldn't we want to see /foo/bar/api/1... as the path here?

modules/metastore/tests/src/Unit/DatasetApiDocsTest.php
```diff
@@ -0,0 +1,177 @@
...
+    foreach ($expected_paths as $path) {
+      $this->assertContains($path, $paths);
+    }
+  }
+
+  /**
+   *
+   */
+  private function getSpec(): array {
+    return [
```
dafeder, line 61: Seems like there's a lot more here than is needed for the test? I'd think you really only need one path and no components/parameters.

dafeder (changes requested):

modules/metastore/tests/src/Unit/DatasetApiDocsTest.php
jastraat: Yes, that's correct. That's what $dkanApiPath is (/foo/bar). The PR appends that value to 4 separate paths, so I'm testing all 4.

modules/metastore/tests/src/Unit/DatasetApiDocsTest.php
jastraat: I know it seems like a lot, but I only put the minimum amount it for it not to fail. I started with just the 4 paths modified by the PR, but that didn't work.  The getDatabaseSpecific() function expect a LOT of arrays to exist and if they don't, it fails. I had to keep adding and adding over the course of 4 hours in order for it to complete. 

For example:
$fullSpec['paths']['/api/1/metastore/schemas/dataset/items/{identifier}']['get']; <- it expects the get which then requires a bunch of additional objects.

And then there's https://github.com/GetDKAN/dkan/blob/15a2ed0c6d0b4e5f7c94f92c72f99fc6f0d878b0/modules/metastore/src/DatasetApiDocs.php#L168 and https://github.com/GetDKAN/dkan/blob/15a2ed0c6d0b4e5f7c94f92c72f99fc6f0d878b0/modules/metastore/src/DatasetApiDocs.php#L145 all all the array values those expect (and all their dependencies).

modules/metastore/tests/src/Unit/DatasetApiDocsTest.php
dafeder: oh right [face-palm emoji]

dafeder (approved):

PR #3784, opened 2022-05-16
Turn off strict mode on import

dafeder:
Strict mode on innodb makes it hard to create tables over 200 columns, which has come up in real world use. This sets strict mode off simply at the session level for the datastore, and keeps it on for the rest of Drupal's operation.

This has been tested locally and successfully created a 252 column table in the same environment where it had previously failed.


modules/datastore/src/Plugin/QueueWorker/Import.php
```diff
@@ -142,6 +143,16 @@ protected function setConnectionTimeout(Connection $connection, int $timeout): v
     $connection->query($command);
   }
 
+  /**
+   * Set the wait_timeout for the given database connection.
```
clayliddell:
```suggestion
   * Disable InnoDB strict mode for the given database connection.
```

clayliddell (changes requested):
Looks good!

Just one documentation suggestion.

dafeder:
Hmm committed your suggestions @clayliddell but seeing it's failing one of the PHPunit runs, but not the other. Weird, will take another look.

dafeder:
Moved this into the datastore's DatabaseTable class, after realizing that's where the table creation is actually happening.

dafeder:
Moved this into the datastore's DatabaseTable class, after realizing that's where the table creation is actually happening.

modules/datastore/src/Storage/DatabaseTable.php
```diff
@@ -158,6 +159,19 @@ public function setSchema($schema) {
...
+  /**
+   * Disable/enable InnoDB strict mode for the given database connection.
+   *
+   * @param string $value
+   *   "ON" or "OFF".
+   */
+  public function setInnodbMode(string $value = "OFF"): void {
+    // Only if we're using MySQL.
+    if ($this->connection instanceof MysqlConnection) {
+      $this->connection->query("SET SESSION innodb_strict_mode=:value", [':value' => $value]);
```
clayliddell, line 172: This method will throw a database exception if `$value` is not "ON" or "OFF". Could we make the `$value` parameter a boolean and rename the function appropriately, or just add a check to make sure `$value` is either "ON" or "OFF"?

clayliddell (changes requested):
Just 1 small thing. Trying not to be picky, but I wanted to call this out.

modules/datastore/src/Storage/DatabaseTable.php
dafeder: Yeah good point, I was rushing here. Will do.

dafeder: @clayliddell see now

clayliddell: :raised_hands: Awesome! Love it!

clayliddell (approved):

PR #3790, opened 2022-06-07
Separate query service from main datastore service

dafeder:
Monolithic \Drupal\datastore\Service class too bloated. It is already at 20 methods which means we can't add any more without failing codeclimate, and it's far too many things for a single class to be responsible for.

This would require a new minor release with a clear warning in the release notes, as it would break any custom code that relies on the query methods of the datastore service class.


clayliddell (approved):
Great work. LGTM!

PR #3795, opened 2022-06-21
Data-Dictionary Settings Form

clayliddell:
fixes WCMS-8869

- [x] Test coverage exists
- [ ] Documentation exists

## QA Steps

- [ ] Log into DKAN.
- [ ] Navigate to DKAN > Data-Dictionary Settings.
- [ ] Confirm the new settings are present, and if modified are preserved on save and reload.

### DIFF
--- a/modules/metastore/tests/src/Kernel/DataDictionarySettingsFormTest.php
+++ b/modules/metastore/tests/src/Kernel/DataDictionarySettingsFormTest.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace Drupal\Tests\metastore\Kernel;
+
+use Drupal\KernelTests\ConfigFormTestBase;
+use Drupal\metastore\Form\DataDictionarySettingsForm;
+use Drupal\metastore\DataDictionary\DataDictionaryDiscoveryInterface;
+
+/**
+ * Data Dictionary Settings Form class test.
+ *
+ * @group Form
+ */
+class DataDictionarySettingsFormTest extends ConfigFormTestBase {
+
+  /**
+   * Modules to enable.
+   *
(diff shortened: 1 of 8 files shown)


modules/metastore/tests/src/Kernel/DataDictionarySettingsFormTest.php
```diff
@@ -0,0 +1,46 @@
...
+
+    $this->form = new DataDictionarySettingsForm($this->container->get('config.factory'));
+    $this->values = [
+      'dictionary_mode' => [
+        '#value' => DataDictionaryDiscoveryInterface::MODE_SITEWIDE,
+        '#config_name' => DataDictionarySettingsForm::SETTINGS,
+        '#config_key' => 'dictionary_mode',
+      ],
+    ];
+    $this->values = [
```
paul-m: Doesn't this overwrite the previous value?

clayliddell: Yeah, I'll need to fix that

modules/metastore/tests/src/Kernel/DataDictionarySettingsFormTest.php
```diff
@@ -0,0 +1,46 @@
...
```
clayliddell:
```suggestion
```

janette (approved):

janette:
@clayliddell should I see the property_list once I save a dictionary source ID? and if I have created a data-dictionary or two, should the ID field auto-complete? or be a select list? Should I get an error when I enter a non-existing ID?

PR #3810, opened 2022-07-26
Fix how datastore DB connection settings are applied

clayliddell:
Fix faulty array merge when DB connection settings are generated. Since 2.13.11 CSV downloads were not working (headers would be sent but no data on any datasets larger than a few thousand rows) because the PDO::MYSQL_ATTR_USE_BUFFERED_QUERY connection setting was not being set correctly.


dafeder (approved):
Tested locally and confirmed fix

PR #3812, opened 2022-07-29
Fulltext searching in datastore query

dafeder:
This is a somewhat experimental feature for now and will not be fully documented until we have used it a bit more in the wild. But this allows fulltext boolean match queries against a datastore resource in MySQL if the column(s) being queries are properly indexed.

This PR also adds two new keys to the schema object in the datastore response. One is `indexes` which lists all standard indexes on the database table, as per the Drupal Schema API spec. Another is `fulltext indexes`, which follows the same pattern. This is not part of the official Drupal Schema spec but I felt was needed to signal to the user what columns are available for fulltext searching, and matches the rest of the spec as closely as possible.

## QA steps

This will need to be test locally since for now we need to create the fulltext indexes manually. Set up a local site with the demo content.

1. Ensure that you have a datastore at /api/1/datastore/query/d460252e-d42c-474a-9ea9-5287b1d595f6/0
2. Try running a query before indexing. You should get a 400 error "You have attempted a fulltext match against a column that is not indexed for fulltext searching."
```http
POST https://fulltext-query.ddev.site/api/1/datastore/query/d460252e-d42c-474a-9ea9-5287b1d595f6/0

{
  "conditions": [
    {
      "property": "city",
      "operator": "match",
      "value": "York"
    }
  ]
}
```
3. Use `drush dkan:dataset-info d460252e-d42c-474a-9ea9-5287b1d595f6` to get the table name of the datastore. Let's say its `datastore_efee6d8c0ee35d0b1a161743806d018d`.
4. Connect to your db CLI with `drush sqlc`.
5. Create a fulltext index on the city field: 
```sql
ALTER TABLE datastore_efee6d8c0ee35d0b1a161743806d018d ADD FULLTEXT INDEX `cityfulltext` (`city`);
```
6. Retry the query from step 2.
7. Confirm that you see a result and got a single hit for "New York."
8. Try again with a wildcard:
```http
POST https://fulltext-query.ddev.site/api/1/datastore/query/d460252e-d42c-474a-9ea9-5287b1d595f6/0

{
  "conditions": [
    {
      "property": "city",
      "operator": "match",
      "value": "Yor*"
    }
  ]
}
```
9. Confirm that you still get a hit for "New York".
10. Check the `schema` object in the response and confirm that the fulltext index is listed properly.
11. Now let's try an index with two columns. Go back to the db CLI and let's create a second index that combines city and state:
```sql
ALTER TABLE datastore_efee6d8c0ee35d0b1a161743806d018d ADD FULLTEXT INDEX `citystate` (`city`, `state`);
```
12. We can try a few queries that should demonstrate how this can be leveraged:
```http
POST https://fulltext-query.ddev.site/api/1/datastore/query/d460252e-d42c-474a-9ea9-5287b1d595f6/0

{
  "conditions": [
    {
      "property": "city,state",
      "operator": "match",
      "value": "San"
    }
  ]
}
```
```http
POST https://fulltext-query.ddev.site/api/1/datastore/query/d460252e-d42c-474a-9ea9-5287b1d595f6/0

{
  "conditions": [
    {
      "property": "city,state",
      "operator": "match",
      "value": "Texas"
    }
  ]
}
```
13. Confirm that in both cases you get multiple hits from both city and state fields. Also confirm that the newer fulltext index is included correctly in the schema object.

### DIFF
--- a/modules/common/tests/src/Unit/Storage/QueryDataProvider.php
+++ b/modules/common/tests/src/Unit/Storage/QueryDataProvider.php
@@ -37,6 +37,7 @@ class QueryDataProvider {
       'likeConditionQuery',
       'containsConditionQuery',
       'startsWithConditionQuery',
+      'matchConditionQuery',
       'arrayConditionQuery',
       'nestedConditionGroupQuery',
       'sortQuery',
@@ -376,6 +377,31 @@ class QueryDataProvider {
 
       case self::VALUES:
         return ['value%'];
+    }
+  }
+
+  public static function matchConditionQuery($return) {
+    switch ($return) {
+      case self::QUERY_OBJECT:
+        $query = new Query();
+        $query->conditions = [
+          (object) [
+            "collection" => "t",
+            "property" => "field1",
+            "value" => "value",
+            "operator" => "match",
+          ],
+        ];
+        return $query;
+
+      case self::SQL:
+        return "WHERE (MATCH(t.field1) AGAINST (:words IN BOOLEAN MODE))";
+
+      case self::EXCEPTION:
+        return '';
+
+      case self::VALUES:
+        return ['value'];
     }
   }
 
--- a/modules/common/tests/src/Unit/Storage/QueryDataProvider.php
+++ b/modules/common/tests/src/Unit/Storage/QueryDataProvider.php
@@ -391,17 +391,23 @@ class QueryDataProvider {
             "value" => "value",
             "operator" => "match",
           ],
+          (object) [
+            "collection" => "t",
+            "property" => "field2",
+            "value" => "value2",
+            "operator" => "match",
+          ],
         ];
         return $query;
 
       case self::SQL:
-        return "WHERE (MATCH(t.field1) AGAINST (:words IN BOOLEAN MODE))";
+        return "WHERE ((MATCH(t.field1) AGAINST (:words0 IN BOOLEAN MODE))) AND ((MATCH(t.field2) AGAINST (:words1 IN BOOLEAN MODE)))";
 
       case self::EXCEPTION:
         return '';
 
       case self::VALUES:
-        return ['value'];
+        return ['value', 'value2'];
     }
   }
 
--- a/modules/datastore/src/Storage/DatabaseTable.php
+++ b/modules/datastore/src/Storage/DatabaseTable.php
@@ -180,7 +180,6 @@ class DatabaseTable extends AbstractDatabaseTable implements \JsonSerializable {
    *
    * @param string $tableName
    *   The table name.
-   *
    * @param array $fieldsInfo
    *   Array of fields info from DESCRIBE query.
    *
--- a/modules/datastore/src/Storage/DatabaseTable.php
+++ b/modules/datastore/src/Storage/DatabaseTable.php
@@ -186,7 +186,7 @@ class DatabaseTable extends AbstractDatabaseTable implements \JsonSerializable {
    * @return array
    *   Full Drupal Schema API array.
    */
-  protected function buildTableSchema(string $tableName, array $fieldsInfo) {
+  protected function buildTableSchema(string $tableName, array $fieldsInfo): array {
     // Add descriptions to schema from column comments.
     $canGetComment = method_exists($this->connection->schema(), 'getComment');
     $schema = ['fields' => []];
@@ -210,7 +210,7 @@ class DatabaseTable extends AbstractDatabaseTable implements \JsonSerializable {
    * @param array $schema
    *   Drupal Schema API array.
    */
-  protected function addIndexInfo(array &$schema) {
+  protected function addIndexInfo(array &$schema): void {
     if ($this->connection->getConnectionOptions()['driver'] != 'mysql') {
       return;
     }
--- a/modules/datastore/src/Storage/DatabaseTable.php
+++ b/modules/datastore/src/Storage/DatabaseTable.php
@@ -190,6 +190,7 @@ class DatabaseTable extends AbstractDatabaseTable implements \JsonSerializable {
   protected function buildTableSchema(string $tableName, array $fieldsInfo) {
     // Add descriptions to schema from column comments.
     $canGetComment = method_exists($this->connection->schema(), 'getComment');
+    $schema = ['fields' => []];
     foreach ($fieldsInfo as $info) {
       $name = $info->Field;
       $schema['fields'][$name] = $this->translateType($info->Type, ($info->Extra ?? NULL));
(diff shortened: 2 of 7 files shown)


dafeder:
CodeClimate is complaining about 21 methods in SelectFactory. I think a refactor there that would actually lead to cleaner code would be a big lift and make more sense as part of work to pull AbstractDatbaseTable and its derivatives out of the rest of DKAN. That's way out of scope for this issue so I'd argue that we should mark this as a known issue for now but give it a pass. I'll leave that up to whoever reviews this though.

modules/datastore/tests/src/Unit/Storage/DatabaseTableTest.php
```diff
@@ -67,6 +69,17 @@ public function testGetSchema() {
...
+        "idx1" => [
+          "first_name",
+        ],
+      ],
+      "fulltext indexes" => [
+        "ftx1" => [
+          "first_name",
+          "last_name",
+        ],
+      ],
```
clayliddell, line 82: Could we have one "indexes" array, similar to what we're planning to do for data-dictionaries?
e.g.
```
"indexes" => [
  "index_name" => {
    "fields" => [
      "first_name",
      "last_name",
    ],
    "type" => "fulltext" or "btree" or "spatial",
  },
]
```

modules/datastore/src/Storage/DatabaseTable.php
```diff
@@ -188,7 +198,34 @@ protected function buildTableSchema($tableName, $fieldsInfo) {
...
+    return $schema;
+  }
+
+  /**
+   * Add index information to table schema.
+   *
+   * @param array $schema
+   *   Drupal Schema API array.
+   */
+  protected function addIndexInfo(array &$schema) {
```
clayliddell: just adding a return type
```suggestion
  protected function addIndexInfo(array &$schema): void {
```

modules/datastore/src/Storage/DatabaseTable.php
```diff
@@ -174,12 +174,22 @@ public function innodbStrictMode(bool $on) {
...
+   * @param string $tableName
+   *   The table name.
+   * @param array $fieldsInfo
+   *   Array of fields info from DESCRIBE query.
+   *
+   * @return array
+   *   Full Drupal Schema API array.
    */
-  protected function buildTableSchema($tableName, $fieldsInfo) {
+  protected function buildTableSchema(string $tableName, array $fieldsInfo) {
```
clayliddell, line 189: just adding a return type
```suggestion
  protected function buildTableSchema(string $tableName, array $fieldsInfo): array {
```

modules/datastore/tests/src/Unit/Storage/DatabaseTableTest.php
```diff
@@ -414,6 +427,27 @@ public function testQueryColumnNotFound() {
...
+   *
+   */
+  public function testNoFulltextIndexFound() {
+    $query = new Query();
+
+    $connectionChain = $this->getConnectionChain()
+      ->add(Connection::class, 'select', Select::class, 'select_1')
+      ->add(Select::class, 'fields', Select::class)
+      ->add(Select::class, 'condition', Select::class)
+      ->add(Select::class, 'execute', new DatabaseExceptionWrapper("SQLSTATE[HY000]: General error: 1191 Can't find FULLTEXT index matching the column list..."));
```
clayliddell, line 440: I think we would benefit from having a integration/functional test here to ensure the underlying MySQL error doesn't change in future versions.

modules/common/tests/src/Unit/Storage/QueryDataProvider.php
```diff
@@ -379,6 +380,31 @@ public static function startsWithConditionQuery($return) {
     }
   }
 
+  public static function matchConditionQuery($return) {
```
clayliddell, line 383: I don't think there's currently a test for this method.

clayliddell (commented):
The overall structure looks good. I've left a couple suggestions related to test coverage and one comment related to the new index portion of the schema.

modules/datastore/tests/src/Unit/Storage/DatabaseTableTest.php
dafeder: That is what I think we should do if we switch to Frictionless' table schema for the schema object. For better or worse we are still using the [Drupal schema API](https://api.drupal.org/api/drupal/core!lib!Drupal!Core!Database!database.api.php/group/schemaapi/9.0.x) here.

modules/common/tests/src/Unit/Storage/QueryDataProvider.php
dafeder: I'm confused... this method is part of a test

modules/datastore/tests/src/Unit/Storage/DatabaseTableTest.php
dafeder: I see your point... FWIW the only part being matched is "Can't find FULLTEXT index matching the column list" which I think is unlikely to change anytime soon. And if it did it would just fall back to our generic error message which wouldn't be a big deal. 

I'm hesitant because we don't actually have any functional tests for the datastore right now and there's a fairly high LOE for adding that into this PR, and I don't know that this particular thing justifies doing that. We should reexamine testing for the datastore API but IMO that's out of scope for this ticket. We can discuss further if you feel strongly though.

dafeder:
Whoops somehow merging those suggestions broke everything @clayliddell - will investigate. FWIW I'm not a big fan of that feature in github. Also I noticed a potential issue in my code - I just have that one placeholder `:words` for the matching terms, which will cause an issue if someone adds multiple MATCH conditions to a single query. I'll add an iterator to the end of it so we get `:words0` `words1` etc.

modules/datastore/tests/src/Unit/Storage/DatabaseTableTest.php
clayliddell: No, I kind of agree - we should hold off for now, at least pending datastore refactor efforts

modules/datastore/src/Storage/DatabaseTable.php
```diff
@@ -174,12 +174,22 @@ public function innodbStrictMode(bool $on) {
...
+   * @param string $tableName
+   *   The table name.
+   * @param array $fieldsInfo
+   *   Array of fields info from DESCRIBE query.
+   *
+   * @return array
+   *   Full Drupal Schema API array.
    */
-  protected function buildTableSchema($tableName, $fieldsInfo) {
+  protected function buildTableSchema(string $tableName, array $fieldsInfo): array {
```
clayliddell: I believe this is the one causing the problem:
```suggestion
  protected function buildTableSchema(string $tableName, array $fieldsInfo) {
```

modules/datastore/tests/src/Unit/Storage/DatabaseTableTest.php
clayliddell: I'm fine with just sticking to the Drupal schema for now - may be worth revisiting later though

clayliddell (approved):

PR #3817, opened 2022-08-18
Use "Data: Create new content" permission for data-dictionary menu access

clayliddell:
Currently only administrators have access to the **DKAN > Data-Dictionary** menu. This changes that so that data-publishers and anyone with permission to create data content also have access.

- [x] Test coverage exists
- [x] Documentation exists

## QA Steps

- [ ] Ensure that as a user with the "Data: Create new content" permission you have access to the **DKAN > Data-Dictionary** menu.

### DIFF
--- a/modules/metastore/metastore.routing.yml
+++ b/modules/metastore/metastore.routing.yml
@@ -165,7 +165,7 @@ metastore.data_dictionary:
     _controller: '\Drupal\system\Controller\SystemController::systemAdminMenuBlockPage'
     _title: 'Data-Dictionary'
   requirements:
-    _permission: 'administer site configuration'
+    _permission: 'create data content'
 
 metastore.data_dictionary.settings:
   path: '/admin/dkan/data-dictionary/settings'


janette (commented):
@clayliddell You want data publishers to have access to the settings as well correct?

My test role required the following permissions:
- create data content
- access admin theme
- Access the Content overview page
- View the administration theme
- DKAN publishing workflow: Use Publish transition.
- View published content
THEN I can create a data-dictionary

BUT I still do not have access to the Data-Dictionary Settings screen because of 
https://github.com/GetDKAN/dkan/blob/2.x/modules/metastore/metastore.routing.yml#L176

modules/metastore/metastore.routing.yml
```diff
@@ -165,7 +165,7 @@ metastore.data_dictionary:
     _controller: '\Drupal\system\Controller\SystemController::systemAdminMenuBlockPage'
     _title: 'Data-Dictionary'
   requirements:
-    _permission: 'administer site configuration'
+    _permission: 'create data content'
 
 metastore.data_dictionary.settings:
   path: '/admin/dkan/data-dictionary/settings'
```
janette, line 171: ~~This route is in both metastore.routing.yml and datastore.routing.yml~~
The route in datastore.routing.yml has the 'administer site configuration'
AND the menu link to the dd settings in datasatore.routing.yml also has the 'administer site configuration' permission requirement.

janette: So guessing the more global DKAN administration permission will be a different effort.... and this PR is just to get the DD links available?

clayliddell: Yes, that is exactly right. This was just to get the DD links available.

janette (approved):

PR #3820, opened 2022-08-22
Add facility for creating datastore column indexes

clayliddell:
fixes WCMS-10795

This PR adds support for indexes to AlterTableQuery services.

- [x] Test coverage exists
- [ ] Documentation exists

Documentation should be added for indexing whenever the data-dictionary schema is updated.

## QA Steps

- [ ] Verify that tests are passing.


modules/datastore/src/DataDictionary/AlterTableQuery/MySQLQuery.php
```diff
@@ -29,41 +27,19 @@ class MySQLQuery implements AlterTableQueryInterface {
...
-    $this->datastoreTable = $this->connection->escapeTable($datastore_table);
-    $this->dictionaryFields = $dictionary_fields;
-  }
-
   /**
    * {@inheritdoc}
    */
-  public function applyDataTypes(): void {
+  public function execute(): void {
     $this->dictionaryFields = $this->mergeDatastoreFields($this->dictionaryFields, $this->datastoreTable);
```
clayliddell: This will need be performed on the indexes as well to prevent errors related to an index field not existing from occurring.

modules/datastore/tests/src/Functional/DictionaryEnforcerTest.php
```diff
@@ -96,8 +98,10 @@ public function setUp(): void {
     $this->webServiceApi = ImportController::create(\Drupal::getContainer());
     $this->datasetStorage = \Drupal::service('dkan.metastore.storage')->getInstance('dataset');
     // Copy resource file to uploads directory.
-    $upload_path = \Drupal::service('file_system')->realpath(self::UPLOAD_LOCATION);
-    \Drupal::service('file_system')->copy(self::TEST_DATA_PATH . self::RESOURCE_FILE, $upload_path);
+    $file_system = \Drupal::service('file_system');
+    $upload_path = $file_system->realpath(self::UPLOAD_LOCATION);
+    $file_system->prepareDirectory($upload_path, FileSystemInterface::CREATE_DIRECTORY);
+    $file_system->copy(self::TEST_DATA_PATH . self::RESOURCE_FILE, $upload_path);
```
clayliddell, line 104: This should fix some of the intermittent test failures we've had related to the uploaded_resources directory not existing before this test was run.

modules/datastore/tests/src/Unit/DataDictionary/AlterTableQuery/MySQLQueryTest.php
```diff
@@ -90,22 +94,26 @@ public function testApplyDataTypes(): void {
         ':date_format' => ''
       ],
     ], $update_query);
-    $this->assertEquals("ALTER TABLE {{$table}} MODIFY COLUMN foo TEXT COMMENT 'Foo', MODIFY COLUMN bar DECIMAL(0, ) COMMENT 'Bar', MODIFY COLUMN baz DATE COMMENT 'Baz';", $query);
+    $this->assertEquals("ALTER TABLE {{$table}} MODIFY COLUMN foo TEXT COMMENT 'Foo', " .
+      "MODIFY COLUMN bar DECIMAL(0, ) COMMENT 'Bar', " .
+      "MODIFY COLUMN baz DATE COMMENT 'Baz', " .
+      "ADD  INDEX index1 (foo 12, bar 6, baz 9), " .
+      "ADD fulltext INDEX index2 (foo 6, baz 3);", $query);
```
clayliddell, line 112: Here's where indexing is being tested.

modules/datastore/src/DataDictionary/AlterTableQuery/Builder.php
```diff
@@ -0,0 +1,218 @@
...
+    return $this;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function addDataDictionary(RootedJsonData $dictionary): self {
+    $this->addFields($dictionary->{'$.data.fields'});
+    // @todo Uncomment once index support has been added to data-dictionaries.
+    // $this->addIndexes($dictionary->{'$.data.indexes'});
```
dafeder: Hmm - isn't "add index support" exactly what this PR does? Why is this still commented out?

modules/datastore/src/DataDictionary/AlterTableQuery/Builder.php
```diff
@@ -0,0 +1,218 @@
...
+   * @param array $field
+   *   Field to validate.
+   *
+   * @return array
+   *   Validated field.
+   */
+  protected function validateField(array $field): array {
+    // Validate required properties.
+    if (empty($field['name']) || empty($field['type'])) {
+      throw new \UnexpectedValueException('"name" and "type" are required properties for data-dictionary fields.');
```
dafeder: Is this necessary? We have a JSON-schema for data dictionaries, so that should take care of this kind of validation.

modules/datastore/src/DataDictionary/AlterTableQuery/Builder.php
```diff
@@ -0,0 +1,218 @@
...
+   *
+   * @return array
+   *   Validated index.
+   */
+  public function validateIndex(array $index): array {
+    // Validate nested properties.
+    foreach ($index['fields'] as $key => $index_field) {
+      // Validate required properties.
+      if (empty($index_field['name'])) {
+        throw new \UnexpectedValueException('"name" is a required property for data-dictionary index fields.');
```
dafeder: Not a big deal but I don't believe this is the correct use of UnexpectedValueException - this seems more like an InvalidArgumentException (or _maybe_ a DomainException).

Still I believe we should not be doing this kind of in-the-weeds validation at all, but rather relying on JSON schema for all of this. That is why we are using JSON so extensively in DKAN. If the Dictionary is a RootedJsonData object that should mean that it we can safely assume it conforms to the dictionary schema.

modules/datastore/src/DataDictionary/AlterTableQuery/Builder.php
```diff
@@ -0,0 +1,218 @@
...
+use Drupal\datastore\DataDictionary\AlterTableQueryBuilderInterface;
+use Drupal\datastore\DataDictionary\AlterTableQueryInterface;
+
+use PDLT\ConverterInterface;
+use RootedData\RootedJsonData;
+
+/**
+ * Alter table query builder.
+ */
+class Builder implements AlterTableQueryBuilderInterface {
```
dafeder: Agree Builder is a better name.

modules/datastore/src/DataDictionary/AlterTableQueryBase.php
```diff
@@ -0,0 +1,84 @@
+<?php
```
dafeder, line 1: Not a huge deal but a bit confused by having the base classes and interfaces in the parent dir? I would expect AlterTableQueryBase to be in the AlterTableQuery namespace.

modules/datastore/src/DataDictionary/AlterTableQuery/Builder.php
```diff
@@ -0,0 +1,218 @@
...
+   */
+  protected array $indexes = [];
+
+  /**
+   * Create an alter table query factory.
+   */
+  public function __construct(
+    DatabaseConnectionFactoryInterface $database_connection_factory,
+    ConverterInterface $date_format_converter,
+    string $query_class
```
dafeder: I'm not sure about this pattern... wouldn't it make more sense to do this through normal dependency injection and type hint it to the `AlterTableQueryInterface`?

modules/datastore/tests/src/Unit/DataDictionary/AlterTableQuery/BuilderTest.php
```diff
@@ -0,0 +1,57 @@
...
+      ->getMock();
+    $converter = (new Chain($this))
+      ->add(ConverterInterface::class)
+      ->getMock();
+
+    $builder = new Builder($connection, $converter, MySQLQuery::class);
+
+    // Test Builder's setConnectionTimeout() returns what's expected.
+    $result = $builder->setConnectionTimeout(1);
+    $this->assertTrue(is_a($result, AlterTableQueryBuilderInterface::class));
```
dafeder, line 69: Should these assertions check that the QueryBuilder did anything? It doesn't really seem to be testing anything other than the return type

modules/datastore/src/DataDictionary/AlterTableQuery/Builder.php
clayliddell: Index support has been added to the AlterTableQuery service, but this is blocked by index support being added to data-dictionaries, which will mean updating the schema of data-dictionaries.

modules/datastore/src/DataDictionary/AlterTableQuery/Builder.php
clayliddell: There's no guarantee that this service will be called with a data-dictionary or data from a data-dictionary. This extra layer of validation ensures the builder is being used properly.

modules/datastore/src/DataDictionary/AlterTableQuery/Builder.php
clayliddell: I would be fine using a InvalidArgumentException, but I do think we benefit from having a layer of validation in the builder, since we can't predict how this will be used in the future.

modules/datastore/src/DataDictionary/AlterTableQuery/Builder.php
clayliddell: Yeah, I think the architecture more closely resembles the GoF builder design pattern which is why chose to change the name: https://refactoring.guru/design-patterns/builder

modules/datastore/src/DataDictionary/AlterTableQueryBase.php
clayliddell: Yeah, I chose to namespace it this way since this is how namespacing is handled in Drupal Core, and there doesn't seem to be consistency throughout DKAN.

modules/datastore/tests/src/Unit/DataDictionary/AlterTableQuery/BuilderTest.php
clayliddell: It's checking that an AlterTableQuery object was built, but no properties or methods are public, except for `->execute()`, which is being tested in `MySQLQueryTest`.

We could potentially create a AlterTableQuery mock object which exposes these protected properties, but at that point were only adding additional test coverage to the constructor, which is already being tested in `MySQLQueryTest`.

modules/datastore/src/DataDictionary/AlterTableQuery/Builder.php
clayliddell: Yeah... this pattern came from the initial factory with the idea of making it generic, but I think we should instead just be using polymorphism here by creating children Builder classes which only build one type of AlterTableQuery object (e.g. `MySQLBuilder`).

dafeder (approved):

PR #3824, opened 2022-08-24
Create or modify a resource manually

dafeder:
It may be helpful for difficult harvest/import workflows, where we have very large remote files that need to be imported and we want to pre-fetch them to the server, to be able to create a resource record _before_ creating or updating a dataset. WIP.


dafeder:
This has been mostly addressed by #4004. Closing.

Issue #3772, opened 2022-03-24
Standardize reference types and logic across schemas


dafeder:
The referencing system contains a lot of implicit rules and types of references, and has business logic spread out between the Referencer and Lifecycle classes. This could be organized much better by defining reference types/classes that contain their own reference and dereferencing logic, and simply mapping these reference types to properties in the schemas. This change would be complementary to what we're starting to sketch out in #3761 for better abstracting/defining the schema system.

An initial diagram of how this would work:

```mermaid
classDiagram

class ReferenceDefinitionInterface {
  <<interface>>
  +create()*
  +property()
  +schemaId()
  +reference(mixed metadata) string
  +dereference(string identifier) mixed
}

class AbstractReferenceDefinition {
  <<abstract>>
  -property
  -schemaId
  +property()
  +schemaId()
}
ReferenceDefinitionInterface --|> AbstractReferenceDefinition

class JsonReference {
  <<service>>
  +reference()
  +dereference()
}
AbstractReferenceDefinition --|> JsonReference


class ResourceReference {
  <<service>>
  +reference()
  +dereference()
}
AbstractReferenceDefinition --|> ResourceReference

class IdReference {
  <<service>>
  +reference()
  +dereference()
}
AbstractReferenceDefinition --|> IdReference

class ReferencerInterface {
  <<interface>>
  +reference(object metadata) object
}

class Referencer {
  <<service>>
  +reference()
  #referenceProperty()
  #referenceMultiple()
}
ReferencerInterface --|> Referencer

class DereferencerInterface {
  <<interface>>
  +dereference(object metadata) object
}

class Dereferencer {
  <<service>>
  +dereference()
  #dereferenceProperty()
  #dereferenceMultiple()
}
DereferencerInterface --|> Dereferencer

class ReferenceMap {
  <<service>>
  +getAllReferences(schemaId) array
  +getReference(schemaId, propertyName) ReferenceDefinitionInterface
}

Referencer ..> ReferenceMap : Dependency
Dereferencer ..> ReferenceMap : Dependency

```

The Referencer will use the ReferenceMap service to find all the reference definitions for a schema. It will then iterate through them and call that reference definition class's `reference` method (e.g. `JsonReference::reference()`). The dereferencer will work essentially the same way, calling the `dereference()` method. So we should be able to send any incoming or outgoing metastore item through the Referencer or Dereferencer without having schema-specific logic in these classes.

[...]

PR #3825, opened 2022-08-24
Refactor metastore and merge with common and dkan modules

### LINKED ISSUE #3736: Metastore code refactor for clearer naming and flow
Proposal to tighten up the naming and typing conventions in the metastore module. Some changes are simply renaming or separating existing classes/methods, but others change the types being passed around. The current code is difficult to understand and maintain. 

This assumes that the metastore module has been merged into the dkan module (see #3746).

## Drupal\dkan\Metastore\Item

Classes and interfaces for defining Metastore item objects. These will be the primary metadata objects passed around the system.

| Class                                                   | Description                                                                  |
|---------------------------------------------------------|------------------------------------------------------------------------------|
| `Drupal\dkan\Metastore\Item\MetastoreItemInterface`       | Interface for a metastore item object                                        |
| `Drupal\dkan\Metastore\Item\MetastoreEntityItemInterface` | Extension interface for entity-based items                                   |
| `Drupal\dkan\Metastore\Item\MetastoreNodeItem`            | Metastore item class based on a node. |

## Drupal\dkan\Metastore\Schema

Classes for defining, retrieving and validating against metastore schemas. Note that `SchemaManager` replaces the old `SchemaRetriever` service. Assumes completion of #3761.

| Class                                                   | Description                                                           |
|---------------------------------------------------------|-----------------------------------------------------------------------|
| `Drupal\dkan\Metastore\Schema\MetadataValidator`        | Utility class for validating metadata against a schema in the system. |
| `Drupal\dkan\Metastore\Schema\MetastoreSchema`          | Schema plugin definition, for schemas that will be defined in YAML.   |
| `Drupal\dkan\Metastore\Schema\MetastoreSchemaInterface` | Interface for schema plugins.                                         |
| `Drupal\dkan\Metastore\Schema\SchemaManager`            | Plugin manager to gather definitions and retrieve schemas.     |

### LINKED ISSUE #3746: Reorganize and simplify module names and dependencies
Need to determine what the process and sequence for rolling out this work

For reasons that are difficult to remember today, we made a decision early in DKAN 2 development to make our module machine names extremely simple -- "metastore," "datastore," etc, without prefixing them all with “dkan_.” We came to regret this fairly quickly. We also have some confusing division of central DKAN functionality between “DKAN,” “common” and “metastore.”

Given what has coalesced as the core functionality in DKAN, we should in version 3 take the opportunity to optimize our naming conventions and separations of concerns. Current recommendation is: 

## DKAN (dkan)

This will for the most part consolidate all code and functionality spread across “dkan,” “common,” and “metastore.” Because there is no DKAN without the metastore, and all additional modules interact with the metastore in one way or another but not necessarily with each other, it makes more sense to consider all of this part of the main dkan module.

The overall Metastore system -- the majority of the code in _modules/metastore/src_ -- will now be considered the core functionality of DKAN. So, `Drupal\metastore` would become the `Drupal\dkan namespace`. Our service and route names could, for the most part, remain unchanged. The search module would now be dkan_search.

The DKAN module should not depend on any additional submodules; we would still have a working data catalog with API if the dkan module is enabled and nothing else is. We would perhaps add another parallel functional test job to the CI that tests the system without datastore, harvest or other additional modules enabled to make sure it works.

There is a class currently in common -- `DatasetInfo` -- that breaks a lot of separation, because it relies on all the other modules to bring together the dataset info from around the system. This will need to implement a plugin system to allow other modules to add information if enabled.

## DKAN Datastore (dkan_datastore)

The `Drupal\datastore` namespace would now be `Drupal\dkan_datastore`. Little here would change, we would simply create more isolation so that it could be disabled.

## DKAN Harvest (dkan_harvest)

This would follow the same pattern as dkan_datastore, mostly remaining the same but possibly with some refactoring of certain functionality to allow the core DKAN Metastore to be enabled without it. We may want to let the Harvest module contribute information to `DatastoreInfo`.

### LINKED ISSUE #3761: Implement New Schema management system
The way schemas are defined and discovered in DKAN is still quite fragile, and not conducive to customization. Flexibility is the major value proposition of DKAN's unorthodox approach to metadata storage in DKAN, so this is technical debt that would be very useful to pay off, in terms of stability, developer experience and ease of adoption.

The current schema system infers a set of schemas by simply looking at any JSON files that exist in one of a few possible filesystem locations. There is no way to differentiate between core, required schemas and user-defined schemas -- if the user has a custom schema directory all core schemas will be ignored. As well, there is an assumption that schema filenames will match their machine names in the system and the property name they are referenced from, and that there will necessarily be a dataset and distribution schema with a few expected properties. Finally, there is no differentiation between the actual metastore schemas and the form-specific ui schemas, but rather we again relay on filename-based conditionals to chose one or the other when appropriate.

After thinking through a number of approaches I'm proposing what I believe is the best solution to this problem. It is less of a break from the current paradigm than other ideas in this direction (implementing schemas as PHP classes, or as Drupal entity bundles), has what I think is a better separation of concerns between PHP code, declarative definitions in YAML, and schema documents in JSON. It gives us a unified way to abstract the non-negotiable parts of schemas in the DKAN business logic and allow everything else to be fully override-able. This solves some inconsistencies in the referencing and storing logic between datasets, refrenced items like distributions and keywords, and "resources" (files).

This is accomplished through a standardized set of reference _types_ and schema _classes_. The latter are equivalent to what we have been calling schema "behaviors" in previous planning documents. They allow us to create universal patterns in our DKAN methods to replace some very brittle conditionals that will begin to break down as soon as anyone significantly alters the default DCAT-US based metastore schemas.

## Overview

A new schema system would define schemas in .yaml files in a module's root directory, similar to Drupal services and routes. No configuration is guessed/inferred simply from the presence or lack of certain files in the filesystem. The intention is for schemas to be tightly coupled to specific, focused modules. 

The current admin pages where reference and trigger fields are set will be removed, as these are all covered in the new schemas file.

This will of course be a significant upgrade of DKAN and will involve these changes:

1. We will move the core default collection of schemas from the root dkan/schema folder into a new module called dkan_dcat_us. Sites using the default schema will need to enable this module, which we can do via hook_update. The module will be optional and only a dependency of the sample_content module.
2. Existing sites that have custom schema in their docroot/schema folder will need to create a new module, following the template of dkan_dcat_us, and be sure to have this module enabled and dkan_dcat_us disabled.
3. Much of what's described here can be refactored incrementally to take advantage of the new abstractions, but the minimum updates for release would be:
  * Updating the SchemaRetriever class to use the new yaml files, and provide new public methods to provide UI Schemas, reference and trigger information, but keeping it backwards compatible for the time being so that new functionality can be adopted gradually.
  * Updating the Referencer classes to use the new reference definitions from the YAML in place of property-name-based conditionals. This will hopefully not change any public method signatures and only affect internal logic and private methods.
  * Updating the datastore event subscriber to pull its list of triggering fields from the schemas YAML instead of the config.

## Structure of a DKAN schemas file

Similar to Drupal's _.services.yml_ files, a _.schemas.yml_ file in a module's root directory will register schemas with DKAN and make them available to `Drupal\metastore\SchemaReriever`. The default DCAT-AP schemas can be described like this:

```yaml
catalog:
  schema_path: schema/catalog.json
  references:
    - { property: dataset, schema: dataset }
  class: catalog
  endpoint_base: data

dataset:
  schema_path: schema/dataset.json
  ui_schema_path: schema/dataset.ui.json
  identifier: { property: identifier, type: uuid }
  references:
    - { property: distribution, schema: distribution }
    - { property: publisher, schema: organization }
    - { property: keyword, schema: keyword }
    - { property: theme, schema: theme }
  triggers:
    - { property: modified, trigger: datastore_import }
  class: dataset

distribution:
  schema_path: schema/distribution.json
  references:
    - { property: downloadURL, type: resource }
    - { property: describedBy, schema: data-dictionary, type: id }
  class: distribution

organization:
  schema_path: schema/organization.json
  ui_schema_path: schema/organization.ui.json

keyword:
  schema: { type: string }
  class: literal

theme:
  schema: { type: string }
  class: literal
```

## Schema definitions

* **class**: Schema class, to map an arbitrary schema to core DKAN functionality. Currently available classes are `catalog`, `dataset`, `distribution`, `dictionary`, and `literal`.
* **identifier**: Object defining an arbitrary identifier property in the JSON, used to inject item ID into JSON body on creation.
  * **property**: The JSON property. For now, only top-level properties supported.
  * **type**: Format of identifier allowed. For now, only `uuid` supported which injects the internal item uuid as a value. Other types may be made available in the future.
* **endpoint_base**: Pattern to use for the endpoint to this type of metastore item. For instance, if set to "data", the catalog JSON endpoint will be _/data.json_. Currently only supported for `catalog` class.
* **references**: Object to tell the DKAN metastore which properties should have their values stored in referenced items rather directly in the item's JSON.
  * **type**: Type of reference. Accepts one of the following values:
    * **schema**: Swap JSON object or string literal with corresponding item UUID.
    * **id**: Swap a URL to a metastore item with a simple item UUID (maintains portability).
    * **resource**: Swap a URL to a DKAN resource ID
    * ~**file**~: _Possible future reference type if we replace the resource system with DKAN's core file entities, which has been proposed a few times recently._
  * **schema**: The schema to use for the referenced item. Not needed if type is `file`.
* **schema**: Full JSON schema, expressed in YAML.
* **schema_path**: Path to JSON Schema file, relative to the schemas file.
* **triggers**: An array of object with properties corresponding to "trigger" names. The trigger will run if the property value changes on item update.
  * **property** The JSON property. For now, only top-level properties supported.
  * **trigger**: Currently only `datastore_import`, which will re-import the datastores associated with this item, is available.
* **ui_schema**: Full UI schema, expressed in YAML.
* **ui_schema_path**: Path to the UI schema for use by form builder.
  * **property**: The property name. Only top-level properties supported at the moment.

## Classes

Schema classes are not references to PHP classes, but map roughly to classes from DCAT and related RDF vocabularies. DKAN can work with essentially any metadata schema as long as it maps to these basic DCAT concepts.

1. `catalog`: Schema used for custom catalog endpoint, usually data.json. Equivalent to `dcat:Catalog`. There may be only one catalog record in the 
5. `dataset`: The main dataset schema in the metastore. Equivalent to `dcat:Dataset`
6. `distribution`: A specific representation of the dataset. Usually, a file. Equivalent to `dcat:Distribution`.
7. `dictionary`: A schema with column-level metadata, such as a table schema or a shared data dictionary. For now, implemented in core and there is no support for standards other than the Frictionless Table Schema. Closest equivalent would be `dct:Standard`, though in DKAN this class is much more specific.
8. `literal`: A "schema" whose items are unstructured values, such as strings. Literals will be wrapped in a JSON-LD structure. For instance, a keyword "health" would be stored as a simple string in the database, but retrieved from the API with the JSON `{"@value": "health"}`. Equivalent to `rdfs:Literal` and similar RDF literal types.

Schemas may map to other DCAT classes (for instance, `foaf:Organization` in the case of the default `organization` schema) but the play no special role in DKAN's architecture and do not need to be assigned a class in the schemas file.

For the moment, there can only be one schema defined for any class other than `literal`. In future iterations we may want to make this more flexible to address use-cases where multiple types of datasets or distributions are supported -- for example, a catalog might allow different schemas for geospatial or financial metadata.

Typical class relationships in a DKAN catalog:

```mermaid
classDiagram
    class catalog {
        array~dataset~ dataset
    }
    class dataset {
        array~distribution~ distribution
        literal keyword
        literal theme
    }
    class distribution {
        dictionary describedBy
    }

    class literal {
        string value
    }

    class dictionary {
        array fields
    }
    
    catalog "1" o-- "many" dataset
    dataset "1" o-- "many" distribution
    dataset "many" o-- "many" literal
    distribution "1" o-- "1" dictionary
```

## Related changes

Specing out this work has provided some clarity on we could better handle identifiers and references in the API, will document this in a second issue.

dafeder:
Fixes #3746 
Fixes #3761 
Fixes #3736 

Very much WIP, started experimenting with some of the changes that we're thinking about for a 3.x release, and was hard to work on any one in isolation. Also addresses some things in #3772 but that issue is already in progress via #3774. Will need to resolve all these things before this branch is usable.


dafeder:
Good start but too stale to be useful at this point.

PR #3834, opened 2022-09-07
Replace empty date values with `NULL` to prevent date format error on data-dictionary application

clayliddell:
fixes WCMS-11331

This PR seeks to fix problems related to data-dictionaries not being applied on Acquia DKAN sites due to a recurring MySQL error:
```
 [error]  SQLSTATE[HY000]: General error: 1411 Incorrect datetime value: '' for function str_to_date: UPDATE "datastore_ee56c165233467fa76024afe3552fad2" SET "effective_date"=STR_TO_DATE(effective_date, :date_format); Array
(
    [:date_format] => %m/%d/%Y
)
```

Since the error is related to empty strings not matching the provided date format, empty strings in date fields are being replaced with `NULL` before dates are re-formatted.

- [x] Test coverage exists
- [X] Documentation exists

## QA Steps

- [ ] Create a data-dictionary with the column "b" of type "date", and with a format of "%m/%d/%Y".
- [ ] Configure the created data-dictionary as the global data-dictionary on the test environment.
- [ ] Create a dataset using https://gist.githubusercontent.com/clayliddell/d4ab207343964a43a7f8c9feae6a790e/raw/3060b38b62d81bd226a80a2decc0a060143d56f1/test.csv as a remote file.
- [ ] Run cron; wait for it to import and for the data-types to be applied.
- [ ] Confirm your able to search on the "b" column on the filtered resources page as though it's a date field.


modules/datastore/src/DataDictionary/AlterTableQuery/MySQLQuery.php
```diff
@@ -61,7 +61,12 @@ public function applyDataTypes(): void {
 
     // Build and execute SQL commands to prepare for table alter.
     $pre_alter_commands = $this->buildPreAlterCommands($this->dictionaryFields, $this->datastoreTable);
+    $this->connection->query('SET SESSION innodb_strict_mode=OFF');
+    $sql_mode = $this->connection->query('SELECT @@sql_mode')->fetchField();
+    $strict_mode = $this->connection->query('SELECT @@innodb_strict_mode')->fetchField();
+    \Drupal::logger('datastore')->notice('b ' . $sql_mode . ' ' . $strict_mode);
     array_map(fn ($cmd) => $cmd->execute(), $pre_alter_commands);
```
clayliddell:
```suggestion
    array_map(function ($cmd) {
      try {
        $cmd->execute();
      }
      catch (\Throwable $e) {
        // do nothing.
      }
    }, $pre_alter_commands);
```

modules/common/src/Storage/DatabaseConnectionFactory.php
```diff
@@ -55,11 +55,21 @@ protected function buildConnectionInfo(): array {
...
+  /**
+   * Prepare database connection.
+   *
+   * @param \Drupal\Core\Database\Connection $connection
+   *   Database connection object.
+   */
+  protected function prepareConnection(Connection $connection): void {
+    $this->doSetConnectionTimeout($connection);
+  }
+
```
clayliddell, line 72: I think this is just cleaner in terms of implementation/extensibility.

modules/datastore/src/DataDictionary/AlterTableQuery/MySQLQuery.php
```diff
@@ -192,6 +192,9 @@ protected function buildPreAlterCommands(array $dict, string $table): array {
       // the format of the date fields to ISO-8601 before importing into MySQL.
       if ($type === 'date' && !empty($format) && $format !== 'default') {
         $mysql_date_format = $this->dateFormatConverter->convert($format);
+        // Replace empty date strings with NULL to prevent STR_TO_DATE errors.
+        $pre_alter_cmds[] = $this->connection->update($table)->condition($col, '')->expression($col, 'NULL');
+        // Convert date formats for date column.
```
clayliddell, line 198: This is the only behavioral change in this PR.

janette (approved):

PR #3837, opened 2022-09-19
WCMS-11097: Migrate Circle-CI from DKTL to DDEV

paul-m:
fixes WCMS-11097

- [x] Test coverage exists
- [x] Documentation exists

## QA Steps


.circleci/old.config.yml
```diff
@@ -0,0 +1,166 @@
+version: 2.1
```
clayliddell: I think we're safe to get rid of this file since we'll still be able to reference the old version via git.

clayliddell (dismissed):
The code in this PR looks good, but it looks like a couple of the PHPUnit upgrade tests are failing. We'll need to resolve those failures in this PR or another before merge, since we don't want a failing pipeline on the main branch.

@paul-m I realize that the changes in this PR are not directly causing the failures and you may not know how to track down the failures in DKAN. I'm currently busy working on a couple of other tickets this sprint, but if I have the bandwidth, I can try to circle back around to investigate this with you. In the meantime, if @dafeder has time this sprint, I think he may have more insight on the specific tests which are failing.

.circleci/old.config.yml
paul-m: Removed.

paul-m:
The issue with failing tests was that we're using Drupal Test Traits, which test in-place, combined with different test methods defining datasets with the same ID. The solution is to hash the function name and a number for the IDs.

paul-m:
Just to point out that Cypress tests are now passing because I marked `cypress/integration/11_admin_dataset_file_upload.spec.js` as skipped. It consistently fails for reasons I can't fathom.

paul-m:
Reverted the tests which seem to be passing. Moving the issue into review.

dafeder (approved):

Issue #3839, opened 2022-09-22
Unable to remove a harvest without deleting its datasets


thierrydallacroce:
With the goal of cleaning up an old harvest from an environment where it was mistakenly added and ran, I tested both of the following approaches in a local environment:
```
drush dkan:harvest:deregister harvest_id 
```
and
```
drush dkan:harvest:revert harvest_id
```
But, both deleted the datasets.

stefan-korn:
@dafeder : Provided a merge request for this as discussed in #3840.

One could probably add some check if there are entities related to the harvest at all and only in that case show the warning and ask for confirmation. But maybe the warning is okay in any case?

There are also some issues with the messages from the deregister comment, which are unrelated to this. See issue #3956. #3840 and #3956 will not apply together. If either of the PRs is accepted, the other will need to be updated.

dafeder:
@stefan-korn I think we consider this resolved, feel free to re-open if you think it needs more work.

PR #3840, opened 2022-09-22
Allow deleting a harvest without deleting its datasets

thierrydallacroce:
Proposed fix for [#3839]

### Steps to test

1. Create and run a harvest
2. Verify the harvest is registered, i.e. shown in the output of `drush dkan:harvest:list`
3. Verify its datasets exist and are published
4. Deregister the harvest: `drush dkan:harvest:deregister HARVEST_ID  --revert`
5. Verify the harvest is no longer shown in the output of `drush dkan:harvest:list`
6. Verify its datasets still exist, and are published


thierrydallacroce:
@dafeder, I re-tested this locally to be sure, spinning up both of my environments from scratch with their minimal content.

In one of them, `drush dkan:harvest:deregister test_minimal__data` removed the harvest but kept the datasets, while in the other `drush dkan:harvest:deregister test_minimal__data --revert` removed both the harvest and its datasets.

dafeder (dismissed):

thierrydallacroce:
@dafeder made me aware of a use case which invalidates this approach and PR.

stefan-korn:
@dafeder @thierrydallacroce : Could you explain why there is a use case which invalidates this approach and PR?

Is there actually a possibility to delete a harvest without deleting its datasets?

dafeder:
@stefan-korn uf, it's been a while and I wish we'd documented the details of this discussion in the thread! My memory here is that the PR went a little too far and removed the revert command, making revert only possible as part of a deregister. There are legitimate use-cases for both deregistering w/o revert, and reverting w/o deregister. I think we were going to just redo this to remove the revert piece from deregister completely, and add a warning if you deregister a harvest with published datasets that deregistering will make it impossible to bulk-revert the datasets connected to it (maybe actually displaying a list of them?). But we must have gotten distracted before making a new PR.

@janette or @jastraat do you know any other details about this? If my memory is more or less correct we would accept a PR from you along those lines @stefan-korn or might have time to redo it on our end soon (can't guarantee a timeline though).

stefan-korn:
@dafeder : Okay, removing the revert from the deregister command seems feasible to me. Will try to provide a PR for that. 

Should this issue be opened again?

dafeder:
The issue (#3839) is still open

PR #3846, opened 2022-10-07
Swap out dkan-tools with ddev-addon.

janette:
fixes [org/repo/issue#]

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps


composer.json
```diff
@@ -5,7 +5,7 @@
     "minimum-stability": "dev",
     "description": "DKAN Open Data Catalog",
     "require": {
-        "drupal/admin_toolbar": "^2.3",
+        "drupal/admin_toolbar": "^3.2",
```
jastraat: I think this is now ^3 in the composer file, so maybe this change should be removed from the PR?

jastraat (approved):
If we're moving to DDEV officially from DKAN Tools, this looks good!

PR #3865, opened 2022-11-03
Add specific permission to administer data-dictionary settings

tiffneybare:
Some sites may not want users who manage data publishing to have the very broad 'Administer site configuration' permission in order to access the data dictionary settings screen.

## QA Steps

- [ ] Add the new _administer data dictionary settings_ permission to an existing role
- [ ] Create a user with that role
- [ ] Log in as the new user
- [ ] Confirm you have access to the following page: /admin/dkan/data-dictionary/settings

### DIFF
--- a/modules/metastore/metastore.permissions.yml
+++ b/modules/metastore/metastore.permissions.yml
@@ -1,3 +1,8 @@
 'post put delete datasets through the api':
   title: 'post, put, and delete datasets through the api'
   description: 'post, put, and delete datasets through the api'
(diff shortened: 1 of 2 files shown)


janette:
@tiffneybare I think it would be better to modify the role permissions on the project as we don't want this exposed to roles that just need access to edit content.

janette (approved):
Adding "Metastore" to the permission titles to follow the pattern set by the Datastore and Harvest modules.

PR #3866, opened 2022-11-03
schema for config

paul-m:
fixes [org/repo/issue#]

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps


modules/metastore/config/install/metastore.settings.yml
```diff
@@ -1,21 +1,21 @@
...
+  'accrualPeriodicity': '0'
+  'describedBy': '0'
+  'describedByType': '0'
+  'issued': '0'
+  'license': '0'
+  'modified': '0'
+  'references': '0'
+  'spatial': '0'
+  'temporal': '0'
+  'isPartOf': '0'
```
janette, line 21: Why all the single quotes? Only @type needs them.

paul-m: They're all strings. https://github.com/GetDKAN/dkan/blob/2.x/modules/metastore/config/schema/metastore.schema.yml#L15 If we need to change them to be something else then we can.

paul-m:
Unsure what the context here is. Closing.

PR #3880, opened 2022-11-21
Add tsv support to datstore_mysql_import

janette:
First draft, ideally we want to allow user to define parser config: delimiter, eol, escape, enclosed.


tiffneybare (approved):
This looks good locally

PR #3881, opened 2022-11-23
Switch to simple built-in FileFetcher on import

dafeder:
The external [FileFetcher library](https://github.com/GetDKAN/file-fetcher) makes file downloads slower, and has led to failures/corruptions of data. Replacing it with a single FileFetcher job that copies a file all at once with Guzzle/cURL will lead to faster, more stable imports of large files. Our experience has been that downloading the file in chunks leads to more problems than it solves, at least as it works now.

The new FileFetcherJob class is all reused code from the [FileFetcher library](https://github.com/GetDKAN/file-fetcher) - specifically, the "LastResort" processor. Under this new paradigm the idea of a "processor" is no longer relevant. It's possible that in the future we would want to re-introduce the concept if our one-size solution does not actually fit all.

Local testing showed that a large remote CSV file that took 65 minutes to import completed in 20 minutes after making this change.

## QA Steps

- Create datasets with both remote and local file URLs in downloadURL
- Run import and confirm file is successfully imported


modules/datastore/src/Plugin/QueueWorker/FileFetcherJob.php
```diff
@@ -0,0 +1,166 @@
...
+    if (!isset($config['filePath'])) {
+      throw new \Exception("Constructor missing expected config filePath.");
+    }
+
+    $state = [
+      'source' => $config['filePath'],
+      'total_bytes' => 0,
+      'total_bytes_copied' => 0,
+      'temporary' => FALSE,
+      'destination' => $config['filePath'],
```
janette: Remove destination line, it gets set in the setupState method

janette (approved):

PR #3890, opened 2022-12-15
Restore innodb_strict_mode toggle for create table

dafeder:
## DO NOT DELETE: Currently in use on QIC

This should restore the innodb_strict_mode toggle (previously added in #3784 and removed in #3887). Something had changed in the code workflow and it was no longer working as originally done.

## QA Steps

- Create a dataset using a table with 250 or more columns (try [this one](https://download.cms.gov/openpayments/PGYR21_P063022/OP_DTL_RSRCH_PGYR2021_P06302022.csv))
- Run import, ensure it imports with no errors


paul-m:
Work happening here: https://github.com/GetDKAN/dkan/pull/3976

paul-m:
https://github.com/GetDKAN/dkan/pull/3976 changed the behavior so that datastore_mysql_import always turns off innodb_strict_mode when creating tables.

Leaving this open because we might wish to become more granular in how we turn that setting off or on.

janette:
closing in favor of #3976

PR #3905, opened 2023-02-06
Index the current, published dataset revision

jastraat:
Currently, search api indexes the most recent revision of a dataset - whether or not that revision is published. To recreate this problem:

- Have a local DKAN installation with the default sample content that is fully search indexed.
- Search datasets for "Florida". Should see one result: "Florida Bike Lanes".
- Create a new revision of the "Florida Bike Lanes" dataset, replacing every instance of "Florida" with "Missouri". Save this revision as a DRAFT
- Re-index the site. 
- Search datasets for "Florida". No results.
- Search datasets for "Missouri". Get the Florida result even though Missouri does not appear anywhere in the published revision.

- [ ] Test coverage exists
- [ ] Documentation exists

Note: looks like the functional tests need to be updated.

## QA Steps

- Checkout the PR branch into a vanilla DKAN site with the sample content.
- Create a new revision of the "Florida Bike Lanes" dataset, replacing every instance of "Florida" with "Missouri". Save this revision as a DRAFT
- Re-index search api. Make sure to fully clear the index before reindexing and then clear the drush cache.
- Search datasets for "Florida". => Get one result
- Search datasets for "Missouri". => No results

### DIFF
--- a/modules/metastore/modules/metastore_search/src/Plugin/search_api/datasource/DkanDataset.php
+++ b/modules/metastore/modules/metastore_search/src/Plugin/search_api/datasource/DkanDataset.php
@@ -75,7 +75,7 @@ class DkanDataset extends DatasourcePluginBase {
 
     foreach (array_combine($ids, $ids) as $id) {
       try {
-        // We only want published revisions to be indexed.
+        // Only index published revisions.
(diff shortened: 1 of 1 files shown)


janette (approved):
Nice

Issue #3898, opened 2023-01-05
PHP 8.1 fails tests because of Guzzle


paul-m:
Search existing issues to make sure the bug has not been reported already.

## Describe the bug

This is an upstream issue with Guzzle and Drupal.

Guzzle:
- https://github.com/guzzle/guzzle/issues/3020
- https://github.com/guzzle/guzzle/issues/2971

Drupal:
- https://www.drupal.org/project/drupal/issues/3281667

DKAN 2.x branch as of commit 15a2ed0c6d0b4e5f7c94f92c72f99fc6f0d878b0.

Using DDEV to run under PHP 8.1, I see this error:
```
% ddev dkan-test-phpunit

[...]

Starting PHPUnit test run.

[..]

Runtime:       PHP 8.1.13
Configuration: /var/www/html/dkan/phpunit.xml

PHP Fatal error:  During inheritance of Countable: Uncaught Return type of GuzzleHttp\Cookie\CookieJar::count() should either be compatible with Countable::count(): int, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice

/var/www/html/vendor/guzzlehttp/guzzle/src/Cookie/CookieJar.php:220
/var/www/html/vendor/symfony/phpunit-bridge/DeprecationErrorHandler.php:126
/var/www/html/vendor/guzzlehttp/guzzle/src/Cookie/CookieJar.php:10
/var/www/html/vendor/fabpot/goutte/Goutte/Client.php:135
/var/www/html/vendor/symfony/browser-kit/Client.php:404
/var/www/html/vendor/friends-of-behat/mink-browserkit-driver/src/BrowserKitDriver.php:145
/var/www/html/vendor/behat/mink/src/Session.php:148
/var/www/html/vendor/weitzman/drupal-test-traits/src/GoutteTrait.php:121
/var/www/html/vendor/weitzman/drupal-test-traits/src/GoutteTrait.php:87
/var/www/html/vendor/weitzman/drupal-test-traits/src/ExistingSiteBase.php:56
/var/www/html/vendor/phpunit/phpunit/src/Framework/TestResult.php:730
/var/www/html/vendor/phpunit/phpunit/src/Framework/TestSuite.php:665
/var/www/html/vendor/phpunit/phpunit/src/Framework/TestSuite.php:665
/var/www/html/vendor/phpunit/phpunit/src/Framework/TestSuite.php:665
[...]

PR #3908, opened 2023-02-07
PHP 8.1 compatibility changes

jastraat:
Related issue: https://github.com/GetDKAN/dkan/issues/3898

This updates weitzman/drupal-test-traits. the drupal core recommended package still requires guzzle version 6 as do a handful of other 3rd party packages listed in the issue above, so this PR does not provide full compatibility with PHP 8.1, but once the core recommended package is updated (D10?) along with the other packages, DKAN should be good to go with PHP 8.1.

As mentioned in the above issue, I was able to get all automated tests to run with PHP 8.1 by manually updating the 3rd party libraries in my local installation. I also ran all automated tests with PHP 7.4 to make sure that these changes did not break sites using an older version of PHP.

Note that the auto_detect_line_endings is no longer supported in PHP 8.1, so this change removes that.

In addition, the Drupal core recommended package still requires guzzlehttp version 6.

Related packages that also require updates:

- https://github.com/fmizzell/maquina - Maquina\StateMachine\Machine::jsonSerialize()
- https://github.com/m1x0n/opis-error-presenter - PresentedValidationError::jsonSerialize()
- https://github.com/GetDKAN/csv-parser - CsvParser\Parser\Csv::jsonSerialize()
- https://github.com/GetDKAN/procrastinator - Result::jsonSerialize() and Job::jsonSerialize()

- [ ] Test coverage exists
- [ ] Documentation exists


jastraat:
Drupal itself is not using the code sniffer checks disabled by this PR. They are excluded here: 

- https://git.drupalcode.org/project/drupal/-/blob/10.1.x/core/phpcs.xml.dist#L69
- https://git.drupalcode.org/project/drupal/-/blob/10.1.x/core/phpcs.xml.dist#L79

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
```diff
@@ -89,15 +89,9 @@ protected function runIt() {
    */
   protected function getColsFromFile(string $file_path, string $delimiter): array {
 
-    // Ensure the "auto_detect_line_endings" ini setting is enabled before
-    // openning the file to ensure Mac style EOL characters are detected.
-    $old_ini = ini_set('auto_detect_line_endings', '1');
```
paul-m, line 94: Maybe make a note in the issue somewhere that this changes compatibility with \r line endings. https://php.watch/versions/8.1/auto_detect_line_endings-ini-deprecated

paul-m (changes requested):
Generally LGTM but let's be sure and have a release note that says we don't support files with \r line endings any more.

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
jastraat: Thanks @paul-m ! I don't see a change log file. Where would I update the release notes as part of this PR?

jastraat: Janette added a mention in the draft release notes for the next DKAN release here: https://github.com/GetDKAN/dkan/releases

janette (approved):
Adding won't fix exceptions to skip the warnings on #[\ReturnTypeWillChange] comments required by PHP 8.1

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
janette: Will document in the release notes.

paul-m: Confirmed @janette drafted a release change. woot!

paul-m (approved):

maietta:
Status?

PR #3253, opened 2020-10-30
Make DKAN work without decoupled frontend

dafeder:
fixes [org/repo/issue#]

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

- [ ] Add manual QA steps in checklist format for a reviewer to perform to confirm that the feature or fix is working. Include as much details as possible so that the reviewer doesn't lose time figuring out how to perform steps.

PR #3914, opened 2023-03-02
Add twig templates

janette:
duplicate of #3253


dgading (approved):
Looks good, works like it did on the original branch 🥇

PR #3915, opened 2023-03-06
Update version dependency for rooted-json-data

jastraat:
Earlier 0.0.2 tags were not correctly interpreted by composer.


paul-m (approved):
Passes CI, only changes one composer constraint to support https://github.com/GetDKAN/RootedJsonData/pull/14 everybody happy. :-)

Issue #3651, opened 2021-09-01
Process for identifying datastores that need column indexes


grugnog:
## User Story

Datastore queries (at least common ones that have previously been slow) are fast and efficient on server resources

## Acceptance Criteria

- Datastore indexes are added (and re-added if dataset re-uploads occur)
- Datastore indexes account for real life usage (since it is impractical to anticipate every possible API query)

grugnog:
Working is ongoing at https://git.drupalcode.org/project/indexer - I have code to parse slow query log data and identify compound indexes with the most value (most rows read in slow queries and highest cardinality) as well as identify existing indexes (so we don't duplicate those). Next I need to parse the tree and actually add or remove indexes as needed.

grugnog:
So the main task of adding/removing indexes based on this analysis is complete. The remaining task is to (optionally) store these in config and automatically add them to new tables that match provided column name combinations.

dafeder:
This never really got the attention it deserved - let's circle back on it when we have a chance.

PR #3916, opened 2023-03-06
Remove dependency on indexer module

jastraat:
The indexer module is only installed in a single OD site, and it is never installed with vanilla DKAN. It is not currently D10 compatible and we're removing it as a dependency in preparation for D10 compatibility.


dafeder:
The [indexer module](https://www.drupal.org/project/indexer) is a great idea and could provide a lot of value, but it's been under-utilized and under-documented so far. Hope to bring it back soon. See #3651 for reference.

dafeder (approved):

PR #3917, opened 2023-03-07
Remove Jenkinsfile

dafeder:
We have kind of stopped using the QA sites from Jenkins and they are a bit of burden to maintain. Removing for now.

There should be no Jenkins check if the PR is successful.


jastraat (approved):
Based on passing tests and the conversation in Slack indicating this is not in use, LGTM!

PR #3918, opened 2023-03-10
Remove dredd

### LINKED ISSUE #3536: Switch from Dredd tests to PHPUnit-based contract testing
Dredd tests are no longer feasible for our complete OpenAPI 3 spec. We should switch to a more flexible framework for contract testing that supports v3. Most convenient would probably be something that integrates with PHPUnit. Current favorite option would be something build on [openapi-validator](https://gitlab.com/mmalawski/openapi-validator).

Read more about that library [here](https://medium.com/@mieszkomalawski/contract-testing-rest-apis-with-php-and-openapi-18651460840f).

If we can implement this completely, it may have the added benefit of being able to replace all or most of our existing functional tests.

dafeder:
fixes #3536 

We've already started using phpunit for basic schema validation stuff, dredd has been disabled for a while now.


paul-m (approved):
There's no dredd in the CI at this point, so removing the config makes sense.

PR #3924, opened 2023-03-23
Convert CSV chunks to utf-8 and strip BOM if present

jastraat:
This removes the dependency on fylax/forceutf8. It uses a combination of the core Unicode functions and mbstring as a replacement. It also strips the BOM at the beginning of the file if present.

Quick mention that Unicode::convertToUtf8 does not work with PDC files, so while this may be passing tests, it would break the PDC datastore imports.

It does still need working tests.

Related ticket: https://jira.cms.gov/browse/WCMS-14692

- [ ] Test coverage exists
- [ ] Documentation exists


jastraat:
We're going a different direction.

PR #3930, opened 2023-03-30
Refactor hydration pattern in DataResource class

paul-m:
The DataResource class contains a bunch of unnecessary complexity to build objects from database records.

We should build the objects directly from the rows rather than encoding them to JSON and hydrating.

Work here will be limited to removing the hydration pattern from DKAN-scoped classes. This pattern is: serializing objects in order to pass them to other classes or methods only to have them re-hydrated. Where possible, a factory pattern will be implemented to take data without being serialized.

Changes here:
- Modify `Drupal\common\DataResource` to add a DB record factory method, called `createFromRecord()`.
- Modify `Drupal\metastore\ResourceMapper` to use the new `DataResource` factory method, rather than a hydration round-trip.
- Remove various references to the hydration pattern in other classes.
- Refactor `DataResource::createCommon()` to be inlined into the methods that call it. We do this because adding the factory method caused CodeClimate to complain about too many methods, so we remove this one.
- Some CS improvements.

fixes WCMS-14059

- [x] Test coverage exists


modules/common/tests/src/Unit/DataResourceTest.php
```diff
@@ -82,4 +84,47 @@ public function testGetIdentifierAndVersionException() {
...
+      'txt',
+      DataResource::DEFAULT_SOURCE_PERSPECTIVE
+    );
+
+    $clone_data_resource = $data_resource->createNewPerspective('local_url', 'uri://foo/bar');
+
+    // Not the same object.
+    $this->assertNotSame($data_resource, $clone_data_resource);
+    // Clone contains 'local_url' perspective.
+    $ref_perspective = new \ReflectionProperty($clone_data_resource, 'perspective');
```
dafeder: why not just use `getPerspective()` here @paul-m ?

modules/common/tests/src/Unit/DataResourceTest.php
```diff
@@ -82,4 +84,47 @@ public function testGetIdentifierAndVersionException() {
...
+      'txt',
+      DataResource::DEFAULT_SOURCE_PERSPECTIVE
+    );
+
+    $clone_data_resource = $data_resource->createNewVersion();
+
+    // Not the same object.
+    $this->assertNotSame($data_resource, $clone_data_resource);
+    // Clone contains new version.
+    $ref_version = new \ReflectionProperty(DataResource::class, 'version');
```
dafeder: Same question - why not use `getVersion()`?

dafeder (changes requested):
See two questions - otherwise looks great!

modules/common/tests/src/Unit/DataResourceTest.php
paul-m: Fixed.

modules/common/tests/src/Unit/DataResourceTest.php
paul-m: Fixed.

dafeder (approved):

PR #3935, opened 2023-04-06
Skip admin tests until admin_toolbar is added to recommended-project

janette:


modules/metastore/modules/metastore_admin/metastore_admin.install
```diff
@@ -113,3 +113,10 @@ function metastore_admin_update_8011() {
 
   drupal_flush_all_caches();
 }
+
+/**
+ * Uninstall admin_toolbar and config_update for move to recommended-project.
+ */
+function metastore_admin_update_8012() {
+  \Drupal::service('module_installer')->uninstall(['admin_toolbar','config_update']);
```
jastraat: @janette I don't know that we want to uninstall the modules here. Someone might be using those modules for something else other than DKAN. And if they don't add these two modules to their project composer before upgrading to this version of DKAN, they wouldn't be able to uninstall the modules anyway; the modules wouldn't be in the code base. I'd suggest adding a note to the release notes similar to the indexer module instead.

janette: Right the release notes will instruct users to add the modules to their projects to keep the functionality. For anyone that does not add it, they will get update errors without this hook.

janette: Will remove the hooks and only provide guidance in the Release Notes.

dafeder (approved):

PR #3938, opened 2023-04-10
Docs contrib instructions and minor sphinx  fixes.

dafeder:
The main thing here is to add a tip to build the docs without using `sudo` for `pip`. 

Also cleaned up a few things in the RST and related files that were throwing errors in the sphinx build.


docs/source/conf.py
```diff
@@ -66,7 +66,6 @@
 from pygments.lexers.web import PhpLexer
 lexers["php"] = PhpLexer(startinline=True, linenos=1)
 lexers["php-annotations"] = PhpLexer(startinline=True, linenos=1)
-primary_domain = "php"
```
dafeder, line 69: This was throwing an error in my build, and some googling suggested it was not necessary. Certainly doesn't seem to change anything?

janette (approved):

PR #3939, opened 2023-04-11
PDC specific customizations (Reference only)

jastraat:
This is not for review or committing to DKAN proper but just for reference. These PDC changes include the following:

- Unify datastore and metastore tags for Swagger UI's Open API spec. (might be removed in WCMS-16863 )
- Add the complete API documentation path to DKAN Common's spec.
- Add the dataset-specific API documentation path to DKAN Metastore's spec.
- Add weight to and sort the ApiDocs plugins (might be removed in WCMS-16863 )
- Tag Search API's DKAN datasource plugin's query to later modify - Could be more broadly useful
- Add datastore filter conditions is_empty and not_empty - Could be more broadly useful; should not have underscores to match other filter conditions.

### DIFF
--- a/modules/common/docs/openapi_spec.json
+++ b/modules/common/docs/openapi_spec.json
@@ -65,5 +65,27 @@
             }
         }
     },
-    "paths": {}
+    "paths": {
+        "/api/1": {
+            "get": {
+                "operationId": "docs-get-complete",
+                "summary": "Complete json documentation",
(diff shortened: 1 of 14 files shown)


jastraat:
Closing since as of this sprint, the only remaining customization in PDC is the addition of the datastore explorer filter conditions "is_empty" and "not_empty"

PR #3958, opened 2023-05-08
Issue #3839 - Disentangle harvest register and revert

stefan-korn:
## Steps to Recreate

- Register a harvest (h1) with a dataset with a given identifier
- Run the h1 harvest
- Confirm the dataset was created
- Register a 2nd harvest (h2) with a new identifier that contains a slightly updated dataset with the same identifier as the first harvest (e.g. The json file was renamed in a new version.)
- Run the h2 harvest
- Confirm the dataset was updated
- Deregister the old h1 harvest: ddev drush dkan:harvest:deregister h1
- Note that this deleted the dataset even though it was also connected to the h2 harvest

## QA Steps

- Register a harvest using a local json file: ddev drush dkan:harvest:register --identifier=h1 --extract-uri=https://mylocal.ddev.site/sites/default/files/h1.json
- Run the h1 harvest: ddev drush dkan:harvest:run h1
- Confirm the dataset node was created
- Register a 2nd harvest using a local json file with an updated dataset with the same identifier as the first harvest: ddev drush dkan:harvest:register --identifier=h2 --extract-uri=https://mylocal.ddev.site/sites/default/files/h2.json
- Run the h2 harvest: ddev drush dkan:harvest:run h2
- Confirm that the dataset node was updated
- Deregister the h1 harvest: ddev drush dkan:harvest:deregister h1
- Asks for confirmation with a warning "If you deregister a harvest with published datasets, you will not be able to bulk revert the datasets connected to this harvest."
- -- After confirming, returns "Successfully deregistered the h1 harvest."
- Confirm that the harvest was deregistered: ddev drush dkan:harvest:list. (h1 should not be listed)
- Confirm that the dataset original created by the h1 harvest still remains


jastraat:
It looks like we can also deregister a harvest via the WebServiceApi (/api/1/harvest/plans/{identifier} with a DELETE); any concerns with having that api call behave differently after the change? There wouldn't be a warning about losing the ability to bulk delete in this case.

dafeder:
Thanks @stefan-korn! One request, you can see this is now failing our code analysis tests for code complexity, because of the three level if > try > if. But, I think actually the try/catch is not necessary in the command method at all though -- it catches the exception and just prints it to the terminal, which is what would have happened to the uncaught exception anyway. Do you mind removing the try/catch and seeing if that causes any ill effects?

modules/harvest/src/Commands/HarvestCommands.php
```diff
@@ -126,13 +126,19 @@ protected function buildPlanFromOpts($opts) {
    * @command dkan:harvest:deregister
    */
   public function deregister($id) {
-    try {
-      if ($this->harvestService->deregisterHarvest($id)) {
-        $message = "Successfully deregistered the {$id} harvest.";
+    $this->logger->warning('If you deregister a harvest with published datasets, you will not be able to bulk revert the datasets connected to this harvest.');
```
janette: @stefan-korn could you: Split this 159 characters long line (which is greater than 140 authorized).

stefan-korn:
Sorry, been some time off the track. Now I have merged the fork with latest 2.x head and incorporated your suggestions about try/catch and line length.

modules/harvest/src/Commands/HarvestCommands.php
```diff
@@ -127,13 +127,16 @@ protected function buildPlanFromOpts($opts) {
...
+       not be able to bulk revert the datasets connected to this harvest.');
+    if ($this->io()->confirm("Deregister harvest {$id}")) {
       if ($this->harvestService->deregisterHarvest($id)) {
         $message = 'Successfully deregistered the ' . $id . ' harvest.';
       }
     }
-    catch (\Exception $e) {
-      $message = $e->getMessage();
+    else {
+      $message = "Skipped deregistering harvest {$id}";
```
dafeder: Just a minor request @stefan-korn - "skipped" seems like the wrong phrasing since the user simply chose to abort the command. I think in other drush commands with a confirmation like this the standard is simply to throw a `Drush\Exceptions\UserAbortException`. 

See for instance https://github.com/drush-ops/drush/blob/12.x/src/Commands/core/SiteInstallCommands.php#L361

dafeder (changes requested):
Looks great overall!

modules/harvest/src/Commands/HarvestCommands.php
stefan-korn: Okay, I have changed this.

jastraat (approved):
LGTM!

dafeder (approved):

PR #3965, opened 2023-05-12
Factor out drupal_static for resource display

jastraat:
Drupal has had a meta issue to remove all drupal_static calls in core and deprecate the functionality. With that in mind, we're working to move the configuration of the resource download url in the metadata to configuration (that can be edited through the UI) instead of using drupal_static.

This change will need to be reflected in the release notes so that sites who are currently calling &drupal_static('metastore_resource_mapper_display')
know that they will need to update their code.

Included changes:
- metastore_resource_mapper_display static variable fully removed
- new configuration setting in metastore.settings called 'resource_perspective_display' with two options: source (default) and local_url
- retrieveDownloadUrlFromResourceMapper() uses the new config setting to determine how to display the download url
- The new config setting can be modified on the Resources form.

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

- Install a demo vanilla DKAN site
- Visit /admin/dkan/resources and confirm the new 'Resource download url display' select defaults to 'source'. Don't save anything.
- Create a new published dataset with a remote distribution download url like http://demo.getdkan.org/sites/default/files/distribution/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/Bike_Lane.csv
- Check the metastore api path for the dataset e.g. https://dkan10.ddev.site/api/1/metastore/schemas/dataset/items/[uuid]
- Confirm that the distribution -> downloadURL shows the remote source url (e.g. http://demo.getdkan.org....)
- Go back to /admin/dkan/resources and change 'Resource download url display' to local url and save
- Clear the drupal cache (ddev drush cr)
- Re-visit the metastore api path for the dataset and confirm that the distribution -> downloadURL now shows the local file path to the resource.

### DIFF
--- a/modules/datastore/src/Form/ResourceSettingsForm.php
+++ b/modules/datastore/src/Form/ResourceSettingsForm.php
@@ -25,7 +25,7 @@ class ResourceSettingsForm extends ConfigFormBase {
    * {@inheritdoc}
    */
   protected function getEditableConfigNames() {
-    return ['datastore.settings'];
+    return ['datastore.settings', 'metastore.settings'];
   }
 
   /**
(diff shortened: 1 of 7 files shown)


dafeder (approved):
Looks great. I think at some point it might make sense to consolidate anything "resource" related into the datastore module since the division of resource between datastore and metastore doesn't make a ton of sense IMO, but that's obviously way beyond the scope of this.

Issue #3512, opened 2021-06-09
Create special form widget for vCard email field


dafeder:
DCAT-AP uses the vCard format to store contact info, which contains a hasEmail field. The hasEmail field requires a `mailto:` prefix before the email address, which is obviously not intuitive to have to enter in the form. Let's create a new email widget available from the UI schema that handles this out of the view of the user.

## Acceptance criteria

- Widget hides `mailto` prefix from input when editing existing vCard with this value
- Prefix is added after submission before validation
- Input gets html5 "type=email" attribute for initial browser validation
- In default dataset.ui schema, `hasEmail` is assigned to the widget

dafeder:
Upon further review, this is a problem more of our own making. We do want to fix the form to deal better with `hasEmail` values that include the `mailto:` prefix, but in the metadata that triggered the issue we are not harvesting from a third party and can simply remove them ourselves. De-prioritizing for the time being.

stefan-korn:
@dafeder : I am currently also stumbling into this.

The vcard format requires the mailto: prefix for the hasEmail property. Therefore this is added in the demo content before the email address, like this: mailto:data.admin@example.com

Now data.json is outputting the vcard:hasEmail with the mailto prefix.

But upon saving the dkan dataset in the drupal backend, you are getting "invalid email" message (probably from HTML-5 validation?) and cannot save until you remove the mailto prefix. Now if you remove the mailto prefix the vcard:hasEmail property is no longer valid I suppose.

So this seems to be an issue at hand. If you are entering datasets via the backend you are not able to provide valid vcard:hasEmail property as far as I can see?

Imho the mailto prefix should be automatically set on the hasEmail property. Only on output, but not in the data?

stefan-korn:
see #3966 for possible fix.

Can probably be improved, but should do it for the moment.  First three acceptance criterias might be met. About the last one I am not sure, whether it is really necessary to have a separate widget.

It's not so easy (for me perhaps) to get into the output of data.json and just inject it there. So I opted for inserting mailto: into the data saved (Prefix is added after submission before validation).

dafeder:
Flagging this for later - not clear if the merged PR fully addresses the issue or if there are more use-cases to address to be fully compliant with DCAT-US

PR #3966, opened 2023-05-12
Issue #3512: Create special form widget for vCard email field

stefan-korn:
possible fix to #3512 

This only addresses the issue with editing the email field through the UI for a vCard with the hasEmail property that was imported via the API.

## Steps to Recreate

- Create a dataset through the API / via a harvest that has a contactPoint with the hasEmail property

`"contactPoint": {
        "@type": "vcard:Contact",
        "fn": "First Name",
        "hasEmail": "mailto:email@example.com"
      },`

- Attempt to edit the dataset through the UI. The validation of the email field fails.

## QA Steps

- Create a dataset through the API / via a harvest that has a contactPoint with the hasEmail property
- Only the email appears in the email form field (no mailto:)
- Edit and save the dataset through the UI without an error.
- When visiting the API path for the metastore ( /api/1/metastore/schemas/dataset/items/{identifier} ) the hasEmail still shows the mailto: prefix.

### DIFF
--- a/modules/json_form_widget/src/StringHelper.php
+++ b/modules/json_form_widget/src/StringHelper.php
@@ -82,6 +82,7 @@ class StringHelper implements ContainerInjectionInterface {
     // Add extra validate if element type is email.
     if ($element['#type'] === 'email') {
       $element['#element_validate'][] = [$this, 'validateEmail'];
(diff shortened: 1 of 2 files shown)


jastraat (approved):
Followed the QA steps as outlined, and this allowed editing a previously created via the API dataset with an email containing a hasEmail property through the UI and saving successfully.

PR #3969, opened 2023-05-17
14821 15104 validate previous mysql import

paul-m:
- Adds `ImportedDatabaseTableInterface` to give a generalized way to determine if a table has been imported.
- Implements `ImportedDatabaseTableInterface` for `DatabaseTable`, where if the table exists and is not empty, it counts as imported.
- Ensure that `MysqlImport` service requires `ImportedDatabaseTableInterface` as storage.
- `MysqlImport` calls `ImportedDatabaseTableInterface::hasBeenImported()` on storage to determine if the import needs to occur.
- Tests to prove it all works.

Todo: 
- Iterate on better validation for imported tables.
- Add strategy for determining whether to download files to local.


paul-m:
New branch: https://github.com/GetDKAN/dkan/pull/3977

janette (approved):

PR #3970, opened 2023-05-17
README indicate Drupal 10 support

grugnog:
Not sure if we are ready for this yet, but IIRC we are close/there. Would we prefer to list 9 or 10, or just 10?


jastraat:
I'd advocate for "Drupal 9/10"

jastraat (approved):
Looks great!

PR #3971, opened 2023-05-18
Fix ltrim() deprecation message when null value

dafeder:
The new ltrim() calls from #3966 throw a deprecation warning in tests in php 8.1, because in some cases they receive a null argument:
image

See test output [here](https://app.circleci.com/pipelines/github/GetDKAN/dkan/4221/workflows/60f0b54d-ae0c-4422-9717-e618dcbc73c9/jobs/26353).


jastraat (approved):
Tested locally using ddev dkan-phpunit. It looks like #3972 added some tests that are evaluating as incomplete. But the depreciation error related to the email default value is resolved by this PR.

PR #3976, opened 2023-06-02
Turn off innodb_strict_mode on tableCreate()

janette:


paul-m:
- Overrides `MysqlDatabaseTable::tableCreate()` so that it always turns off `innodb_safe_mode`.
- Adds tests to verify that this gives us what we want: More table columns with longer names.
- Removes `MySqlDatabaseTable::setTable()` and `MySqlDatabaseTable::query()` which are the reason we couldn't deploy some sites using DKAN 2.16.

janette (commented):
Nice

jastraat (approved):
Per Paul & Janette this is good to go.

PR #3984, opened 2023-06-28
Remove use of dktl in user guide

janette:


jastraat (approved):
This is a great change.

At some point we may want to standardize on how we write drush commands. e.g. with all colons or with the dashes...

PR #3987, opened 2023-07-04
15750 Jobstore use deprecated table name

paul-m:
Allows for using both hashed and deprecated jobstore default table names.

Some other cleanup related to the `JobStore` class, such as correcting type hints.

Also deletes `modules/datastore/tests/src/Unit/Storage/JobStoreTest.php` because it's a duplicate of `modules/common/tests/src/Unit/Storage/JobStoreTest.php`


jastraat:
I tested this PR with a local site that I had running a pre- 2.16.1 version of DKAN. With the earlier version of DKAN, I checked the import dashboard and confirmed that I had successful imports. Then I updated to the latest dev version of DKAN. I confirmed that the import status was changed from "done" to "waiting". Then I applied the patch for the code in Paul's PR and the imports were again listed as "done".

I did have a question as to how the new destruct() function is currently being used outside of the new test...? Aside from that, this looks good.

jastraat (approved):
Paul explained that we needed to destruct the deprecated table name in addition to the new hashed one if present.

jastraat:
Quick mention that this really only kicks the can down the road to limit damage for sites that have not updated to 2.16.1 or later. Open Payments did their imports with the new hashed table names. But it ALSO has the old table names. So we still need to address that issue.

PR #3996, opened 2023-08-01
11948 Use pre-existing local file for import

paul-m:
- Since it was unclear exactly which layer should contain this solution, I ended up reading a lot of code and thus doing minor refactoring in some places, and adding a lot of docblocks and CS fixes for some files.
    - Notably, `DataResource` now has `getUniqueIdentifierNoPerspective()`, to go along with its existing `getUniqueIdentifier()`, and updating places where the no perspective id was being created on the fly.
    - This all works towards making the whole thing more readable overall.
- Adds a new DKAN common config called `always_use_existing_local_perspective`. It’s a bool, and makes your DKAN override the file fetcher to use an existing file rather than copy one from the source.
- Adds a warning to the site status page telling you that you have `always_use_existing_local_perspective` enabled.
- `\Drupal\common\FileFetcher\Factory` is now `\Drupal\common\FileFetcher\FileFetcherFactory`. This name change comes along with the ability to check the `always_use_existing_local_perspective` configuration and use our new file-fetcher remote processor depending on what it says.
- Adds `\Drupal\common\FileFetcher\FileFetcherRemoteUseExisting`, which is a file fetcher processor to replace `Remote` if we want to use the existing file.
- Adds tests to prove all this stuff does what it says.
- Added a page of documentation: https://dkan--3996.org.readthedocs.build/en/3996/user-guide/guide_local_files.html

There's no integrity check performed on the file, other than whether it exists in the file system... For instance the file could exist but be zero bytes long. We'll follow up with issues about this process.


docs/source/user-guide/guide_local_files.rst
```diff
@@ -0,0 +1,175 @@
...
+                    "importer_percent_done": 0,
+                    "importer_status": "waiting",
+                    "importer_error": null,
+                    "table_name": "not found"
+                }
+            ]
+        }
+    }
+
+We want the resource_id of the distribution. Here’s an example of extracting it using jq:
```
dafeder:
```suggestion
We want the resource ID from the distribution. You can just select and copy it from the JSON output, or extract it using `jq <https://jqlang.github.io/jq/>`_:
```

docs/source/user-guide/guide_local_files.rst
```diff
@@ -0,0 +1,175 @@
+Importing Very Large Files
+--------------------------
+
+Sometimes datasets are very large. They are represented by large files and create large tables in the site database.
+
+It can help to download the files as a separate step from the import process, outside of the DKAN queue process. Then we can instruct DKAN to use those existing files for processing into the database.
+
+These instructions assume you’ve already set up a dataset, either through the UI or by running a harvest or through the dataset API. We’ll need the ID of the existing dataset.
```
dafeder:
```suggestion
These instructions assume you’ve already created a dataset (through the the UI, a harvest or the API) that contains a distribution pointing to a remote CSV file, and that this CSV file has not yet been imported to the datastore. We’ll need the ID of the existing dataset.
```

docs/source/user-guide/guide_local_files.rst
```diff
@@ -0,0 +1,175 @@
+Importing Very Large Files
+--------------------------
+
+Sometimes datasets are very large. They are represented by large files and create large tables in the site database.
+
+It can help to download the files as a separate step from the import process, outside of the DKAN queue process. Then we can instruct DKAN to use those existing files for processing into the database.
```
dafeder:
```suggestion
Very large datasets can present challenges for the datastore import process. What "very large" means will depend on your server infrastructure, but it's safe to say that when dealing with multi-gigabyte CSV files you may start to encounter broken file transfers or timeouts.

Because DKAN will make a local copy of a remote file in order to import it, it can help to prepare the import by transfering the files to the web server manually (or with shell scripts). Then you can instruct DKAN to use that already-existing local copy for processing into the database.
```

docs/source/user-guide/guide_local_files.rst
```diff
@@ -0,0 +1,175 @@
...
+
+These instructions assume you’ve already set up a dataset, either through the UI or by running a harvest or through the dataset API. We’ll need the ID of the existing dataset.
+
+TL;DR
+=====
+
+Abbreviated instructions. Read further for more details.
```
dafeder:
```suggestion
Below are some abbreviated instructions. Read past this section for details.
```

docs/source/user-guide/guide_local_files.rst
```diff
@@ -0,0 +1,175 @@
...
+
+- Use :code:`drush dkan:datastore:prepare-localized [id]` to set up the local file directory.
```
dafeder:
```suggestion
- Use :code:`drush dkan:datastore:prepare-localized [resource_id]` to set up the local file directory.
```

docs/source/user-guide/guide_local_files.rst
```diff
@@ -0,0 +1,175 @@
...
+- Transfer the file in some way other than a DKAN import.
```
dafeder:
```suggestion
- Transfer the file to that directory on the server using the command line or your file transfer program of choice.
```

docs/source/user-guide/guide_local_files.rst
```diff
@@ -0,0 +1,175 @@
...
+- Tell DKAN to use local files with this configuration: :code:`drush config:set common.settings always_use_existing_local_perspective 1`
```
dafeder:
```suggestion
- Set DKAN to use existing local files for imports: :code:`drush config:set common.settings always_use_existing_local_perspective 1`
```

docs/source/user-guide/guide_local_files.rst
```diff
@@ -0,0 +1,175 @@
...
+- Perform the import: :code:`drush dkan:datastore:import --deferred [id] && drush queue:run datastore_import`
+- Turn off the local file preference with configuration: :code:`drush config:set common.settings always_use_existing_local_perspective 0`
```
dafeder:
```suggestion
- Disable the local file configuration: :code:`drush config:set common.settings always_use_existing_local_perspective 0`
```

docs/source/user-guide/guide_local_files.rst
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+We want the resource_id of the distribution. Here’s an example of extracting it using jq:
+
+.. code-block::
+
+    % drush dkan:dataset-info bf215cd3-dd81-498c-b57a-4847dbeaac44 | jq -r '.latest_revision.distributions[].resource_id'
+    1fecf29222b12fc1ce2678abbc8f870f
+
+Now that we have the resource ID, we can tell DKAN to prepare for some other script or process to download the file. We use Drush to do this:
```
dafeder:
```suggestion
Now that we have the resource ID, we can tell DKAN expect some other process to download the file. We use :code:`drush dkan:datastore:prepare-localized` to do this:
```

docs/source/user-guide/guide_local_files.rst
```diff
@@ -0,0 +1,175 @@
...
+    % drush dkan:datastore:prepare-localized 1fecf29222b12fc1ce2678abbc8f870f
+    {
+        "source": "http:\/\/demo.getdkan.org\/sites\/default\/files\/distribution\/cedcd327-4e5d-43f9-8eb1-c11850fa7c55\/Bike_Lane.csv",
+        "path_uri": "public:\/\/resources\/1fecf29222b12fc1ce2678abbc8f870f_1691778866",
+        "path": "\/var\/www\/html\/docroot\/sites\/default\/files\/resources\/1fecf29222b12fc1ce2678abbc8f870f_1691778866",
+        "file_uri": "public:\/\/resources\/1fecf29222b12fc1ce2678abbc8f870f_1691778866\/Bike_Lane.csv",
+        "file": "\/var\/www\/html\/docroot\/sites\/default\/files\/resources\/1fecf29222b12fc1ce2678abbc8f870f_1691778866\/Bike_Lane.csv"
+    }
+
+We can pipe this to jq as well:
```
dafeder:
```suggestion
We can pipe this to jq as well, to extract the specific path where DKAN expects the file:
```

docs/source/user-guide/guide_local_files.rst
```diff
@@ -0,0 +1,175 @@
...
+
+    % ddev drush dkan:dataset-info bf215cd3-dd81-498c-b57a-4847dbeaac44 | jq -r '.latest_revision.distributions[].file_path'
+    public://resources/1fecf29222b12fc1ce2678abbc8f870f_1691778866/Bike_Lane.csv
+
+Transfer The File
+=================
+
+In this example we’ll just use wget to copy the file at the command line. At the moment, automating this process is left as an exercise for the reader, but a combination of bash and jq should be able to accomplish this.
+
+From the output of :code:`dkan:datastore:prepare-localized` we get the path. In our case this is :code:`/var/www/html/docroot/sites/default/files/resources/1fecf29222b12fc1ce2678abbc8f870f_1691778866`
```
dafeder:
```suggestion
From the output of :code:`dkan:datastore:prepare-localized` we get the path. In our case this is */var/www/html/docroot/sites/default/files/resources/1fecf29222b12fc1ce2678abbc8f870f_1691778866*
```

modules/common/common.install
```diff
@@ -44,6 +44,19 @@ function common_requirements($phase): array {
...
       }
     }
+    // Alert the user if they have always_use_existing_local_perspective on.
+    if ($config = \Drupal::config('common.settings')) {
+      if ($config->get('always_use_existing_local_perspective') ?? FALSE) {
+        $requirements['dkan using local files'] = [
+          'title' => t('DKAN Using Local Files'),
+          'value' => 'always_use_existing_local_perspective: true',
+          'description' => t(
+            'DKAN is configured to use localized files rather than download them during import.'
```
dafeder:
```suggestion
            'DKAN is configured to use localized files rather than download them during import.
            Disable this setting on the CLI by running: <code>drush config:set common.settings always_use_existing_local_perspective 0</code>'
```

modules/common/common.install
```diff
@@ -44,6 +44,19 @@ function common_requirements($phase): array {
         ];
       }
     }
+    // Alert the user if they have always_use_existing_local_perspective on.
+    if ($config = \Drupal::config('common.settings')) {
+      if ($config->get('always_use_existing_local_perspective') ?? FALSE) {
+        $requirements['dkan using local files'] = [
+          'title' => t('DKAN Using Local Files'),
+          'value' => 'always_use_existing_local_perspective: true',
```
dafeder, line 52:
```suggestion
```

dafeder: Value is not required and this does not really add clarity

modules/common/src/DataResource.php
```diff
@@ -344,7 +394,12 @@ private static function getDistribution($identifier) {
    * Generates MD5 checksum for a file.
    */
   public function generateChecksum() {
-    $this->checksum = md5_file($this->filePath);
```
dafeder, line 347: Can this get run on other perspectives? Is it possible that failing silently like this would have unexpected effects? Consider adding try/catch

modules/datastore/src/Drush.php
```diff
@@ -198,36 +219,58 @@ public function drop(string $identifier, array $options = ['keep-local' => FALSE
...
+   *
+   * @param string $identifier
+   *   Datastore resource identifier, e.g., "b210fb966b5f68be0421b928631e5d51".
+   *
+   * @command dkan:datastore:prepare-localized
+   *
+   * @todo Move all this to the ResourceLocalizer so we're responsible for test
+   *   coverage.
    */
-  protected function jobstorePrune($ref_uuid) {
```
dafeder, line 209: What happened here? we don't need this anymore?

dafeder (changes requested):
Amazing work here @paul-m. Take a look at a few suggestions and questions I had.

modules/common/common.install
paul-m: Only the value is displayed if you use `drush rq`

modules/datastore/src/Drush.php
paul-m: I looked and looked, and it's dead code.

modules/common/common.install
dafeder: Ah ok. Might be good to make it more human-readable at some point but not a blocker

modules/datastore/src/Drush.php
dafeder: Cool, nice catch.

dafeder (approved):
Approved, looks like we need to re-run a cypress test though

PR #3999, opened 2023-08-10
16181 Chunk Imports Are Marked Done Before They're Done

paul-m:
Chunked imports (using the datastore module, without datastore_mysql_import) were discovered to be quitting after partial imports on multiple queue instances.

Investigation by @jastraat revealed this was due to the DB import short-circuit which was intended to allow graceful recovery from DB import timeouts in datastore_mysql_import imports.

Solution:

Limit the short-circuit behavior to datastore_mysql_imports imports by only implementing `ImportedItemInterface` on `MySqlDatabaseTable` and not `DatabaseTable`.

### DIFF
--- a/modules/datastore/tests/src/Kernel/Storage/DatabaseTableTest.php
+++ b/modules/datastore/tests/src/Kernel/Storage/DatabaseTableTest.php
@@ -31,10 +31,13 @@ class DatabaseTableTest extends KernelTestBase {
   /**
    * Ensure that non-mysql-import tables do not implement hasBeenImported().
    *
+   * We don't want DatabaseTable to be able to report that the table has already
+   * been imported.
+   *
    * We exercise the whole import service factory pattern here to make sure we
    * get the DatabaseTable object we expect.
    */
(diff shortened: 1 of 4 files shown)


jastraat (approved):
Tested this out with the chunking importer and a file that needed to be requeued and looks good!

modules/datastore/tests/src/Kernel/Storage/DatabaseTableTest.php
```diff
@@ -27,9 +29,12 @@ class DatabaseTableTest extends KernelTestBase {
...
 
   /**
-   * @covers ::hasBeenImported
+   * Ensure that non-mysql-import tables do not implement hasBeenImported().
+   *
+   * We exercise the whole import service factory pattern here to make sure we
+   * get the DatabaseTable object we expect.
    */
-  public function testHasBeenImported() {
+  public function testNoHasBeenImported() {
```
janette: @paul-m can this be testHasNotBeenImported?

PR #4003, opened 2023-08-21
Fix pass by reference error

janette:
When uploading new distribution file:
```
Notice: Only variables should be passed by reference in _json_form_widget_update_file() (line 132 of /var/www/html/dkan/modules/json_form_widget/json_form_widget.module) 
```


jastraat (approved):
LGTM! (Pairs well with PR 4006)

PR #4004, opened 2023-08-22
Make FileFetcher a separate queue item

paul-m:
- Modifies `ResourceLocalizer` so it can perform its function or defer to a queue item.
- Adds `LocalizeQueueWorker`, which can perform localization for a dataset within the queue.
- Adds `drush dkan:datastore:localize` which is analogous to `dkan:datastore:import`. It can queue items or perform the localization directly.
- `ResourceLocalizer` now sends an event when localization is complete: `ResourceLocalizer::EVENT_RESOURCE_LOCALIZED`
- `DatastoreSubscriber` subscribes to this new event and creates corresponding datastore import queue items in response.
- Modifies tests to pass where appropriate. Mostly this involves running the new `localize_import` queue in addition to the `datastore_import` queue.
- Some docblock and other CS improvements on files touched by the PR changes.
- Modifies the docs for readthedocs: https://dkan--4004.org.readthedocs.build/en/4004/index.html

QA:

- Add a dataset. Suggested source URL: https://demo.getdkan.org/sites/default/files/distribution/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/Bike_Lane.csv
- Visit the dataset import status page: `/admin/dkan/datastore/status` You should see that the dataset is not fetched and not stored.
- Use Drush to list the queue items: `drush queue:list`
- There should be one item for the `localize_import` queue.
- Run this queue:`drush queue:run localize_import`
- Now running `drush queue:list` should show you one queue item for `datastore_import`.
- Visit the import status page again, and it should show you that the fetch has occurred, and that the store is ‘stopped.’
- Run the `datastore_import` queue, either with cron or `drush queue:run`.
- You should be able to visit the dataset normally and see it. (Ensure you have built and installed the frontend.)


modules/datastore/src/Form/DashboardForm.php
```diff
@@ -115,7 +115,7 @@ public static function create(ContainerInterface $container): self {
       $container->get('dkan.metastore.service'),
       $container->get('pager.manager'),
       $container->get('date.formatter'),
-      $container->get('dkan.datastore.service.post_import'),
+      $container->get('dkan.datastore.service.post_import')
```
jastraat: Drupal coding standards say to keep the ending comma: https://www.drupal.org/docs/develop/standards/php/php-coding-standards#:~:text=Note%20that%2C%20as%20seen%20above,end%20of%20the%20list%20later.

paul-m: It's not an array.

jastraat: *heel to forehead. Completely missed that these were function parameters. Of course they are given they are all services. I saw "DashboardForm" and my head jumped straight to Form API. My bad :/

dafeder:
I could be wrong but it seems to me that this has caused a regression with the new functionality added in #11948. @paul-m I built DKAN off your branch and created the following dataset:

```json
{
    "@type": "dcat:Dataset",
    "identifier": "0b747da6-a17d-573c-9739-98f011a8bb5e",
    "accessLevel": "public",
    "title": "OP Dataset",
    "description": "An open payments dataset",
    "contactPoint": {
        "fn": "Jane Doe",
        "hasEmail": "mailto:data.admin@example.com"
    },
    "issued": "2016-06-22",
    "license": "http://opendatacommons.org/licenses/by/1.0/",
    "modified": "2016-06-22",
    "publisher": {
        "@type": "org:Organization",
    },
    "distribution": [
        {
            "@type": "dcat:Distribution",
            "downloadURL": "https://download.cms.gov/openpayments/PGYR22_P063023/OP_DTL_RSRCH_PGYR2022_P06302023.csv",
            "mediaType": "text\/csv",
            "format": "csv",
            "title": "Dataset"
        }
    ],
    "keyword":["tag1"]
}
```

I then ran `drush dkan:datastore:prepare-localized e6f2dc5de571162b8ec4a71cc46de625`, set `always_use_existing_local_perspective` to 1, downloaded the CSV file to _sites/default/files/resources/e6f2dc5de571162b8ec4a71cc46de625_1695137431_, and ran `drush queue:run localize_import`.  It did not seem to recognize the existing local perspective and re-downloaded the file into the same place. 

Can you confirm? If this is a regression I think we should try to fix before merging.

paul-m:
Fixing the regression in https://github.com/GetDKAN/dkan/pull/4035 before proceeding here.

dafeder (commented):
One issue I'm seeing - it would be nice if it output a notice when finishing a localize task, as it does when finishing an import during a queue run. See `ImportQueueWorker::processResult()` for where this happens in the import job. This does actually have a functional impact as if you're running this via Drush over ssh, too long without i/o will cause the ssh session to time out.

dafeder (changes requested):
Another issue that is I think an actual bug - if the localize_import job encounters a 404, it throws an exception and re-queues. So if there is a typo in one of your URLs, it will be re-queued forever (or at least until you delete the queue manually). Suggest that this also work more like the import job, where exceptions from curl are caught the way mysql errors are, stored in the error field in the jobstore result so that they can be seen in the dashboard, and then the item is considered done with statuis "error".

dafeder (changes requested):
Need to add docs on dkan:datastore:localize

paul-m:
RE: 404 leading to an endless loop of queued localizations:

Added a case where if `localizeTask()` returns a `Result::ERROR` object, we don't throw an exception, so the queue item won't be re-queued.

This also changes the way logging happens, so it improves the output when you localize through Drush.

This is tested in `Drupal\Tests\datastore\Kernel\Plugin\QueueWorker\LocalizeQueueWorkerTest::testNotRequeuedOnError()`

paul-m:
Updated the docs.

jastraat (commented):
Overall love these changes and tested with PDC successfully.

I also copied the PDC DB from production and ran some harvests to see what would happen and I did run into some errors because it looked like localizer was still looking for the files on the production server versus in the local environment.

dafeder (approved):

PR #4011, opened 2023-09-07
API functional test refactor to BrowserTestBase

dafeder:
Following lead of #4008, converting the API functional tests to use BrowserTestBase instead of ExistingSiteBase. Test logic unchanged.

Functional test run time is significantly higher now. We may want to consider using parallelism on CircleCI if it starts to take longer than the Cypress tests.

### DIFF
--- a/modules/common/tests/src/Functional/Api1TestBase.php
+++ b/modules/common/tests/src/Functional/Api1TestBase.php
@@ -66,7 +66,6 @@ abstract class Api1TestBase extends BrowserTestBase {
       RequestOptions::JSON => $data,
       RequestOptions::AUTH => $this->auth,
       RequestOptions::HTTP_ERRORS => $httpErrors,
-      RequestOptions::TIMEOUT => 10000,
     ]);
   }
 
--- a/modules/common/tests/src/Functional/Api1TestBase.php
+++ b/modules/common/tests/src/Functional/Api1TestBase.php
@@ -2,49 +2,56 @@
 
 namespace Drupal\Tests\common\Functional;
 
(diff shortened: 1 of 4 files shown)


modules/common/tests/src/Functional/Api1TestBase.php
```diff
@@ -58,10 +62,11 @@ protected function assertJsonIsValid($schema, $json) {
   abstract public function getEndpoint(): string;
 
   protected function post($data, $httpErrors = TRUE) {
-    return $this->http->post($this->endpoint, [
+    return $this->httpClient->post($this->endpoint, [
       RequestOptions::JSON => $data,
       RequestOptions::AUTH => $this->auth,
       RequestOptions::HTTP_ERRORS => $httpErrors,
+      RequestOptions::TIMEOUT => 10000,
```
jastraat: @dafeder Just for my understanding, did you run into an issue with timeouts which resulted in adding this line?

dafeder: @jastraat oh huh, I forgot I left that in there. Not in a normal test run, but when using the debugger the request would timeout before I had a chance to inspect what was going on at different breakpoints, which this fixed. I should probably clean that up though.

jastraat (approved):
It was interesting to look at how this was refactored. Moving to BrowserTestBase seems like a great change. LGTM

dafeder:
Combining some test methods to speed up CI

dafeder:
OK seems to have shaved off 20 seconds or so, not nothing

paul-m (approved):
LGTM.

PR #4013, opened 2023-09-08
WCMS-16115: Apply HTML filtering to only properties that are configured to allow HTML.

jastraat:
Adds a new configuration setting for properties that can contain HTML and then limits HTML filtering (via the API) to those fields. If the config setting is not set, the allowed properties default to dataset and distribution description fields.

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

- [ ] Set (check) dataset and distribution descriptions as HTML allowed on the /admin/dkan/properties configuration form and save.
- [ ] Create a harvest file called data2.json with the contents of the data2.pdf file attached to WCMS-16115 and put it in docroot/sites/default/files
- [ ] Register the harvest: ddev drush dkan:harvest:register --identifier=h1 --extract-uri=https://yourdomain.ddev.site/sites/default/files/data2.json
- [ ] Run the harvest: ddev drush dkan:harvest:run h1
- [ ] Visit the dataset link ( /dataset/069d-826b ). Confirm that the dangerous HTML (and ampersand) are displayed as plain text in the dataset title and that the ampersand is displayed correctly in the keyword. Confirm that the onauxclick link was stripped from the dataset description
- [ ] Visit the node link for the dataset ( /node/[nid] ). Confirm that the dangerous HTML (and ampersand) are displayed as plain text in the dataset title and that the ampersand is displayed correctly in the keyword. Confirm that the onauxclick link was stripped from the dataset description

### DIFF
--- a/tests/src/Functional/DatasetBTBTest.php
+++ b/tests/src/Functional/DatasetBTBTest.php
@@ -319,6 +319,37 @@ class DatasetBTBTest extends BrowserTestBase {
     $this->assertDirectoryExists('public://resources/' . $refUuid);
   }
 
+  /**
+   * Test sanitization of dataset properties.
+   */
+  public function testSanitizeDatasetProperties() {
+    $config = $this->container->get('config.factory');
+    $metastoreSettings = $config->getEditable('metastore.settings');
+
+    // Set HTML allowed on dataset description.
+    $metastoreSettings->set('html_allowed_properties', ['dataset_description'])->save();
+
+    // Title with HTML and an ampersand.
+    $datasetRootedJsonData = $this->getData(123, 'This & That <a onauxclick=prompt(document.domain)>Right click me</a>', ['1.csv']);
+
+    $uuid = $this->getMetastore()->post('dataset', $datasetRootedJsonData);
+
+    $datasetRootedJsonData = $this->getMetastore()->get('dataset', $uuid);
+    $retrievedDataset = json_decode((string) $datasetRootedJsonData);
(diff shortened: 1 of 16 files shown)


tests/src/Functional/DatasetBTBTest.php
```diff
@@ -319,6 +319,37 @@ public function testDatastoreImportDeleteLocalResource() {
...
+    // Set HTML allowed on dataset description.
+    $metastoreSettings->set('html_allowed_properties', ['dataset_description'])->save();
+
+    // Title with HTML and an ampersand.
+    $datasetRootedJsonData = $this->getData(123, 'This & That <a onauxclick=prompt(document.domain)>Right click me</a>', ['1.csv']);
+
+    $uuid = $this->getMetastore()->post('dataset', $datasetRootedJsonData);
+
+    $datasetRootedJsonData = $this->getMetastore()->get('dataset', $uuid);
+    $retrievedDataset = json_decode((string) $datasetRootedJsonData);
```
dafeder, line 338: @jastraat no need to string then re-decode a RootedJsonData object, you should be able to access the title via `$datasetRootedJsonData->{"$.title"}`

dafeder (approved):

tests/src/Functional/DatasetBTBTest.php
jastraat: Ah - alrighty. I was mimicking what another one of the functional tests was doing.

PR #4014, opened 2023-09-11
Convert remaining DatasetTest method to BTB

paul-m:
Following through on this todo in `Drupal\Tests\dkan\Functional\DatasetTest::testResourcePurgeDraft`:

> @todo Move this test to \Drupal\Tests\dkan\Functional\DatasetBTBTest.


jastraat (approved):
These changes look great! Thank you for correcting that assertFalse message :)

tests/src/Functional/DatasetBTBTest.php
```diff
@@ -29,23 +29,83 @@ class DatasetBTBTest extends BrowserTestBase {
...
   /**
    * {@inheritdoc}
    */
-  protected $defaultTheme = 'starterkit_theme';
+  protected $defaultTheme = 'stark';
+
+  /**
+   * {@inheritdoc}
+   */
+  protected $strictConfigSchema = FALSE;
```
jastraat, line 48: @paul-m and I discussed this. This setting is related to change record https://www.drupal.org/node/3362879 and if it is left true, the test will fail due to poorly defined config schema in contrib modules as well. Suggest adding a comment referencing this!

PR #4017, opened 2023-09-14
Integrate getdkan/sql-parser

paul-m:
As part of a move to integrate all the getdkan-named external packages into the module, here we integrate [getdkan/sql-parser](https://github.com/GetDKAN/sql-parser).

This is fairly straightforward, since it's only one PHP class, and is only used in one place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,5 +30,8 @@ DKAN and related modules are freely-available under the ["GNU General Public Lic
 
 ---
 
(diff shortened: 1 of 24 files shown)


dafeder (changes requested):
Are you opposed to moving the sqlParser into datastore? We don't traditionally have a src folder in the dkan "module" and I don't really want to break that pattern for this.

dafeder (approved):

PR #4019, opened 2023-09-15
ResourceLocalizer constructor arg should be an interface

paul-m:
One DKAN site uses a bespoke file fetcher for S3 localization.

Changes in https://github.com/GetDKAN/dkan/pull/3996 broke this functionality, since we changed the signature of `ResourceLocalizer::__construct()`.

This PR changes the signature back so it's `\Contracts\FactoryInterface`.


jastraat (approved):
Tested this with PDC, and this addressed the issues we were seeing with PDC's decorated Filefetcher factory. Thanks!

PR #4027, opened 2023-09-24
Convert resource map to Drupal Entity API

paul-m:
- Creates a full-fledged resource mapping db table as a content entity: `ResourceMapping`.
  - All the columns in the current mapping schema are defined as base fields (that is, non-attached fields).
  - The table name is the same as the current abstract table implementation: `dkan_metastore_resource_mapper`
- Adapt `DataResource`, `Referencer`, and `ResourceMapper` to use the `ResourceMapping` entity.
- Add and update tests.

(internal: 17017)


paul-m:
Maybe we should add the content add/edit/delete niceties given the existence of https://github.com/GetDKAN/dkan/pull/3824

modules/metastore/src/Entity/ResourceMapping.php
```diff
@@ -0,0 +1,70 @@
...
+use Drupal\Core\Entity\ContentEntityBase;
+use Drupal\Core\Entity\EntityTypeInterface;
+use Drupal\Core\Field\BaseFieldDefinition;
+use Drupal\Core\StringTranslation\TranslatableMarkup;
+use Drupal\metastore\ResourceMappingInterface;
+
+/**
+ * Defines the resource mapping entity class.
+ *
+ * Used as storage by \Drupal\metastore\ResourceMapper.
```
dafeder, line 14: Suggestion... 
ResourceMappings store particular "perspectives" on a data file resource. The "source" perspective will be a URL to the original location of the resource. Local and LocalUrl perspectives record local copies of remote files for use by the DKAN datastore. A ResourceMapping entity has three levels of definition - the identifier, which is hash of the source URL; the version, which is a timestamp of when the source URL was accessed/downloaded; and the perspective itself, which will be one of those three values. These mappings allow DKAN to work with local copies of remote files, particularly for import into the datastore, and keep track of which remote files they represent.

modules/metastore/src/ResourceMapper.php
```diff
@@ -32,17 +33,30 @@ class ResourceMapper {
...
-   * @var \Drupal\Component\EventDispatcher\ContainerAwareEventDispatcher
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
    */
-  private $eventDispatcher;
+  private EntityTypeManagerInterface $entityTypeManager;
+
+  /**
+   * Entity storage service.
+   *
+   * @var \Drupal\Core\Entity\EntityStorageInterface
```
dafeder, line 47: Additional comment: The data used by the ResourceMapper is stored in ResourceMapping entities.

dafeder (commented):
Some suggestions for more explanatory code comments

dafeder (approved):

PR #4032, opened 2023-10-02
Use consistent naming conventions

janette:
## QA Steps

- [ ] Visit the dkan menu
- [ ] Confirm the link to metastore config is "Metastore settings" to be consistent with Datastore settings, and Data Dictionary settings.

### DIFF
--- a/modules/metastore/metastore.links.menu.yml
+++ b/modules/metastore/metastore.links.menu.yml
@@ -1,5 +1,5 @@
 dkan.metastore.config_properties:
-  title: Metastore configuration
+  title: Metastore settings
   route_name: dkan.metastore.config_properties
   description: Configure dataset properties for referencing sub-schemas and for HTML sanitization.
   parent: system.admin_dkan
--- a/modules/metastore/src/Form/DkanDataSettingsForm.php
+++ b/modules/metastore/src/Form/DkanDataSettingsForm.php
@@ -89,7 +89,7 @@ class DkanDataSettingsForm extends ConfigFormBase {
 
     $form['html_allowed_properties'] = [
       '#type' => 'checkboxes',
-      '#title' => $this->t('Properties that allow HTML'),
+      '#title' => $this->t('Dataset properties that allow HTML'),
       '#description' => $this->t('Metadata properties that may contain HTML elements.'),
       '#options' => $this->schemaHelper->retrieveStringSchemaProperties(),
       '#default_value' => $config->get('html_allowed_properties') ?:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,6 +96,7 @@ commands:
                   ddev composer update getdkan/dkan -W
                   ddev composer show getdkan/dkan
                   ddev drush updb -y
+                  ddev drush cr
                   ddev drush rq
       - unless:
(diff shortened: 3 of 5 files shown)


modules/metastore/src/Form/DkanDataSettingsForm.php
```diff
@@ -89,7 +89,7 @@ public function buildForm(array $form, FormStateInterface $form_state) {
 
     $form['html_allowed_properties'] = [
       '#type' => 'checkboxes',
-      '#title' => $this->t('Properties that allow HTML'),
+      '#title' => $this->t('Dataset properties that allow HTML'),
```
jastraat, line 92: Do we consider distribution properties other referenced entity properties also dataset properties? That was the reason I used just "properties"

modules/metastore/metastore.links.menu.yml
```diff
@@ -27,4 +27,4 @@ metastore.data_dictionary.settings_form:
   description: Data-Dictionary Settings
   parent: metastore.data_dictionary
   route_name: metastore.data_dictionary.settings
-  weight: 2
```
jastraat, line 30: If updating the menu, you'll need to add a post update hook.

modules/metastore/src/Form/DkanDataSettingsForm.php
dafeder: FWIW I think this is fine, on this page all properties are presented as

janette: Why are we listing every property, most of these should never have html allowed.

modules/metastore/metastore.links.menu.yml
janette: I think anyone updating code is familiar with clearing cache afterwards.

modules/metastore/src/Form/DkanDataSettingsForm.php
jastraat: We actually aren't listing every property - only string properties. But DKAN prides itself on swappable schemas, so we cannot hardcode assumptions about what the properties are.

janette: Not suggesting we hardcode anything, just be more specific where html could be useful. Can this config be put into the *.ui.json files?

dafeder: @janette that's an interesting idea. The thing is this would apply for harvests and API posts as well, so it's not exactly intuitive that such config would live in the ui schema... but in theory we could check it and allow HTML in any text area fields, or add a property to the UI schema like `allowHTML`. Hmmm....

jastraat: I would say this would be a different ticket; the one I was working stored the configuration as a Drupal configuration setting. You could re-write it to be part of the schema, but as Dan said it would need to work for harvests and API posts which don't really use the ui schema.

dafeder: Yeah. And I think "dataset" is appropriate here, we are showing the dataset and including other schemas within it but they are all really experienced as part of the dataset tree.

dafeder (approved):

PR #4033, opened 2023-10-02
Make PHP 8.1 + Drupal 10.1 the default in CI matrix tests

jastraat:
This updates the matrix tests to make PHP 8.1 and Drupal 10.1 the default configuration to test. This means the install and upgrade tests as well as the Cypress tests running now on that configuration. 

Also updated the Cypress tests to reflect a change (likely in VBO) where bulk view operations no longer trigger confirmation popups.


paul-m (approved):
LGTM.

PR #4034, opened 2023-10-05
Add justinrainbow/json-schema to composer.json

janette:
In the 2.16.10 release, _getdkan/json-schema-provider_ was removed. If you upgraded and were not building with dev dependencies including _drupal/core-dev_, you may have run into:

```
Uncaught PHP Exception Error: "Class "JsonSchema\Validator" not found" at /mnt/www/html/datamedicaidstg/docroot/modules/contrib/dkan/modules/datastore/src/Controller/AbstractQueryController.php line 344 request_id="v-b70ba20c-6395-11ee-a3c1-0bb257c19d59" 
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,6 +38,7 @@ commands:
         type: string
     steps:
       - run:
+          # TODO: We might not need to do this.
           name: Set up composer config
           command: |
             mkdir ~/.composer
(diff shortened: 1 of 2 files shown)


paul-m (approved):
Re-adds justinrainbow/json-schema which was previously added by getdkan/json-schema-provider. Also fixes adding a github oauth token to the CI.

PR #4035, opened 2023-10-06
FileFetcher should always reflect use-local-file config

paul-m:
Depends on https://github.com/GetDKAN/file-fetcher/pull/32

Since `FileFetcher` is an `AbstractPersistentJob`, any part of DKAN which creates a `FileFetcher` object will end up persisting that object within the database.

This means that if you were to check the status of a file being fetched (which will create a `FileFetcher` object), and then changed the configuration to always use local files, that configuration would be ignored on all subsequent calls to create a `FileFetcher` object for that file.

One solution to this would be to refactor the way such status items are checked. This requires a great deal of refactoring.

Instead, this PR subclasses `FileFetcher` and allows for changing the way the file is fetched to reflect the runtime configuration.


jastraat (commented):
These changes work when added to PDC, so no problem there.

dafeder (approved):

PR #4037, opened 2023-10-06
Issue #4036: Allow for an empty option in select fields via schema pr…

### LINKED ISSUE #4036: Allow for an empty option in select fields via schema property
## User Story
Currently you can not make use of the #empty_value property of the Drupal [Select](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21Element%21Select.php/class/Select) form element if using the basic select widget with the json_form_widget for a schema field.

You can use the select_other type for the list widget, but then you get the empty option and the other option.

There maybe cases where you only want the empty option but not allow the other option, i. e. an optional list field.

I propose to use a schema property on the dataset field to indicate this. Why not just name this "empty_value" and if this is set to true the #empty_value property will be used on the select form element.

One could maybe just inject an empty value in the "enum" property of the schema field, but this seems kind of "dirty".

stefan-korn:
fixes #4036 

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

- [ ] create or edit a dataset with an optional list (enum) field (like accrualPeriodicity) and note that there will be an empty option beside the real values in the select dialog
- [ ] if you save with the empty option the value should not exist in the json stored

### DIFF
--- a/modules/json_form_widget/src/StringHelper.php
+++ b/modules/json_form_widget/src/StringHelper.php
@@ -77,6 +77,9 @@ class StringHelper implements ContainerInjectionInterface {
     // Add options if element type is select.
     if ($element['#type'] === 'select') {
       $element['#options'] = $this->getSelectOptions($property);
+      if ($property->empty_value) {
+        $element['#empty_value'] = '';
+      }
     }
 
     // Add extra validate if element type is email.
--- a/modules/json_form_widget/src/StringHelper.php
+++ b/modules/json_form_widget/src/StringHelper.php
@@ -77,7 +77,7 @@ class StringHelper implements ContainerInjectionInterface {
     // Add options if element type is select.
     if ($element['#type'] === 'select') {
       $element['#options'] = $this->getSelectOptions($property);
-      if ($property->empty_value) {
+      if (!$this->checkIfRequired($field_name, $element_schema)) {
         $element['#empty_value'] = '';
       }
     }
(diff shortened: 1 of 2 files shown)


dafeder:
See my comment on the issue @stefan-korn, I think the proper behavior here is just to add an empty option in the select for any `enum` that is not required, and make sure the JSON that is produced does not contain the property.

modules/json_form_widget/src/StringHelper.php
```diff
@@ -77,6 +77,9 @@ public function handleStringElement($definition, $data, $object_schema = FALSE)
     // Add options if element type is select.
     if ($element['#type'] === 'select') {
       $element['#options'] = $this->getSelectOptions($property);
+      if ($property->empty_value) {
+        $element['#empty_value'] = '';
```
dafeder, line 81: Looks like we're going a different direction anyway, but not sure where `$property->empty_value` is supposed to be getting set?

dafeder (changes requested):

modules/json_form_widget/src/StringHelper.php
stefan-korn: I supposed it to be set in the schema .json file, just like "type", "format", etc. But yes now we are going a different and better way.

dafeder:
@stefan-korn do you feel up to updating the tests? I think you should be able to see the circleCI results and that the tests are all broken now... fairly trivial to update them to expect this and possibly to confirm the right result for both required and non-required fields, but if you're not familiar with PHPUnit let me know and we can help with that part.

stefan-korn:
I fixed the tests. The html output itself is not checked I suppose?

dafeder (approved):
Sorry I let this languish for so long @stefan-korn - just tested and it works perfectly.

PR #4038, opened 2023-10-09
Apply round of code quality improvements from Rector

paul-m:
Use Drupal-Rector to do the following:
- PHP 8.0 set list to add nice things like `str_starts_with()`, etc.
- Find ‘dead code’, which means unused variables, easy return short-circuits, etc.
- The Drupal 10 set list, which does things like ensure the $modules variable is protected in tests, etc.

Currently excludes `data_dictionary_widget` to avoid merge issues.


dafeder (approved):
This looks good. I think we'll hold off on merging till we have a 2.17 release ready, which will include PHP 8.2 support and dropping PHP 7.4

dafeder (approved):
Ready I think. Just one thing, I think the title/commit message is a little unclear. Can we make it something like "Apply round of code quality improvements from Rector" or something?

PR #4039, opened 2023-10-17
Documentation updates

janette:
fixes 
```
dkan/docs/source/user-guide/guide_data_dictionaries.rst:131: WARNING: Could not lex literal_block as "http". Highlighting skipped.
```


docs/source/user-guide/guide_harvest.rst
```diff
@@ -44,14 +44,18 @@ Example data.json file with various placeholders:
 The above example contains all the required properties for a dataset if using the default schema provided with DKAN. The
 default schema requires a distribution and at least one keyword. The identifier must be a unique string within the given
 catalog. You can find descriptions of the above properties and additional optional properties by viewing the example
-dataset schema that ships with DKAN at schema/collections/dataset.json.
+dataset schema that ships with DKAN at `schema/collections/dataset.json <https://github.com/GetDKAN/dkan/blob/2.x/schema/collections/dataset.json>`_.
+
+If you are using the harvest to bulk generate datasets, you should deregister the harvest after the datasets are created. If you plan
```
jastraat: I think we might want wait to put this into the docs. We still don't have a release of DKAN where deregistering the harvest doesn't already revert the imports.

docs/source/user-guide/guide_local_files.rst
```diff
@@ -77,51 +89,51 @@ We can pipe this to jq as well, to extract the specific path where DKAN expects
...
-    % drush dkan:datastore:prepare-localized 1fecf29222b12fc1ce2678abbc8f870f | jq -r .path
+    $ drush dkan:datastore:prepare-localized 1fecf29222b12fc1ce2678abbc8f870f | jq -r .path
     /var/www/html/docroot/sites/default/files/resources/1fecf29222b12fc1ce2678abbc8f870f_1691778866
 
 This Drush command, :code:`dkan:datastore:prepare-localized`, will add this file path information to the dataset map as well, which we can check by re-running our dataset info:
 
 .. code-block::
 
-    % ddev drush dkan:dataset-info bf215cd3-dd81-498c-b57a-4847dbeaac44 | jq -r '.latest_revision.distributions[].file_path'
+    $ ddev drush dkan:dataset-info bf215cd3-dd81-498c-b57a-4847dbeaac44 | jq -r '.latest_revision.distributions[].file_path'
```
jastraat: The other drush commands don't use ddev -

docs/source/user-guide/guide_harvest.rst
janette: Excellent point. I've made some edits but thinking maybe we just leave that last paragraph out altogether.

docs/source/user-guide/guide_local_files.rst
janette: Thanks, removed ddev. But also we don't use % or $ anywhere else, so we should confirm if this is something we need to update everywhere or if these prompt characters are out-dated and just leave those out as well.

jastraat (approved):
The two changes I requested earlier look good!

PR #4044, opened 2023-10-26
Allow select2 without entity reference (#4043)

### LINKED ISSUE #4043: Allow select2 without entity reference
## User Story
At this moment you can use a select2 widget for a schema property if you specify "list" for the widget and "autocomplete" for the type in {schema}.ui.json. This way the select2 widget is always tied to a node entity reference (target_type = node). This means this is only working for properties that are stored as separate entities.

There maybe cases where you have a lot of options (enum) for a property, but do not want to store this property as a separate entity, but still use the select2 widget for better UX. This can be easily achieved if we do not set the target_type for every select2 widget.

My proposal would be to use "select2" for the type in {schema}.ui.json and do not set the target_type in that case. That way one can still use the "autocomplete" type to get a select2 widget with entity reference, but also get a select2 widget without entity reference if desired.

## Acceptance Criteria
Allow a select2 widget for a property with enum values that is not stored as separate entity.

stefan-korn:
fixes #4043

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

- [ ] Use select2 instead of select_other for license in dataset.ui.json
- [ ] see that the select widget changes to select2 (providing search in select widget)

### DIFF
--- a/modules/json_form_widget/src/WidgetRouter.php
+++ b/modules/json_form_widget/src/WidgetRouter.php
@@ -185,7 +185,7 @@ class WidgetRouter implements ContainerInjectionInterface {
       $element['#multiple'] = isset($spec->multiple) ? TRUE : FALSE;
       $element['#autocreate'] = isset($spec->allowCreate) ? TRUE : FALSE;
     }
-    if (isset($element['#autocreate'])) {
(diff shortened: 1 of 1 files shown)


janette (approved):
Looks good @stefan-korn

PR #4046, opened 2023-10-26
If no time set after preview node set default time

### LINKED ISSUE #4045: When using the preview button, Date Modified requires a time to save
Search existing issues to make sure the bug has not been reported already.

## Describe the bug
When using the preview button and not setting a time for modified, you can see the preview and it has a time of 00:00:00 but when returning to the node edit for there is an error that a time is required. Adding a time is not required if the preview button isn't used and the node is saved first.

## Steps To Reproduce

1. Open a DKAN admin site with the preview button re-enabled.
2. Create a new dataset
3. Set just the date for the Modified field and all other required fields.
4. Before saving, click the preview button
5. On the preview page, click the back to dataset creation button
6. You should now see the an error that a time is required
7. Try to save the dataset and the error will prevent this action
8. Set a time and successfully save

As a control, on the same site create a dataset and add a modified date with no time and save the dataset. It should save successfully and now you can preview it and come back and the error will not appear.

## Expected behavior
A default time that is added during save is also added when using the preview button.

dgading:
fixes #4045

When using the Preview feature a `#default_value` is set when returning to the form. It seems like this is causing the error in the form. 

When saving the node or previewing the node, the `FlexibleDateTime:valueCallback` method correctly sets the time. When navigating back to the form, the default value for time was an empty string. I'm guessing this callback doesn't set form_state so checking to see if there is a default value and no time set, we can set the time the same as in the previous conditional.

### DIFF
--- a/modules/json_form_widget/src/Element/FlexibleDateTime.php
+++ b/modules/json_form_widget/src/Element/FlexibleDateTime.php
@@ -25,11 +25,8 @@ class FlexibleDateTime extends Datetime {
   /**
    * {@inheritdoc}
    */
-  public static function valueCallback(&$element, $input, FormStateInterface $form_state) {
-    if (empty($element['#default_value']) && !empty($input['date'])) {
-      $input['time'] = !empty($input['time']) ? $input['time'] : '00:00:00';
-    }
-    elseif (isset($element['#default_value']) && empty($input['time'])) {
+  public static function valueCallback(&$element, $input, FormStateInterface $form_state) {    
+    if (!empty($input["date"]) && empty($input['time'])) {
       $input['time'] = '00:00:00';
     }
     return parent::valueCallback($element, $input, $form_state);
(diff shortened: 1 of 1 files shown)


modules/json_form_widget/src/Element/FlexibleDateTime.php
```diff
@@ -26,8 +26,8 @@ public function getInfo() {
    * {@inheritdoc}
    */
   public static function valueCallback(&$element, $input, FormStateInterface $form_state) {
-    if (empty($element['#default_value']) && !empty($input['date'])) {
-      $input['time'] = !empty($input['time']) ? $input['time'] : '00:00:00';
+    if (!empty($input["date"]) && empty($input['time'])) {
```
jastraat: @dgading would we want to use single quotes for "date"?

jastraat (approved):
LGTM!

PR #4047, opened 2023-10-27
Use the latest published dataset revision when returning results for the datastore query endpoint

jastraat:
Use the latest published dataset revision when returning results for the datastore query endpoint. 
Also does some slight cleanup of the DatasetBTB test.

To recreate the error in a vanilla DKAN site:

- Create a published dataset with a distribution.
- Set the default moderation state to "draft".
- Edit the dataset and replace the distribution file with data that is recognizably different than the published version. Save as a draft.
- Run cron to do the imports
- Visit the dataset page in the site and confirm that the published version of the dataset is still displayed.
- Use the datastore query endpoint {domain}/api/1/datastore/query/{dataset}/{index} and observe that the data returned is from the unpublished draft, not the published version.

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

- [ ] Create a published dataset with a distribution.
- [ ] Set the default moderation state to "draft".
- [ ] Edit the dataset and replace the distribution file with data that is recognizably different than the published version. Save as a draft.
- [ ] Run cron to do the imports
- [ ] Verify that the draft revision exists by running ddev drush dkan:dataset-info {datasetID}
- [ ] Visit the dataset page in the site and confirm that the published version of the dataset is still displayed.
- [ ] Use the datastore query endpoint {domain}/api/1/datastore/query/{dataset}/{index} and observe that the data returned is still from the published version.

### DIFF
--- a/modules/common/src/DatasetInfo.php
+++ b/modules/common/src/DatasetInfo.php
@@ -136,7 +136,8 @@ class DatasetInfo implements ContainerInjectionInterface {
   }
 
   /**
-   * Get the distribution UUID for the most recent published revision of a dataset.
+   * Get the distribution UUID for the most recent published revision
+   * of a dataset.
    *
    * @param string $dataset_uuid
    *   The uuid of a dataset.
--- a/modules/common/src/DatasetInfo.php
+++ b/modules/common/src/DatasetInfo.php
@@ -136,9 +136,9 @@ class DatasetInfo implements ContainerInjectionInterface {
   }
 
   /**
-   * Get the current distribution UUID for a dataset.
+   * Get the distribution UUID for the most recent published revision of a dataset.
    *
-   * @param string $uuid
+   * @param string $dataset_uuid
    *   The uuid of a dataset.
    * @param string $index
    *   The index of the resource in the dataset array. Defaults to first.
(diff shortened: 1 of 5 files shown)


modules/common/src/DatasetInfo.php
```diff
@@ -135,6 +135,34 @@ public function gather(string $uuid) : array {
     return $info;
   }
 
+  /**
+   * Get the current distribution UUID for a dataset.
+   *
```
paul-m, line 143: How about adding a line here to explain the logic of what distribution id is returned based on published and draft scenarios?

modules/common/src/DatasetInfo.php
```diff
@@ -135,6 +135,34 @@ public function gather(string $uuid) : array {
...
+   * @param string $uuid
+   *   The uuid of a dataset.
+   * @param string $index
+   *   The index of the resource in the dataset array. Defaults to first.
+   *
+   * @return string
+   *   The distribution UUID
+   */
+  public function getDistributionUuid(string $uuid, string $index = '0'): string {
```
paul-m: Rename `$uuid` to `$dataset_uuid` please.

modules/common/src/DatasetInfo.php
```diff
@@ -135,6 +135,34 @@ public function gather(string $uuid) : array {
...
+    $metadata = $this->gather($uuid);
```
paul-m: Similarly, rename `$metadata` to `$dataset_info` to match the docs from `gather()`, and the naming used in the test below.

paul-m (changes requested):

paul-m (approved):

dafeder (approved):

PR #4053, opened 2023-11-07
#4051: Flexible DateTime widget: Small UX improvement when clearing date

### LINKED ISSUE #4051: Flexible DateTime widget: Small UX improvement when clearing date
## User Story
The flexible DateTime widget (flexible_datetime) is setting the time part to "00:00:00" if a date is chosen but no time. This is convenient and avoids annoying validation error message.

It would also be nice, to have the time part cleared when you clear the date part via the HTML5 widget, but only when the time part is 00:00:00 probably. I would propose to have the time part cleared the same way the "00:00:00" time part is set (in valueCallback).

## Acceptance Criteria
Clearing a date with time part "00:00:00" should allow to save and clear the time part while processing the value.

stefan-korn:
fixes #4051 

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

- [ ] Clearing a date with time part "00:00:00" should allow to save and clear the time part while processing the value. Can use Release Date in dataset for example.

### DIFF
--- a/modules/json_form_widget/src/Element/FlexibleDateTime.php
+++ b/modules/json_form_widget/src/Element/FlexibleDateTime.php
@@ -29,7 +29,7 @@ class FlexibleDateTime extends Datetime {
     if (!empty($input['date']) && empty($input['time'])) {
       $input['time'] = '00:00:00';
     }
-    if ($element['#date_date_element'] !== 'none' && empty($input['date'])) {
+    if (!empty($input['time']) && $element['#date_date_element'] !== 'none' && empty($input['date'])) {
       if ($input['time'] === '00:00:00') {
         unset($input['time']);
       }
--- a/modules/json_form_widget/src/Element/FlexibleDateTime.php
+++ b/modules/json_form_widget/src/Element/FlexibleDateTime.php
@@ -29,6 +29,11 @@ class FlexibleDateTime extends Datetime {
     if (!empty($input['date']) && empty($input['time'])) {
       $input['time'] = '00:00:00';
     }
+    if ($element['#date_date_element'] !== 'none' && empty($input['date'])) {
+      if ($input['time'] === '00:00:00') {
+        unset($input['time']);
+      }
+    }
     return parent::valueCallback($element, $input, $form_state);
   }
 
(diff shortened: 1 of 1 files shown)


dafeder:
Looks good @stefan-korn technically missing tests but i don't think that widget has test coverage to begin with so I won't consider it a blocker.

The thing codeclimate is complaining about is this unnecessary nested conditional:

```php
    if (!empty($input['time']) && $element['#date_date_element'] !== 'none' && empty($input['date'])) {
      if ($input['time'] === '00:00:00') {
```

At least we should combine all three conditions into a single if, but I think two of them are redundant anyway, so I think this should produce the same result:

```php
    if (($input['time'] ?? FALSE === '00:00:00') && $element['#date_date_element'] !== 'none' && empty($input['date'])) {
```

dafeder (changes requested):
Looks good @stefan-korn technically missing tests but i don't think that widget has test coverage to begin with so I won't consider it a blocker.

The thing codeclimate is complaining about is this unnecessary nested conditional:

```php
    if (!empty($input['time']) && $element['#date_date_element'] !== 'none' && empty($input['date'])) {
      if ($input['time'] === '00:00:00') {
```

At least we should combine all three conditions into a single if, but I think two of them are redundant anyway, so I think this should produce the same result:

```php
if (
  ($input['time'] ?? NULL === '00:00:00') 
  && $element['#date_date_element'] !== 'none' 
  && empty($input['date'])
) {
```

stefan-korn:
@dafeder : Okay, I changed the condition like you proposed. I am totally having a hard time understanding this right away

> $input['time'] ?? NULL === '00:00:00'

and assume that others might have too. So sometimes I would prefer easy code readability over "code climate", but this is surely opinionated.

dafeder:
Replacing with #4157

dafeder:
> @dafeder : Okay, I changed the condition like you proposed. I am totally having a hard time understanding this right away
> 
> > $input['time'] ?? NULL === '00:00:00'
> 
> and assume that others might have too. So sometimes I would prefer easy code readability over "code climate", but this is surely opinionated.

Sorry @stefan-korn going to override you on this one :). The null coalesce operator is increasingly encouraged in PHP and can be found throughout DKAN. I recognize it's subjective, but IMO once you've seen it a few times it becomes intuitive and improves the code flow a lot.

PR #4055, opened 2023-11-09
#4054: proposal for saving embedded references

### LINKED ISSUE #4054: Hashes not updated when modifying sub-schema items (publishers, etc) directly
## Describe the bug
Editing an embedded distribution inside a dataset (embedded distribution in dataset is default in DKAN) creates a new distribution node every time something is changed in the distribution. Is this intended behavior? Would not think so?

The reason for this is as far as I can see, that the function checkExistingReference() in Referencer class searches for the distribution title and that is a hash of all the values of the distribution. Now if the values have changed the hash is no longer the same and therefore checkExistingReference() fails and a new distribution is created.

This does not solely affect distributions but would affect any other referenced item that can be edited inside another.

Sorry for maybe asking some stupid questions:
Why is this hash of the data used as a title for a referenced node?
Couldn't one maybe use UUID as title for referenced nodes? The reference is linked by UUID as identifier anyway?

## Steps To Reproduce
Create a dataset with a distribution.
check admin/dkan/datasets for distributions and see the distribution appearing once
Edit the created dataset and changes something in the distribution
check admin/dkan/datasets for distributions again and see there another distribution that has been created

## Expected behavior
Do not create a new distribution on editing a distribution inside a dataset.

stefan-korn:
proposal for different handling of embedded elements (i. e. distribution in dataset). Just a proof of concept, surely not ready for production at this moment. see #4054 

- [ ] Test coverage exists
- [ ] Documentation exists

### DIFF
--- a/modules/metastore/src/LifeCycle/LifeCycle.php
+++ b/modules/metastore/src/LifeCycle/LifeCycle.php
@@ -318,7 +318,7 @@ class LifeCycle {
       return $data instanceof MetastoreItemInterface;
     });
 
-    $metadata = $this->referencer->reference($metadata);
+    $metadata = $this->referencer->reference($metadata, $data->getRawMetadata());
 
     $data->setMetadata($metadata);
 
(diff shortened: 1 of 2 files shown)


dafeder:
Can you explain the concept here a bit more? I don't totally grok the change. See my explanation in the comments of #4054; it's a known problem but not one with an easy fix without causing regressions.

stefan-korn:
@dafeder : The change was about saving the uuid as reference and not the hash on the values. I have tested a while with this change and came across problems that were introduced with these changes. So I am closing this merge request.

PR #4060, opened 2023-11-15
Create a search page using Drupal views

tiffneybare:
- 15261
- 12919

## QA Steps

- [ ]  Checkout the `metastore_search_view_2` branch in your local DKAN
- [ ] drush en sample_content metastore_facets
- [ ] drush dkan:sample-content:create
- [ ] drush cron
- [ ] drush cim --partial --source=modules/contrib/dkan/modules/metastore/modules/metastore_search/config/install
- [ ] Change the default theme to olivero
- [ ] Go to /admin/structure/block
- [ ] Place Categories, Tags, and Publishers blocks into the sidebar region, limit display to the /dataset-search page and save.
- [ ] Go to /dataset-search
- [ ] Confirm you see the 10 sample datasets
- [ ] Confirm the results change as expected when facet checkboxes are checked
- [ ] Confirm the dataset titles link to the dataset page

## Notes
This introduces a breaking change to the search api config so existing sites will need to import the config manually IF they have not altered the search api config. And if this feature is desired on a site that has modifications to the dkan search api, they will need to be reconciled first with the changes this branch is adding.


janette:
~~This looks good, I don't think we should merge until we add a twig template for the results list.~~

modules/metastore/modules/metastore_facets/composer.json
```diff
@@ -0,0 +1,34 @@
+{
```
dafeder: We need this to be added to the DKAN composer.json, this won't get read be composer for a project

modules/metastore/modules/metastore_search/config/install/search_api.index.dkan.yml
```diff
@@ -26,6 +26,11 @@ field_settings:
     datasource_id: dkan_dataset
     property_path: modified
     type: string
+  nid:
```
dafeder, line 29: I think we need an update hook to do the config import task automatically as well, you shouldn't have to run `drush cim` when you enable a module.

dafeder (changes requested):
Looks great overall. See two changes.

dafeder (commented):
Another thing  -- maybe to discuss. How would one adapt this if they changed the schema or wanted different facets then those provided here - would you need to disable this module and write your own? may need more documentation

modules/metastore/modules/metastore_facets/composer.json
janette: This is an optional module that is available but not enabled by default, I think the follow up ticket is to create a second ddev site install command for a standing up a non-decoupled FE site. Or am I missing your point here?

dafeder (approved):

PR #4066, opened 2023-11-20
Re-reference distributions on presave

jastraat:
fixes [org/repo/issue#]

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

- [ ] Create a vanilla DKAN site with the sample content.
- [ ] Disable the default data dictionary if present. 
- [ ] Check the resource settings. ( admin/dkan/resources ) Resource download display should be 'source'
- [ ] Set the default dkan moderation state to draft. ( admin/config/workflow/workflows/manage/dkan_publishing )
- [ ] Create a dataset that has a distribution. (Can use a link to a dkan demo site CSV). Save as draft.
- [ ] Take node of the node IDs for the draft dataset and the draft distribution.
- [ ] Run cron: ddev drush cron
- [ ] View the contents of the node__field_json_metadata field for both the dataset and the distribution nodes. e.g. "SELECT field_json_metadata_value FROM node__field_json_metadata where entity_id = [nid]" for each
- [ ] No "%Ref:" arrays should be present in the json in the DB. All referenced entities should simply have a value of the uuids of the nodes they are referencing. The distribution downloadURL value should be the reference to the resource. (resourceID__version__perspective)
- [ ] On the dataset admin view (/admin/dkan/datasets), check the draft dataset and publish the latest revision.
- [ ] View the contents of the node__field_json_metadata field for both the dataset and the distribution nodes again. 
- [ ] No "%Ref:" arrays should be present in the json in the DB. All referenced entities should simply have a value of the uuids of the nodes they are referencing. The distribution downloadURL value should be the reference to the resource. (resourceID__version__perspective)
- [ ] Confirm that the dataset published and displays correctly.

### DIFF
--- a/modules/metastore/src/LifeCycle/LifeCycle.php
+++ b/modules/metastore/src/LifeCycle/LifeCycle.php
@@ -340,10 +340,30 @@ class LifeCycle {
   }
 
   /**
-   * Private.
+   * Distribution presave.
+   *
+   * @param \Drupal\metastore\MetastoreItemInterface $data
+   *   Dataset metastore item.
    */
-  protected function distributionPresave(MetastoreItemInterface $data) {
+  protected function distributionPresave(MetastoreItemInterface $data): void {
     $metadata = $data->getMetaData();
+
+    if (!$data->isNew()) {
+      $distributionUuid = $data->getIdentifier();
+      $storage = $this->dataFactory->getInstance('distribution');
+      $resource = $storage->retrieve($distributionUuid);
+      $resource = json_decode($resource);
+
+      $resourceId = $resource->data->{'%Ref:downloadURL'}[0]->data->identifier ?? NULL;
+
+      // Replace download url with the resource reference ID again.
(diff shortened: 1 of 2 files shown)


dafeder (changes requested):
@jastraat we hit the code complexity gate because of adding a conditional nested inside another one, and actually in this case I don't think the conditional is needed at all. There is no error thrown if you unset a non-existent property. So in `Referencer::reference()` we can just call unset without checking if the Ref thing is there.

jastraat:
Updated to remove the check to see if the properties were set before unsetting -

dafeder (changes requested):
This looks good, but as discussed we are going to hold off on merge until we at least have a functional test covering the changes to LifeCycle.

jastraat:
I'm going to go ahead and close this PR. All the changes in this PR have been included in the more comprehensive PR: https://github.com/GetDKAN/dkan/pull/4002 along with 5 new functional tests.

PR #4069, opened 2023-11-27
#4068: Provide number widget

### LINKED ISSUE #4068: Number widget
## User Story
Would be nice to have a number widget available for specific schema properties.

Proposal:
```
"ui:options" {
  "widget": "number",
  "step": "0.01",
  "min": "10",
  "max": "20"
}
```
step, min and max are optional.

stefan-korn:
fixes #4068

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps


janette:
@stefan-korn this would need test coverage before it can be merged.

dafeder:
@stefan-korn are you able to add some coverage? The easiest thing would probably be to add a number field to `Drupal\Tests\json_form_widget\Unit\JsonFormBuilderTest`

modules/json_form_widget/src/WidgetRouter.php
```diff
@@ -410,4 +411,33 @@ public function handleDateRangeElement($spec, array $element) {
...
 
+  /**
+   * Helper function for getting a number element.
+   *
+   * @param mixed $spec
+   *   Object with spec for UI options.
+   *   optional specs
+   *   - step
+   *   - min
+   *   - max
```
dafeder:
```suggestion
   * @param object $spec
   *   Specification for UI options. Optional properties:
   *   - step: Ensures that the number is an even multiple of step.
   *   - min: Minimum value.
   *   - max: Maximum value.
```

modules/json_form_widget/src/WidgetRouter.php
```diff
@@ -410,4 +411,33 @@ public function handleDateRangeElement($spec, array $element) {
...
+   * @param array $element
+   *   Element to convert into number.
+   *
+   * @return array
+   *   The element configured as number.
+   */
```
dafeder:
```suggestion
   * @return array
   *   The element configured as number.
   *
   * @see \Drupal\Core\Render\Element\Number
   */
```

dafeder (changes requested):
I think this will satisfy codeclimate about code/doc standards

modules/json_form_widget/src/WidgetRouter.php
```diff
@@ -410,4 +411,33 @@ public function handleDateRangeElement($spec, array $element) {
...
+  public function handleNumberElement($spec, array $element) {
+    $element['#type'] = 'number';
+    if (isset($spec->step)) {
+      $element['#step'] = $spec->step;
+    }
+    if (isset($spec->min)) {
+      $element['#min'] = $spec->min;
+    }
+    if (isset($spec->step)) {
```
dafeder: @stefan-korn shouldn't this be `$spec->max`?

dafeder (changes requested):
@stefan-korn have given this another review. See my one comment here, I was able to cause an error by giving the ui schema a step but not a max.

I'd like to get this in, but I think it needs a little more work. Beyond the comments in this and previous reviews, I think it should also work to define a field as "number" directly in the schema, with no additional definition in the UI schema. This may mean adding a new "helper" class, or else consolidating "number" and "integer" since they ultimately use the same field type.

stefan-korn:
@dafeder : I addressed your comments.

Regarding "number" in the schema. I wasn't aware that there is the integer field. It is only used once in the data-dictionary.json I suppose and we do not need data dictionary right now, so I missed that.

Should the schema be configurable about step, min and max? That way no widget "number" would be needed at all?

dafeder:
@stefan-korn re-thinking this I think the "number" field in the schema itself can be dealt with separately, should not be a blocker for this. I did a new PR from this that hits the test coverage, closing this in favor of #4215. Thanks!

PR #4076, opened 2023-12-04
#4075: Catch Guzzle Exception to avoid breaking harvest

### LINKED ISSUE #4075: Catch Guzzle Exception to avoid breaking harvest
## Describe the bug
During harvesting the method getRemoteMimeType is called to determine the mime type of distribution resource by calling the downloadURL. If for some reason the downloadURL is not available or broken, the harvesting will fail alltogether with a Guzzle Exception logged.

## Steps To Reproduce
Do a harvest with one resource download url not being available.

## Expected behavior
The method getRemoteMimeType should retun NULL on failure to call the downloadURL (as it is already said in the methods declaration) instead of quitting with an Exception.

stefan-korn:
fixes #4075

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

### DIFF
--- a/modules/metastore/src/Reference/Referencer.php
+++ b/modules/metastore/src/Reference/Referencer.php
@@ -13,6 +13,7 @@ use Drupal\metastore\MetastoreService;
 use Drupal\metastore\ResourceMapper;
 
 use GuzzleHttp\Client as GuzzleClient;
+use GuzzleHttp\Exception\GuzzleException;
 
 /**
  * Metastore referencer service.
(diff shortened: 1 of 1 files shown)


janette:
@stefan-korn thanks for bringing up this issue. We will want to check access to the URL before we get to the check mimetype step. And if a URL is not accessible, send a warning message to the user. So once we locate the best spot for that check (filefetcher, harvest json source validation, etc.) then we can re-work the code and tests, and merge a fix.

paul-m:
Merging into a feature branch to expedite working on this.

Thanks @stefan-korn !

paul-m:
Work continues: https://github.com/GetDKAN/dkan/pull/4084

PR #4079, opened 2023-12-08
Update dataset template to display page layout when view_mode == full

janette:
## QA Steps

- [ ] Create a dataset and save
- [ ] Edit the dataset and save so there is a revision
- [ ] Click edit again and then click the revisions tab
- [ ] Select the revision and confirm it displays the full dataset page and not just the title and description.


jastraat (approved):
LGTM!

PR #4085, opened 2023-12-20
Support PHP 8.2

paul-m:
This PR mainly does the following:
- Use PHP 8.2-compatible versions of dependencies in `composer.json`.
- Add PHP 8.2 to the testing matrix.

We then use a local dev environment and CI test results to adjust the following categories of changes to avoid deprecation messages:
- Explicitly declare class properties rather than using dynamic properties.
- Annotate a return type on `JsonSerializable::jsonSerialize()`.

We also take the opportunity to update `rector.php`. This brings our codebase up to snuff for PHP 7.4, and conservatively ends up making these changes:
- Ternary to NULL-Coalescing operators
- `$this->staticMethod()` to `self::staticMethod()` conversions.
- `list()` to `[$item, $item2]` conversion.
- Fully-qualified class name to imported class name conversion.

It's important to note that if you look at the test results for the PHP 8.2 environment in this CI, you'll see deprecation messages like this:
```
Deprecated: Creation of dynamic property Drupal\Core\TypedData\TypedData@anonymous::$value is deprecated in /var/www/html/docroot/core/lib/Drupal/Core/TypedData/TypedData.php on line 102
```
This is a known issue for Drupal core, seen here: https://www.drupal.org/project/drupal/issues/3368250

This led to a 'workaround' PR to fix a test: https://github.com/GetDKAN/dkan/pull/4104


dafeder (approved):
Looks great! Awesome work everyone

PR #4086, opened 2023-12-28
Documentation updates

janette:
- Move remote image into the repo and update the link
- Move custom schema info to the user guide section
- Add harvest commands and info about them to the harvest guide


docs/source/user-guide/guide_harvest.rst
```diff
@@ -81,11 +81,37 @@ Run the harvest
...
 
+Revert a harvest
+----------------
+
+  .. prompt:: bash $
+
+    drush dkan:harvest:revert myHarvestId
+
+This will delete all of the datasets listed in the specified harvest ID. Any referenced
+terms will be set to the orphaned state. And distributions from the harvest will be unpublished.
```
rovcase: Is this supposed to be "Any referenced terms will be set to the orphaned state and distributions from the harvest will be unpublished." or "Any referenced terms will be set to the orphaned state. Any distributions from the harvest will be unpublished."?

rovcase (changes requested):
Looks good just one call out for a typo in case you wanted to change it before merge.

docs/source/user-guide/guide_harvest.rst
janette: @rovcase I think it works either way, if you think it reads better with 'Any' I can update it.

rovcase: Nope- works for me if it works for you, thanks!

rovcase (approved):
LGTM, thanks!

PR #4087, opened 2023-12-28
Replace ltrim with str_replace in json_form_widget

janette:
An `ltrim` followed by a second parameter will trim any matching contiguous character(s) from the left side of the email address. e.g. matildafoo@test.com will become dafoo@test.com

Switching to `str_replace` will result in only removing an exact match on "mailto:"

## QA Steps

- [ ] Create a dataset, fill in the ContactPoint email field with "matildafoo@test.com", and save
- [ ] view or edit the dataset to confirm the value was saved correctly


tiffneybare (approved):
Looks good locally

PR #4090, opened 2024-01-09
Use literal identifier for FileFetcherJobStoreFactory

paul-m:
Following the work of https://github.com/GetDKAN/dkan/pull/4083

Adds a `FileFetcherJobStoreFactory` class, which subclasses `AbstractJobStoreFactory`.

`FileFetcherJobStoreFactory` uses a literal table name of jobstore_524493904_filefetcher.

Adds this new class as a service: `dkan.common.filefetcher_job_store_factory`.

Changes service definitions which use `dkan.common.job_store` to use `dkan.common.filefetcher_job_store_factory`.

Changes all the services to accept `FileFetcherJobStoreFactory` instead of `JobStoreFactory`.

Modifies tests to use the new service.

### DIFF
--- a/modules/common/common.services.yml
+++ b/modules/common/common.services.yml
@@ -27,8 +27,13 @@ services:
   dkan.common.file_fetcher:
     class: \Drupal\common\FileFetcher\FileFetcherFactory
     arguments:
-      - '@dkan.common.job_store'
+      - '@dkan.common.filefetcher_job_store_factory'
       - '@config.factory'
+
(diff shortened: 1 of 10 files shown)


dafeder:
Tested from upgrade perspective; imported, switched to branch, confirmed datastore and dashboard statuses still match.

jastraat (commented):
Tested this out with PDC and did not see any issues.

dafeder (approved):
Code looks good, want to check behavior of datastore and import status dashboard after upgrade

PR #4099, opened 2024-01-19
Drupal 10.2 compatibility additions, exclude PHP 7 & Drupal 9 from testing matrix

jastraat:
Quick note that I'm using str_contains() so this won't be PHP 7.4 compatible.

Includes:

- Adjusts the DB driver check to be compatible with 10.2
- Removes PHP 7.4 and Drupal 9.x from the test matrix.
- Drupal 10 requires at least PHP 8.1 so dropping Drupal 9.x means dropping PHP < 8.1
- Adds Drupal 10.2 to test matrix with PHP 8.1 and PHP 8.2 (Currently expected to fail PHP 8.2 tests right now)

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -247,7 +247,7 @@ workflows:
          matrix:
            parameters:
              dkan_recommended_branch: [ '10.0.x-dev']
-             php_version: [ '8.0', '8.1' ]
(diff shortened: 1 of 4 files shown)


paul-m (approved):
LGTM

PR #4104, opened 2024-01-25
QueryDownloadControllerTest throws deprecation error

paul-m:
Under PHP 8.2, deprecation messages related to the `$dataDictionaryFields` property sneak their way into the output buffer and throw off the test results.

- Addresses all deprecation errors related to `\Drupal\Tests\datastore\Unit\Controller\QueryDownloadControllerTest`.
  - `queryResultCompare()` doesn't need to add `dataDictionaryFields` to the stream response.
  - Deprecated `['self', 'method']` callables changed to `[self::class, 'method']` style.
- Removes the unused `queryResultReformatted()` method.
- Removes the unused `getSelectFactory()` method and associated code in `__construct()`.

PHP 8.2 is not in the testing matrix because we need to fix this test before we can merge https://github.com/GetDKAN/dkan/pull/4085

To review this PR:
- Make a local dkan site with ddev.
- Change the PHP version to 8.2.
- `ddev composer update`
- `ddev dkan-phpunit --filter QueryDownloadControllerTest`
- You should see no deprecation messages and all tests should pass.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -249,11 +249,6 @@ workflows:
               dkan_recommended_branch: [ '10.1.x-dev']
               php_version: [ '8.1' ]
       - phpunit:
-          matrix:
-            parameters:
-              dkan_recommended_branch: [ '10.1.x-dev']
-              php_version: [ '8.2' ]
-      - phpunit:
          matrix:
            parameters:
              dkan_recommended_branch: [ '10.0.x-dev']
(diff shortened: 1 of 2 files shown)


dafeder (approved):
A bit concerned that the test still passes without dataDictionaryFields. I mean the way the property was being added it was never doing anything so it makes sense removing it has no effect -- it needed to be added to the JSON content not as an arbitrary property on the HTTP response object. So presumably there is some feature of data dictionaries lacking test coverage that we thought we had. I think for now it probably just is what it is though.

PR #4110, opened 2024-02-07
Convert harvest_id_hashes tables to Drupal entity API

paul-m:
- Create an entity called harvest_hash.
- Harvest hashes will have the following base fields: data_uuid (this is a node UUID), harvest_plan_id, hash
- Use a 'repository' type pattern to act as a facade between Drupal entity API and the DatabaseTableInterface API. This repository will implement DatabaseTableInterface.
  - The repository will accept a harvest plan identifier as a construction argument (normal for DatabaseTableInterface objects), and will use this harvest plan id as a condition on all the queries it performs (not normal DatabaseTableInterface behavior). This way all the different hash tables will share the same entity table in the database. Restated: The factory will create a different repository object per harvest plan ID. Since the tables have been flattened to one, all repository objects will need to add the constraint of their harvest plan ID to queries so that they can act as if they are dealing with a separate table per plan ID.
  - We have to do this because we need to give a DatabaseTableInterface object to Harvest\Harvester, so that it can perform harvests. We will refactor this later, as the external harvest library is factored away.
- Create a database table factory service class consistent with the rest of DKAN, which can produce a DatabaseTable object which is this entity's repository.
- Create a tool to convert harvest_id_hashes tables into this new entity table on update, and/or using Drush, etc.


modules/harvest/harvest.permissions.yml
```diff
@@ -17,3 +17,6 @@
 administer harvest plan:
   title: 'Administer harvest plan settings'
   restrict access: true
+administer harvest hash:
```
dafeder: Unused, remove

dafeder (changes requested):

dafeder (approved):

PR #4133, opened 2024-02-22
#4132: Provide text_format render element

### LINKED ISSUE #4132: Provide text_format render element for string schema properties
## User Story
Editors want to use [text_format](https://api.drupal.org/api/drupal/core%21modules%21filter%21src%21Element%21TextFormat.php/class/TextFormat/10) render elements for long text fields (currently designated via textarea widget in schema.ui.json), for example to edit long text in CKEditor.

## Acceptance Criteria
- provide option "text_format" in schema.ui.json to indicate the text_format which should be used for the long text property.
- show "text_format" editing in JSON form widget with the indicated format.

## Caveats
It is not (easily) possible to store the format together with the string in json. Therefore I would propose to not store the format in the JSON and keep this fixed in the schema.ui.json. So it will be allowed only one format (via allowed_formats) in the text format widget and no changing is possible via UI.
The output of the string needs to be taken care independently. For example currently the [node--data template](https://github.com/GetDKAN/dkan/blob/6850890d43aec65117b5e56335254bb4073b0910/modules/metastore/templates/node--data.html.twig#L109) escapes any HTML in the string (though I suppose it is not even necessary here, because Twig does escaping itself).
If you want to output HTML in the description you need to change this and take care yourself. You also need to ensure that only the HTML you want is output, so maybe custom processing is needed (but this is done anyway via metastore_preprocess_node__data), so you can do it like there for yourself.

stefan-korn:
fixes #4132 

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps


stefan-korn:
Hm, too many return statements in code climate ... seems 4 was acceptable, but 5 are not :-)

Should one switch to "switch" then?

dafeder (changes requested):
Picking this back up - I can see where this rule is coming from in the coding standards. It makes it a little harder to read for there to be so many possible returns. I believe if you just have it just set a variable on each condition (call it `$string` or `$return` or whatever, and just return that at the end, you should be fine. You may get hit on code complexity for having so many conditions but I am willing to override that in codeclimate since it makes sense to handle this widget that way IMO. 

What we will need to bring this in is, you guessed it, text coverage. You were able to provide good test imputs and assertions for what became #4212 do you think you could do the same here? Overall I very much agree with the approach you're taking here -- set it once in the UI schema and do not give the option on the form.

dafeder:
Closing in favor of #4371

PR #4135, opened 2024-02-22
#4134: Datasets view should only list nodes of type Data

### LINKED ISSUE #4134: View DKAN Metastore (Datasets) should only list nodes of type Data
## Describe the bug
I suppose that the view DKAN Metastore (Datasets) (dkan_dataset_content) should only list nodes of type "Data". Currently if you set filter "Data type" to "any" nodes of types other than "Data" will be listed as well.

## Steps To Reproduce
Create a node of a different type than "Data"
Go to DKAN Metastore (Datasets) at admin/dkan/datasets and set "Data Type" filter to "Any"
see that the node you created is listed

## Expected behavior
do not list nodes other than type "Data"

stefan-korn:
fixes #4134 

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

### DIFF
--- a/modules/json_form_widget/tests/src/Functional/AdminDatasetFileUploadTest.php
+++ b/modules/json_form_widget/tests/src/Functional/AdminDatasetFileUploadTest.php
@@ -26,6 +26,11 @@ class AdminDatasetFileUploadTest extends BrowserTestBase {
   protected $defaultTheme = 'stark';
 
   /**
+   * @todo Remove this when we drop support for Drupal 10.0.
+   */
+  protected $strictConfigSchema = FALSE;
+
+  /**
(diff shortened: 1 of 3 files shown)


stefan-korn:
I suppose the test failed because it uses Drupal 10.0.x and the "settings.tooltip" was introduced to the schema in 10.1.x.

paul-m:
Nice catch. Add the following to those two tests, and they'll pass:
```
  /**
   * @todo Remove this when we drop support for Drupal 10.0.
   */
  protected $strictConfigSchema = FALSE;
```

paul-m (approved):
Manually tested and the changes fix the issue.

PR #4143, opened 2024-03-08
Use original headers on csv downloads

janette:
Replaces machine names with original CSV headers in CSV download.

## QA Steps

- On a demo site, try the following request:

```http
POST https://csvdown.ddev.site/api/1/datastore/query/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/0

{
    "format": "csv",
    "properties": [
        "roadway"    
    ],
    "limit": 10
}
```
Note that the column header is ROADWAY.

- Now try with an expression:

```http
POST https://csvdown.ddev.site/api/1/datastore/query/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/0

{
    "format": "csv",
    "properties": [
        "roadway",
        {
            "expression": {
                "operator": "+",
                "operands": [
                    "objectid",
                    1
                ]
            },
            "alias": "objectid_plus_1"
        }
    ],
    "limit": 10
}
```
Note that ROADWAY is shown in caps, and alias is preserved.

If you want to try a join, try this but you will need to find the correct distribution IDs:

```http
POST https://csvdown.ddev.site/api/1/datastore/query

{
    "format": "csv",
    "resources": [
        {
            "id": "5733123b-6b70-5ff2-8876-c4e3a7c58398",
            "alias": "t"
        },
        {
            "id": "a180fe3c-e88d-55d9-841f-912ab0c4362d",
            "alias": "s"
        }
    ],
    "joins": [
        {
            "resource": "s",
            "condition": {
                "resource": "t",
                "property": "record_number",
                "value": {
                    "resource": "s",
                    "property": "record_number"
                }
            }
        }
    ],
    "properties": [
        "roadway",
        {
            "resource": "s",
            "property": "city"
        },
        {
            "resource": "s",
            "property": "violent_crimes"
        } 
    ],
    "limit": 10
}
```

## Known issues

* ~If `schema` property in request is set to `false`, the machine names will be shown. This is arguably expected behavior?~ The "schema" property in the request will be ignored if the format is "csv", and forced to true behind the scenes.
* In the case of a join, the properties from the second resource will have their machine names. Decided this was out of scope for this PR and will fix in the future.
* If "rowIds" is set to TRUE, adding the record_number column, we will still return machine names for all other columns as well.
* In general a lot of the CSV header logic relies on information pulled from the query or results objects which seems less than ideal.

### DIFF
--- a/modules/datastore/src/Controller/QueryController.php
+++ b/modules/datastore/src/Controller/QueryController.php
@@ -3,8 +3,8 @@
 namespace Drupal\datastore\Controller;
 
 use Drupal\datastore\Service\DatastoreQuery;
-use RootedData\RootedJsonData;
 use Ilbee\CSVResponse\CSVResponse;
+use RootedData\RootedJsonData;
 use Symfony\Component\HttpFoundation\ParameterBag;
 
 /**
@@ -17,9 +17,9 @@ class QueryController extends AbstractQueryController {
   /**
    * Return correct JSON or CSV response.
    *
-   * @param Drupal\datastore\Service\DatastoreQuery $datastoreQuery
+   * @param \Drupal\datastore\Service\DatastoreQuery $datastoreQuery
    *   A datastore query object.
-   * @param RootedData\RootedJsonData $result
+   * @param \RootedData\RootedJsonData $result
    *   The result of the datastore query.
    * @param array $dependencies
    *   A dependency array for use by \Drupal\metastore\MetastoreApiResponse.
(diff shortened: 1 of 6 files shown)


modules/datastore/src/Controller/QueryController.php
```diff
@@ -46,6 +47,28 @@ public function formatResponse(
...
+   * Add the header row from specified properties or the schema.
+   *
+   * Alters the data array.
+   *
+   * @param \Drupal\datastore\Service\DatastoreQuery $datastoreQuery
+   *   A datastore query object.
+   * @param \RootedData\RootedJsonData $result
+   *   The result of the datastore query.
+   */
+  private function fixHeaderRow(DatastoreQuery $datastoreQuery, RootedJsonData &$result) {
```
paul-m: Can we rename this method to `useCsvHeader` or something? 'Fix' is a little vague.

modules/datastore/src/Service/Query.php
```diff
@@ -220,4 +216,25 @@ private function runCountQuery(DatastoreQuery $datastoreQuery) {
     return (int) $storageMap[$primaryAlias]->query($query, $primaryAlias)[0]->expression;
   }
 
+  /**
+   * Under most circumstances, we want an explicit list of properties.
+   *
+   * @param \Drupal\datastore\Service\DatastoreQuery $datastoreQuery
+   *   Datastore query object to be modified.
+   */
+  private function fixProperties(DatastoreQuery $datastoreQuery) {
```
paul-m: Again, maybe have a better name than 'fix.' Not sure what's fixed, etc.

paul-m (changes requested):
Looks like everything works... I exported and it used the CSV column names rather than the table column names. Just a few nitpicky things here about method names.

paul-m (approved):

PR #4144, opened 2024-03-13
19105 Use Drupal entity API for harvest_runs table, with refactor

paul-m:
- Creates an entity called `harvest_run`. This entity has the following features:
  - Various UUID-oriented fields from the JSON data blob are normalized into the following base fields with unlimited cardinality:
    - `extracted_uuid`
    - `load_new_uuid`
    - `load_updated_uuid`
    - `load_unchanged_uuid`
    - `orphan_uuid`
  - This allows for instance doing a normalized query to find which run created which dataset, etc.
  - Data not stored in entity base fields remains in the JSON blob, which is written to a column called `data`.
- Many calls to `HarvestService::getAllHarvestRunInfo()` only need the IDs, so add `HarvestService::getAllHarvestRunIds()` so we can just do an entity query for IDs without the overhead of loading entities or parsing JSON.
- Do entity conversion on update.
- Add `HarvestRunRepository` for ease of calling between `HarvestService` and `HarvestRun` entity methods.
- Some refactor of harvest Drush commands
- Update `HarvestService` to use new run repo.
- Update `HarvestUtility` for the new run repo.
- Update tests to use run entity/repo.


dafeder (approved):
I think we have a winner

PR #4146, opened 2024-03-14
Fix sample_content

paul-m:
### Changes: 

- Allows for test coverage of `sample_content`.
- Requires `drush/drush` as dev dependency, so we can test drush runs.
- `Api1TestBase` erroneously used `sample_content.json` to perform functional API tests. `sample_content.json` is generated to account for the real paths to the resource files, so it only works as a test fixture by accident.
- Since we need that `sample_content.json` file to exist for the tests, and also to be generated for adding sample content for the user, we create a service called `SampleContentService` (AKA `dkan.sample_content.service`). This is used in both places.
- Adding the `SampleContentService` class means we can simplify the Drush command implementations, using both `SampleContentService` and `HarvestService`.
- We add `SampleContentCommandsTest` to exercise the commands and ensure they work.

### Manual testing:

- Follow the instructions for adding sample data: https://dkan--4146.org.readthedocs.build/en/4146/user-guide/guide_dataset.html#add-demo-site-content


dafeder (approved):
Looks good

PR #4151, opened 2024-03-26
Add css and js config objects to dkan_js_frontend

dgading:
Fixes:
When splitting up JS files into chunks, they need to be loaded using `type="module"`. This is handled by Drupal via the attributes array on each library entry, but our dynamic hook in the `dkan_js_frontend` module didn't allow for granular updates like this. This PR adds a `js` and `css` object to the config that will allow for custom attributes, but also setting `weight`, `minified`, and `preprocess` for css and js separately. If only the top level minified and preprocess are set, this will still work too. 

- [ ] Documentation exists

## QA Steps

- [ ] Pull new PR
- [ ] Build the site using the decoupled frontend
- [ ] If building a new site, the install config should now add type module to the build file after install
- [ ] If tested on existing site, update config to include  js => attribute setup in the included config. 
- [ ] Once config has been imported, load a React page and view the source code and look the type module on the build/static/js file. 
- [ ] The site should still work with no obvious issues.


brdunfield (approved):
So cool! Confirmed that I can see the added `type="module"` when testing with Open Payments locally! Very nice readme update as well :)

PR #4157, opened 2024-04-09
Reroll #4051: Flexible DateTime widget: Small UX improvement when clearing date

### LINKED ISSUE #4051: Flexible DateTime widget: Small UX improvement when clearing date

dafeder:
fixes #4051 
replaces #4053 

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

- [ ] Clearing a date with time part "00:00:00" should allow to save and clear the time part while processing the value. Can use Release Date in dataset for example.


dafeder:
This looks ready to merge for me, but github won't let me approve (I assume because I added a commit)? Passing along to someone else on the team to push the button...

paul-m (approved):

PR #4164, opened 2024-04-17
Update Docs with recipe for modifying the dataset form

janette:
Schema update mentioned in #4067

Inform users who may be using the distribution level data dictionaries how to modify the dataset form to create a select field.


dafeder:
@janette see https://github.com/GetDKAN/dkan/pull/4067#issuecomment-1830392295

I think we should not make this change to core dataset.ui.json?

dafeder (changes requested):
See convo

janette:
@dafeder revised to update the docs instead

dafeder (approved):
Updated with a few tweaks, approved

PR #4165, opened 2024-04-17
Include security vunerability report process

grugnog:
I think we had something along these lines on the old website, but that isn't active currently, so would be good to include here. This alias goes to our main internal DKAN list.


dafeder (approved):
Looks good, just fighting with this one flaky test.

janette:
@dafeder the test is fixed in #4166

PR #4166, opened 2024-04-18
Redirect to data-dictionary list after form submission

janette:
When user submits a new data dictionary, and it saves successfully, user should end up on /admin/dkan/data-dictionaries.

## QA Steps

- [ ] checkout this branch
- [ ] ddev dkan-site-install
- [ ] ddev drush uli
- [ ] Under the DKAN menu, select Datasets > Create
- [ ] Save new dataset
- [ ] Confirm you end up on the datasets admin view
- [ ] Under the DKAN menu, select Data Dictionary > Create
- [ ] Save new data dictionary
- [ ] Confirm you end up on the data dictionary admin view

### DIFF
--- a/modules/metastore/metastore.install
+++ b/modules/metastore/metastore.install
@@ -145,8 +145,15 @@ function metastore_update_8009() {
     $dict->save();
     $count++;
   }
-  return t("Updated $count dictionaries. If you have overridden DKAN's core schemas, 
+  return t("Updated $count dictionaries. If you have overridden DKAN's core schemas,
     you must update your site's data dictionary schema after this update. Copy
     modules/contrib/dkan/schema/collections/data-dictionary.json over you local
     site version before attempting to read or write any data dictionaries.");
+}
+
+/**
+ * Enable data dictionary widget.
+ */
+function metastore_update_8010() {
+  \Drupal::service('module_installer')->install(['data_dictionary_widget']);
 }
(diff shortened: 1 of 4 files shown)


modules/metastore/metastore.install
```diff
@@ -145,8 +145,15 @@ function metastore_update_8009() {
...
     you must update your site's data dictionary schema after this update. Copy
     modules/contrib/dkan/schema/collections/data-dictionary.json over you local
     site version before attempting to read or write any data dictionaries.");
 }
+
+/**
+ * Enable data dictionary widget.
+ */
+function metastore_update_8010() {
+  \Drupal::service('module_installer')->install(['data_dictionary_widget']);
```
paul-m: I noticed that this works both with and without explicitly enabling the widget module, if you start with a fresh site built with 2.x.

IOW if I build the site with 2.x, and then switch to this branch, and then don't do a `drush updb`, I still get the redirect.

So is the enable necessary?

paul-m (changes requested):
Needs a regression test... I think this should be able to be a BTB test.

modules/metastore/metastore.install
janette: This was added because the Cypress test was failing due to using the older data dictionary form. So we could change the test so that it will pass with either form or look at why the pipeline testing is not using the DD widget.

paul-m (approved):

PR #4168, opened 2024-04-23
#4159: Increase textfields maxlength

### LINKED ISSUE #4159: Increase textfields maxlength
## User Story
Due to core's kind of strange implementation of [Textfield](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21Element%21Textfield.php/class/Textfield/10) any textfield receives maxlength of 128 chars by default.

There is some discussion about clearing/changing this: https://www.drupal.org/project/drupal/issues/3331028

For DKAN any standard schema property of type string will also receive this maxlength of 128 chars. This seems kind of limiting and for no real value. I came across this, while harvesting datasets with strings longer than 128 chars and this is fine, but then when editing the dataset the maxlength will kick in.

As far as I can see there is no option to change this via schema(.ui) properties because the default string field will not go through Widget Router. The place where an increase of maxlength could be integrated would be [handleStringElement](https://github.com/GetDKAN/dkan/blob/2.x/modules/json_form_widget/src/StringHelper.php#L61).

What do you think about this?

- nothing to care in DKAN, either wait for core to change this or change it yourself on a general base?
- change it to something fixed like "256" in DKAN
- make it configurable in DKAN via some schema(.ui) attribute

## Acceptance Criteria
(be able to) increase maxlength on default string textfields

stefan-korn:
fixes #4159
- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

- [ ] textfields have maxlength of 256 instead of 128 previously


stefan-korn:
see #4159 : Currently this is not a complete solution provided in this PR.

dafeder (changes requested):
@stefan-korn I am ok with having this not work with array/object helper quite yet, but as implemented this will only be applied if there is something in the UI schema for the field in question; if you don't have an entry in your .ui.json file for the field it won't get the maxlength. I think it needs to at least work on things like title field even if there is no customization via the UI schema. I'm going to try to come at this in another way as I do need this to work for another project.

dafeder:
Closing in favor of #4212

PR #4173, opened 2024-04-30
Data Dictionary Data Types

kaise-lafrai:
fixes [org/repo/issue#]

- [x] Test coverage exists
- [x] Documentation exists

## QA Steps

- [ ] Navigate to Data Dictionary Form here /node/add/data?schema=data-dictionary
- [ ] Fill out required fields
- [ ] Select Add field
- [ ] Confirm you see these new data types under the Data type dropdown:
datetime
year
boolean
- [ ] Select datetime and confirm you see the following format options
default
any
other
- [ ] Select year and confirm you see the following format options
default
- [ ] Select boolean and confirm you see the following format options
default
- [ ] Confirm when selecting each data type that the description below shows the correct descriptions.
- [ ] Datetime
The format of the data in this field. Supported formats depend on the specified field type:
default: An ISO8601 format string of datetime.
any: Any parsable representation of a date. The implementing library can attept to parse the datetime via a range of strategies.
other: If your date values follow a collective but non-ISO8601 pattern, select this option and define the incoming format using the syntax of [C / Python strftime](https://strftime.org/). For example, if your data had dates formatted as MM/DD/YYYY, you would enter %m/%d/%Y into the Other format field.
- [ ] Year
The format of the data in this field. Supported formats depend on the specified field type:
default: 4-digit numbers in the range 1901 to 2155.
- [ ] Boolean
The format of the data in this field. Supported formats depend on the specified field type:
default: 1/0 values, or True/False values (not case sensitive).
- [ ] Continue adding/editing fields and confirm all the above still functions correctly.

### DIFF
--- a/modules/data_dictionary_widget/src/Fields/FieldValues.php
+++ b/modules/data_dictionary_widget/src/Fields/FieldValues.php
@@ -26,7 +26,7 @@ class FieldValues {
   }
 
   /**
-   * Retrun information about the string field option.
+   * Return information about the string field option.
    */
   public static function returnStringInfo($type) {
     if ($type == 'options') {
@@ -52,7 +52,7 @@ class FieldValues {
   }
 
   /**
-   * Retrun information about the date field option.
+   * Return information about the date field option.
    */
   public static function returnDateInfo($type) {
     if ($type == 'options') {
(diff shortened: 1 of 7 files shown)


modules/data_dictionary_widget/src/Fields/FieldValues.php
```diff
@@ -107,4 +129,38 @@ public static function returnNumberInfo($type) {
...
+  public static function returnYearInfo($type) {
+    if ($type == 'options') {
+      return [
+        'default' => 'default',
+      ];
+    }
+    elseif ($type == 'description') {
+      return "
+        <ul>
+          <li><b>default</b>: Any valid string.</li>
```
janette: Usual lexical representation is YYYY. There are no format options.

modules/data_dictionary_widget/src/Fields/FieldValues.php
```diff
@@ -107,4 +129,38 @@ public static function returnNumberInfo($type) {
...
+  public static function returnBooleanInfo($type) {
+    if ($type == 'options') {
+      return [
+        'default' => 'default',
+      ];
+    }
+    elseif ($type == 'description') {
+      return "
+        <ul>
+          <li><b>default</b>: Any valid string.</li>
```
janette: default:  1/0 values, or True/False values (not case sensitive)

janette (changes requested):
Please update the descriptions on Year and Boolean

modules/data_dictionary_widget/src/Fields/FieldValues.php
janette: or "4-digit numbers in the range 1901 to 2155"

janette (changes requested):
One last change, can you update the format description for the 'number' option from "Any valid string" to "An exact fixed-point number. No non-numeric characters allowed other than the decimal."

https://github.com/GetDKAN/dkan/pull/4173/files#diff-0b145a5ba4f24440bf1ff9183143eb7b2ca0d6389a79bbf0b52b484e735c5a32R127

janette (approved):

PR #4183, opened 2024-05-08
Use the #config_target property in the datastore settings form

jastraat:
This is a demonstration of the new way of tying configuration properties directly to form elements as of Drupal 10.2. This won't be backwards compatible to versions before Drupal 10.2.

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

- [ ] Change the settings on admin/dkan/resources and save
- [ ] Export configuration and confirm that the changes are reflected in the exported configuration.

### Upgrade QA

- [ ] Build site on 2.x
- [ ] Run `drush config:get datastore.settings`
- [ ] Confirm purge_file and purge_table are integers
- [ ] Run `ddev drush config:set datastore.settings purge_table 0` so that we have one true and one false
- [ ] Check out this branch, run `drush updatedb`
- [ ] Run `drush config:get datastore.settings`
- [ ] Confirm purge_file and purge_table are now bools


jastraat:
@paul-m I mostly wanted to use this PR to remind myself of some of the Drupalcon config validation talk and demonstrate the changes in a concrete way; I know DKAN isn't ready yet to drop 10.1. (Although it does lose security support REALLY soon!)

paul-m:
A bunch of fails related to my bad upstream merge, sorry.

dafeder:
I merged `2.x` again and: 
* fixed everything I think the previous merge had broken
* updated to accommodate one field that had been added to the form
* added an update hook to cast existing values to the right type

I am not sure that last thing was needed. Seemed like existing values in the DB could be problematic if they're suddenly not valid, but I could not actually make it error. Perhaps there is some logic in Drupal core that corrects for invalid values?

I am very pro-merging this and starting to use the same strategy in other settings forms.

modules/datastore/datastore.install
```diff
@@ -169,3 +169,13 @@ function datastore_update_9004(&$sandbox) {
...
 }
+
+/**
+ * Update datastore.settings, cast purge_file and purge_table to bool.
+ */
+function datastore_update_9005() {
+  $config = \Drupal::service('config.factory')->getEditable('datastore.settings');
+  $config->set('purge_file', (bool) $config->get('purge_file'));
+  $config->set('purge_table', (bool) $config->get('purge_table'));
+  $config->save();
```
dafeder, line 180: Not sure this is needed? Couldn't hurt 🤷🏽

dafeder (commented):
Approved by me but also includes new work by me so @paul-m should give final OK + merge

paul-m (approved):
We added https://github.com/GetDKAN/dkan/issues/4430 to get the rest of the config forms.

PR #4187, opened 2024-05-14
Reorganize datastore classes

dafeder:
Some expirements with a better organization of datastore module. Would like to divide everything between:

1. Import
2. Alter
3. Query

and for now,

4. Storage


dafeder:
Don't think I'm going to go further with this for now. One thing is given that we're pulling getdkan/harvest into dkan harvest module which has a standard "ETL" terminology, might make more sense to use that if we come back to this.

PR #4188, opened 2024-05-16
Remove deprecated modules/datastore/src/Service.php

dafeder:
Was deprecated in favor of DatastoreService last version.

As long as tests pass we should be fine.


paul-m (approved):
LGTM.

PR #4191, opened 2024-06-03
#4190: Search API integration - Property label and description

### LINKED ISSUE #4190: Search API integration - Property label and description
## Describe the bug
The metastore_search module provides Search API integration. This means you can use every schema property as a Search API field in an index of Datasource DKAN Dataset.

But the Search API fields are only described by the property key and do not have a label nor a description.

Screenshot adding a field (see for example that 'accessLevel" is only described by property key, while normally there would also be a label for a field like with "Rendered HTML output" above.)
grafik
A place where this is more of an issue is if you add an aggregated field and want to choose the "contained fields". Any field from the DKAN Dataset Datasource is just shown as "DKAN Dataset >>" with no additional information. This is also because of missing label and property for the Search API field.
Screenshot:
grafik

## Steps To Reproduce
Add a field to the DKAN search API index

## Expected behavior
Label (and description) for Search API fields of DKAN Dataset Datasource should show up. Take the title property from schema for the label and the description property from schema for the description.

stefan-korn:
fixes #4190

### DIFF
--- a/modules/metastore/modules/metastore_search/src/ComplexData/Dataset.php
+++ b/modules/metastore/modules/metastore_search/src/ComplexData/Dataset.php
@@ -52,6 +52,8 @@ class Dataset extends ComplexDataFacade {
     }
     else {
       $defs[$property_name] = self::getDefinitionObject($type);
+      $defs[$property_name]->setLabel($object->properties->{$property_name}->title ?? $property_name);
+      $defs[$property_name]->setDescription($object->properties->{$property_name}->description ?? '');
     }
     return $defs;
   }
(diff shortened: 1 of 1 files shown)


stefan-korn:
One remark: I am not sure if the code would ever get here: https://github.com/GetDKAN/dkan/blob/2.x/modules/metastore/modules/metastore_search/src/ComplexData/Dataset.php#L77
With default DKAN schema it does not get there I suppose, if it would get there there might be a problem with this MR.

dafeder:
> One remark: I am not sure if the code would ever get here: https://github.com/GetDKAN/dkan/blob/2.x/modules/metastore/modules/metastore_search/src/ComplexData/Dataset.php#L77 With default DKAN schema it does not get there I suppose, if it would get there there might be a problem with this MR.

@stefan-korn yeah tbh I understand that final `else` looking at it. There are no other types than object or array that make sense here, the conditional should either not be there or throw an exception.

dafeder (approved):
Looks good @stefan-korn! What an improvement.

I think I'll add some test coverage for this and update the default search config in another PR but IMO this can be merged as is since technically the code is "covered" by a test just not by any specific assertions.

PR #4192, opened 2024-06-06
Cache filtered CSV downloads for one hour

paul-m:
The issue is that even if our infrastructure is set up to cache streaming CSV responses, DKAN tells it not to by defaulting to `Cache-Control: private`.

Luckily, we can use `Drupal\common\CacheableResponseTrait` to change the default on the existing controller, `Drupal\datastore\Controller\QueryDownloadController`. Setting a max age of one hour also tells this trait to use `Cache-Control: public`.

- Fixes minor bug in `Drupal\common\CacheableResponseTrait`.
- Leverages `Drupal\common\CacheableResponseTrait` to set max-age cache header to an hour.
- Updates `Drupal\Tests\datastore\Unit\Controller\QueryDownloadControllerTest` to check for the cache header.
- Updates other tests with `@group` so I can run them locally without running the whole test suite.

### DIFF
--- a/modules/common/src/CacheableResponseTrait.php
+++ b/modules/common/src/CacheableResponseTrait.php
@@ -50,7 +50,8 @@ trait CacheableResponseTrait {
     if (!isset($this->cacheMaxAge)) {
       // A hack to bypass the controllers' tests.
       if (\Drupal::hasService('config.factory')) {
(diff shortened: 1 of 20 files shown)


dafeder (approved):
Note the new @todo in the code; this should really be parameterized; we shouldn't assume 1 hour is right for all use cases.

PR #4197, opened 2024-06-20
Check if downloadURL isset before rendering

### LINKED ISSUE #4194: Undefined array key "downloadURL" in metastore process node data
## Describe the bug
If a site has distributions with an accessUrl and no downloadUrl the undefined key error below appears. This is from metastore module's theme.inc file were some variables for themeing and a table are created. 

`Warning: Undefined array key "downloadURL" in metastore_preprocess_node__data() (line 101 of /var/www/html/docroot/modules/contrib/dkan/modules/metastore/metastore.theme.inc)`

## Steps To Reproduce
Build a new DKAN site with a schema that doesn't require downloadUrl key in the distribution objects. 
Import data and check for the errors after a page render. 

## Expected behavior
This shouldn't cause an error if the key doesn't exist.

dgading:
fixes #4194 

## QA Steps

1. [ ] Build the demo site and install sample data. Visit a dataset /node page to make sure the template still adds the download urls to the page.
2. [ ] Create a new dataset. Fill in the required fields.
3. [ ] For the distribution section, add a new distribution, leaving Download URL section empty and adding an Access URL. This will make the frontend render this section and you will not see the error. (The access url does not render well since it didn't have an explicit variable set in the theme file.)

I've also added the variable `metastore` for templates to use. This is just the raw json object with no processing on it. This is the first step towards deprecating the dataset variables. The new metastore variable can be used in node--data templates by using something like `metastore.title` or `metastore.keyword[0]`.

`Warning: Undefined array key "downloadURL" in metastore_preprocess_node__data() (line 101 of /var/www/html/docroot/modules/contrib/dkan/modules/metastore/metastore.theme.inc)`

### DIFF
--- a/modules/metastore/metastore.theme.inc
+++ b/modules/metastore/metastore.theme.inc
@@ -15,6 +15,7 @@ function metastore_preprocess_node__data(&$variables) {
   $json = $variables['node']->get('field_json_metadata')->value;
   $rows = [];
   $metadata = json_decode($json);
+  $variables['metastore'] = $metadata;
   if (property_exists($metadata, 'description')) {
     $variables['dataset']['description'] = $metadata->description;
   }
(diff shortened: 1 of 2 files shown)


dafeder (approved):
This accomplishes the goal of the ticket, which was to remove the error at the top. We should not as a known issue though that distributions do not render particularly coherently if there is an accessURL instead of a downloadURL.

PR #4201, opened 2024-06-25
Fix broken reference in OpenAPI spec, improve validation tests

dafeder:
There were some inconsistencies in the responses for the metastore section of the OpenAPI spec. Changing them broke the crude response validator I had implemented using OPIS schema validation. This fixes up the spec and switches to using [osteel/openapi-httpfoundation-testing](https://github.com/osteel/openapi-httpfoundation-testing) for response validation, which is I think a better approach.

## QA Steps

1. Stand up side, and grab the full spec from /api/1
2. Paste the spec into https://editor.swagger.io/ and confirm it throws no errors
3. Perform any additional validation you wish

## Merging instructions

1. Review and approve this PR. DO NOT MERGE YET
2. Merge GetDKAN/recommended-project#39
3. Merge GetDKAN/recommended-project#40
4. Update .circleci/config.yml to use correct branches of GetDKAN/recommended-project again (feel free to reassign back to @dafeder for this part)
5. Confirm tests still pass and merge this PR

## Notes

Re: codeclimate, that function is completely declarative and wouldn't benefit from breaking up, so I think ok to override.

This PR removes Drupal 10.0 from the CI matrix, which was creating dependency issues for the openapi library. 10.0 is no longer supported so we feel comfortable dropping testing for it. This does not mean that production builds with DKAN will fail for 10.0, but core PHPUnit tests will.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -236,33 +236,33 @@ workflows:
           report_coverage: true
           matrix:
             parameters:
-              dkan_recommended_branch: [ '10.2.x-dev']
+              dkan_recommended_branch: [ 'osteel-openapi-10-2']
               php_version: [ '8.3' ]
       - phpunit:
           matrix:
             parameters:
-              dkan_recommended_branch: [ '10.2.x-dev']
+              dkan_recommended_branch: [ 'osteel-openapi-10-2']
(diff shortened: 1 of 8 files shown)


paul-m (approved):
Approving this change, NOT MERGING YET per instructions....

PR #4202, opened 2024-06-26
Apply dictionary date formats on csv downloads when in reference mode

janette:
ref: 21261

## QA Steps

- [ ] Set the data dictionary mode to 'reference' https://dkan.ddev.site/admin/dkan/data-dictionary/settings
- [ ] Create a data dictionary with a field of type date, set the format to %m/%d/%Y
- [ ] Create a dataset with a data file containing a date column that you used in your data dictionary, dates should be in mm/dd/YYYY format
- [ ] Use your data dictionary endpoint in the distribution describedBy field
- [ ] Save and run import queues
- [ ] Visit your dataset, adjust the url to call the streamed download, or use the link below replacing the uuid with your dataset uuid

https://dkan.ddev.site/api/1/datastore/query/505a1f80-7a95-4d9e-8eda-c57efdbf5c19/0/download?&format=csv

- [ ] Confirm the resulting csv has the dates in the DD format, not ISO

### DIFF
--- a/modules/datastore/src/DatastoreService.php
+++ b/modules/datastore/src/DatastoreService.php
@@ -217,8 +217,8 @@ class DatastoreService implements ContainerInjectionInterface {
   /**
    * Returns the Data Dictionary fields.
    */
-  public function getDataDictionaryFields() {
-    return $this->dictionaryEnforcer->returnDataDictionaryFields();
+  public function getDataDictionaryFields($id = NULL) {
+    return $this->dictionaryEnforcer->returnDataDictionaryFields($id);
   }
 
   /**
--- a/modules/datastore/src/DatastoreService.php
+++ b/modules/datastore/src/DatastoreService.php
@@ -216,9 +216,12 @@ class DatastoreService implements ContainerInjectionInterface {
 
   /**
    * Returns the Data Dictionary fields.
+   *
+   * @param string $identifier
+   *   A resource's identifier. Used when in reference mode.
    */
-  public function getDataDictionaryFields($id = NULL) {
-    return $this->dictionaryEnforcer->returnDataDictionaryFields($id);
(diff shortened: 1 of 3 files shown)


modules/datastore/src/Service/ResourceProcessor/DictionaryEnforcer.php
```diff
@@ -124,14 +124,25 @@ public function applyDictionary(RootedJsonData $dictionary, string $datastore_ta
    *
    *  {@inheritdoc}
    */
-  public function returnDataDictionaryFields() {
-    // Get DD is mode.
+  public function returnDataDictionaryFields($id = NULL) {
```
dafeder: This function should have a docbloc, the method is unique to this class and is not defined in the interface.

modules/datastore/src/DatastoreService.php
```diff
@@ -217,8 +217,8 @@ public function getImportService(DataResource $resource): ImportService {
   /**
    * Returns the Data Dictionary fields.
    */
-  public function getDataDictionaryFields() {
-    return $this->dictionaryEnforcer->returnDataDictionaryFields();
+  public function getDataDictionaryFields($id = NULL) {
```
dafeder: Let's document this method properly with docbloc and type/return hints

dafeder (changes requested):

modules/datastore/src/DatastoreService.php
janette: Good call thanks

modules/datastore/src/Service/ResourceProcessor/DictionaryEnforcer.php
janette: Right again

dafeder (approved):

PR #4206, opened 2024-06-28
Skip the data dictionary when one does not exist

janette:
ref: 21364

## QA Steps

- [ ] Put site in DD [reference mode](https://dkan.ddev.site/admin/dkan/data-dictionary/settings)
- [ ] Create a dataset without a DD
- [ ] Go to the dataset page, create a filter, grab the query url
- [ ] Add download bits like the following
```
https://dkan.ddev.site/api/1/datastore/query/c59d6e04-0735-5a18-b07b-a41b27b0659b/0/download?limit=20&offset=0&conditions[0][property]=year&conditions[0][value]=2012&conditions[0][operator]==&format=csv
```
- [ ] Confirm you get a file with the results you expect
- [ ] Confirm no errors in the file

### DIFF
--- a/modules/datastore/src/DatastoreService.php
+++ b/modules/datastore/src/DatastoreService.php
@@ -219,8 +219,11 @@ class DatastoreService implements ContainerInjectionInterface {
    *
    * @param string $identifier
    *   A resource's identifier. Used when in reference mode.
+   *
+   * @return array|null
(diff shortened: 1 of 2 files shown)


dafeder (approved):
This lgtm, but I tested it without adding a filter, does that actually matter?

PR #4209, opened 2024-07-01
#4208: New permission for administering metastore settings

### LINKED ISSUE #4208: [Feature] New permission for administering metastore settings
### Feature Description

Administration of metastore settings (admin/dkan/properties) should use its own permission

Currently the metastore settings use the drupal core permission _access administration pages_. Imho this should better use its own permission allowing to grant permission for this part of DKAN configuration individually and not based on a (widely used) core permission.

### Problem and Motivation

If you want to grant or deny permission for administering metastore settings finegrained, you are currenlty not able to do this (you need to add or remove a drupal core permssion).

### Potential Benefits

more finegrained permission set for DKAN.

### Target Audience

- [ ] End Users
- [ ] Developers
- [X] Administrators
- [ ] Other

### Possible Implementation

provide a new permission for administering the metastore settings and use it on the route.

### Additional Context

_No response_

stefan-korn:
fixes #4208 

- [ ] Test coverage exists
- [ ] Documentation exists

## QA Steps

- [ ] check permissions under admin/people/permissions at section Metastore and see the new permission 'Metastore: Administer Metastore Settings"
- [ ] grant this permission to a role
- [ ] check that the role with this permission can access admin/dkan/properties
- [ ] check that a role without this permission cannot access admin/dkan/properties

### DIFF
--- a/modules/metastore/metastore.permissions.yml
+++ b/modules/metastore/metastore.permissions.yml
@@ -4,7 +4,9 @@
 'administer data dictionary settings':
   title: 'Metastore: Administer Data Dictionary Settings'
   description: 'Administer data dictionary settings'
-
+'administer metastore settings':
+  title: 'Metastore: Administer Metastore Settings'
+  description: 'Administer metastore settings'
 
 
 administer resource mapping:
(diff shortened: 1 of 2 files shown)


dafeder (approved):
Yes discussed this at last team meeting and all agree this makes sense. We'll put something in the release notes to flag it since it could mean users lose access to this after update.

PR #4211, opened 2024-07-02
Pass empty array if no schema is found

janette:
ref: 20621

```
TypeError: Drupal\common\Storage\AbstractDatabaseTable::getSchema(): Return value must be of type array, null returned in Drupal\common\Storage\AbstractDatabaseTable->getSchema() (line 397 of /var/www/html/dkan/modules/common/src/Storage/AbstractDatabaseTable.php). 
```

When data publishers rely on cron to run the queues, a datastore query may return the above 500 error on getSchema() when:
1. a data file (downloadURL) has been updated
2. the localize_import has run
3. the datastore_import has not run 
4. a datastore/query API call is made to the missing datastore

To avoid the 500 error:
Allow **getSchema** to return an empty array, the query will then be caught by **queryResource** and return a 400

### DIFF
--- a/modules/common/src/Storage/AbstractDatabaseTable.php
+++ b/modules/common/src/Storage/AbstractDatabaseTable.php
@@ -391,7 +391,7 @@ abstract class AbstractDatabaseTable implements DatabaseTableInterface {
    * Get the schema for this table.
    *
    * @return array
-   *   A schema array.
+   *   A Drupal Schema API array, or an empty array if none found.
    */
   public function getSchema(): array {
     return $this->schema ?? [];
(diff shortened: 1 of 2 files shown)


modules/common/src/Storage/AbstractDatabaseTable.php
```diff
@@ -394,7 +394,7 @@ public function setSchema(array $schema): void {
    *   A schema array.
    */
   public function getSchema(): array {
-    return $this->schema;
+    return $this->schema ? $this->schema : [];
```
dafeder:
```suggestion
    return $this->schema ?? [];
```

modules/common/src/Storage/AbstractDatabaseTable.php
```diff
@@ -394,7 +394,7 @@ public function setSchema(array $schema): void {
    *   A schema array.
```
dafeder:
```suggestion
   *   A Drupal Schema API array, or an empty array if none found.
```

dafeder (changes requested):
This should have a test, either a unit test for the class or in a functional test of the API. Pretty sure this once worked and then broke, a test should have caught that in the first place.

dafeder (approved):

PR #4212, opened 2024-07-02
Increase textfields maxlength

### LINKED ISSUE #4159: Increase textfields maxlength

dafeder:
Fixes #4159 

Based on  #4168 by @stefan-korn, add a default string length of 256 to better accommodate longer titles and other textfields that don't quite warrant a textarea.

Also, the default length will be overridden if the schema itself provides a `maxLength`.

## QA Steps

- [ ] Try adding a dataset with a title longer than 128 chars but equal or less than 256 chars. For example, "This is a dataset, this is a dataset, this is a dataset, this is a dataset, this is a dataset, this is a dataset, this is a dataset, this is a dataset, this is a dataset, this is a dataset, this is a dataset, this is a dataset, this is a dataset" (246 chars).
- [ ] Confirm the dataset saves without issue
- [ ] Now, edit your dataset schema and add to the title property `"maxLength": 20`.
- [ ] Edit and re-save your dataset, confirm your submission is rejected with the message "Title cannot be longer than 20 characters but is currently 245 characters long."
- [ ] Now, edit your dataset schema again, remove the `maxLength` from the title property and add to the describedByType property `"maxLength": 400`.
- [ ] Reload the form and try pasting this 277-character string into your describedByType field (under the main dataset, not the distribution):

/application/application/application/application/application/application/application/application/application/application/application/application/application/application/application/application/application/application/application/application/application/application/application


dgading (approved):
Worked great.

PR #4215, opened 2024-07-03
Provide number widget in form

dafeder:
Re-roll/add test coverage for #4069 by @stefan-korn 

## QA steps

Add a field to the dataset.json schema file:

```json
    "number": {
      "title": "Number",
      "description": "A unique number for the dataset or API as maintained within an Agency catalog or database.",
      "type": "string",
      "minLength": 1
    }
```

now add settings to dataset.ui.json:

```json
  "number": {
    "ui:options": {
      "widget": "number",
      "step": 10
    }
  }
```

Go to create dataset form, confirm there is a number widget that increments by 10 on every click.


modules/json_form_widget/tests/src/Unit/WidgetRouterTest.php
```diff
@@ -71,6 +71,26 @@ public static function dataProvider(): array {
           '#target_type' => 'node',
         ],
       ],
+      // Tag field is a free-tagging autocomplete that populates from metastore.
```
kaise-lafrai: Please update this comment to reflect the numberField and not the Tag field.

kaise-lafrai (changes requested):
Looks good and QA steps passed. One small request is to update the code comment for the numberField in the WidgetRouterTest.

modules/json_form_widget/tests/src/Unit/WidgetRouterTest.php
dafeder: OOps I did that and didn't commit it 🤦🏽

kaise-lafrai (approved):

PR #4229, opened 2024-07-18
Change private methods to protected in WidgetRouter

stefan-korn:
@dafeder : Some time ago we had a conversation about use of private vs protected for methods in DKAN and you indicated that you might be willing to change some methods from private to protected if there is no special reason for being private and no issues arise.

This PR is to change three methods in WidgetRouter from private to protected.

Background to this is, that I am working on an integration for Group module to allow publishers to depend on Group membership. Therefore I want to allow only those options for choosing a publisher to which the current user has access based on his group membership. I want to decorate WidgetRouter and override getOptionsFromMetastore and have the need to call metastoreOptionTitle() and metastoreOptionValue() from the decorating class which is not possible if they are private.

I changed handleSelectOtherDefaultValue() from private to protected too, though I do not needs this at this moment, but seems more consistent to have all three methods changed from private to protected.

### DIFF
--- a/modules/json_form_widget/src/WidgetRouter.php
+++ b/modules/json_form_widget/src/WidgetRouter.php
@@ -240,7 +240,7 @@ class WidgetRouter implements ContainerInjectionInterface {
    * @return string
    *   String to be used in title.
    */
-  private function metastoreOptionTitle($item, $titleProperty): string {
+  protected function metastoreOptionTitle($item, $titleProperty): string {
     if ($titleProperty) {
       return is_object($item) ? $item->data->$titleProperty : $item;
     }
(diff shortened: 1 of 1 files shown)


stefan-korn:
@dafeder : While looking at WidgetRouter code, I am asking myself if WidgetRouter really needs the create method:
https://github.com/GetDKAN/dkan/blob/2.x/modules/json_form_widget/src/WidgetRouter.php#L42
The WidgetRouter is a service and gets its dependent services from services.yml arguments?

dafeder:
@stefan-korn that sounds very useful, I hope you'll share the code when you finish. This looks good, will approve and merge.

We likely don't need that create() method, I think there are a few services throughout DKAN that had those, primarily because it made it easier to mock and test these classes, at least in the way certain people were writing tests at the time. We can try factoring it out but my guess is just deleting it would break existing tests.

dafeder (approved):

stefan-korn:
@dafeder : Thanks for merging this.

I have already created a module: https://www.drupal.org/project/dkan_group - Will mention this in discussions after has become a bit more stable.

PR #4235, opened 2024-07-22
Use Request in harvest WebServiceApi

paul-m:
- Removes useage of request_stack service in the harvest WebServiceApi controller.
- Adjusts tests to reflect this new reality.


dafeder (approved):
Looks good!

PR #4244, opened 2024-08-01
Adds QueueRunnerTrait for consistency in testing + DRY

paul-m:
Approximately 958723498502834759230485723094587 of our tests have their own copypasted definition of a `runQueues()` method.

Let's turn that into a trait which is then reused.


rovcase (approved):
Thanks for jumping on a call to give some context @paul-m, cool way of reducing repeated code. Looks good to me!

PR #4245, opened 2024-08-02
Convert DkanDocsTest to BTB

paul-m:


dmundra (approved):
Code cleanup looks good and the tests were passing with 78% code coverage, the last changes bumped that to 100% so I think good to go once tests are confirmed to be passing in latest pipeline.

PR #4246, opened 2024-08-02
Convert ResourcePurgerTest to BTB

paul-m:


dmundra (approved):
Nice. Much simpler now that you can just specify the modules instead of having to scaffold the environment.

Issue #4210, opened 2024-07-02
Small inconsistency naming DKAN dataset/content overview (admin/dkan/datasets)


stefan-korn:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The menu link for the dataset overview is defined in https://github.com/GetDKAN/dkan/blob/2.x/modules/metastore/metastore.links.menu.yml#L32 with description _Dataset content view_
If the module metastore_admin is enabled, the default dataset content view from metastore.routing.yml is replaced by the DKAN Dataset content view https://github.com/GetDKAN/dkan/blob/2.x/modules/metastore/modules/metastore_admin/config/install/views.view.dkan_dataset_content.yml with description 'Administer all content in the DKAN metastore'.
But the description for the menu link is not changed still reading _Dataset content view_

### Expected Behavior

If the metastore_admin module is enabled, the description of the menu link should read the same as is provided by the DKAN Dataset view: _Administer all content in the DKAN metastore_
And since dkan requires metastore_admin module as a dependency, this probably applies generally.

This could be solved in different ways I suppose. Directly changing the in metastore.links.menu.yml, changing/overwriting in metastore_admin module ...

In addition I am not sure about this defintion in metastore.routing.yml:
https://github.com/GetDKAN/dkan/blob/2.x/modules/metastore/metastore.routing.yml#L185
Is this necessary? Seems using duplicate path admin/dkan/datasets.

### Anything else?

If you give me a hint if and what direction in changing this you may want to follow I will provide a MR.

dafeder:
> In addition I am not sure about this defintion in metastore.routing.yml:
https://github.com/GetDKAN/dkan/blob/2.x/modules/metastore/metastore.routing.yml#L185
Is this necessary? Seems using duplicate path admin/dkan/datasets.

@janette I think this was supposed to create a sub-menu item so you could get straight to the create form from the admin menu -- I could swear this worked at one point but the path is clearly wrong?

dmundra:
Fixed by #4248

PR #4248, opened 2024-08-05
Match view description to menu description

janette:
fixes [#4210]

## QA Steps

- [x] ddev dkan-site-install
- [x] ddev drush uli
- [x] Confirm no change to the DKAN menu


dmundra (commented):
@janette this change makes the view description match https://github.com/GetDKAN/dkan/blob/2.x/modules/metastore/metastore.links.menu.yml#L34 but the ticket #4210 was asking to change the menu link to match the view description 'Administer all content in the DKAN metastore'. Wouldn't that be better?

dmundra (approved):
Looks great and thank you for making the items consistent.

PR #4251, opened 2024-08-05
Converts NodeDataTest to BTB

paul-m:


dmundra (approved):
Changes to use services make sense and the tests are passing

PR #4252, opened 2024-08-05
Convert HideCurrentRevisionActionTest to BTB

paul-m:


dmundra (approved):
Also straightforward like #4251.

PR #4254, opened 2024-08-06
Convert SearchTest to BTB

paul-m:
- Converts `SearchTest` to BTB.
- Removes unneeded references to the request stack service.


dmundra (approved):
Looking good especially the improved drush command and asserting 0 results.

PR #4256, opened 2024-08-07
Remove dependency on weitzman/drupal-test-traits

paul-m:
- Removes weitzman/drupal-test-traits from dependencies.
- Constraints PHPUnit to `^9.6` which fits our current requirements.
- Deprecates `Drupal\Tests\common\Traits\CleanUp` which should not be necessary in DKAN core.
- Requires a change to getdkan/ddev-dkan to allow for different test autoloaders. Merged and released: https://github.com/GetDKAN/ddev-dkan/releases/tag/1.2.0 ✅
- Follow-up with a change to getdkan/recommended-project which will remove DTT from its dev dependencies.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ commands:
         type: string
       addon_branch:
         description: 'Repo branch name for the dkan-ddev-addon you want to test against.'
-        default: 'main'
+        default: 'dev-21895-remove-dtt'
         type: string
       dkan_recommended_branch:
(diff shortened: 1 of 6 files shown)


paul-m:
Added a branch to make sure we can still use dtt with the dkan ddev addon: https://github.com/GetDKAN/dkan/tree/21895-remove-dtt-test-only

dmundra (approved):
Looks good to me and nice to see a dependency being removed.

dafeder (approved):

PR #4261, opened 2024-08-12
Add config for which column header to use on filtered csv downloads

janette:
re: 21838

## Describe your changes
To allow flexibility on the column headings being used for datastore query csv downloads when a data dictionary is in use, this new global configuration setting will determine between the original column headings or the title values saved to the data dictionary in case these two values differ from each other.

## QA Steps

- [x] details in 21838.

### DIFF
--- a/modules/datastore/src/DataDictionary/AlterTableQuery/MySQLQuery.php
+++ b/modules/datastore/src/DataDictionary/AlterTableQuery/MySQLQuery.php
@@ -3,7 +3,6 @@
 namespace Drupal\datastore\DataDictionary\AlterTableQuery;
 
 use Drupal\Core\Database\StatementInterface;
-
 use Drupal\datastore\Plugin\QueueWorker\ImportJob;
 use Drupal\datastore\DataDictionary\AlterTableQueryBase;
 use Drupal\datastore\DataDictionary\AlterTableQueryInterface;
@@ -93,6 +92,20 @@ class MySQLQuery extends AlterTableQueryBase implements AlterTableQueryInterface
   ];
 
   /**
+   * Config for which column heading values to use for csv downloads.
+   *
+   * @var string
+   */
+  protected string $csvHeadersMode = 'resource_headers';
+
+  /**
+   * Assign the csvHeaderMode based on the config setting.
+   */
+  public function setCsvHeaderMode($mode) {
+    $this->csvHeadersMode = $mode;
+  }
+
+  /**
    * {@inheritdoc}
    */
   public function doExecute(): void {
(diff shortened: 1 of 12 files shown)


modules/metastore/src/Form/DataDictionarySettingsForm.php
```diff
@@ -119,6 +119,20 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
 
+    $form['export_titles_mode'] = [
+      '#type' => 'select',
+      '#title' => $this->t('Export Titles Mode'),
+      '#options' => [
+        'dictionary_titles' => $this->t('Use data dictionary titles'),
+        'resource_titles' => $this->t('Use the column names from the resource file'),
+      ],
+      '#default_value' => $config->get('export_titles_mode'),
+      '#description' => $this->t("Chose which title values will be used for filtered csv exports."),
```
dmundra: Nitpick, I think it should say 'Choose which title ...' and also capitalize CSV?

janette: 🤦‍♀️ yes

modules/datastore/src/DataDictionary/AlterTableQuery/MySQLQuery.php
```diff
@@ -92,6 +95,31 @@ class MySQLQuery extends AlterTableQueryBase implements AlterTableQueryInterface
...
+  protected ConfigFactoryInterface $configFactory;
+
+  /**
+   * Api constructor.
+   */
+  public function __construct(
+    ConfigFactoryInterface $configFactory
+  ) {
+    $this->configFactory = $configFactory;
+  }
```
dmundra: @janette because you are introducing a constructor here, you have to also handle calling the parent constructor for `AlterTableQueryBase` which is what the test does on line https://github.com/GetDKAN/dkan/pull/4261/files#diff-0136d115ea55175e8fb916adbd8e0e9de87e7a713fc6a31d0d1852bd3753c0e9R101

dmundra: One option could be moving the config factory to the base class which might be useful but could have downstream effects.

modules/datastore/src/DataDictionary/AlterTableQuery/MySQLQuery.php
```diff
@@ -441,21 +452,29 @@ protected function buildDatabaseTypeMap(array $fields, string $table): array {
...
+   * @param string $table
+   *   Mysql table name.
    *
    * @return string[]
    *   Modify column options.
    */
-  protected function buildModifyColumnOptions(array $fields, array $type_map): array {
+  protected function buildModifyColumnOptions(array $fields, array $type_map, string $table): array {
     $modify_column_options = [];
+    $table_cols = $this->getTableColsAndComments($table);
```
dmundra: Nitpick, since this is only used in the 'else' condition it might be better to put it in that section.

dmundra (approved):
Left one small nitpick but good to go I believe.

PR #4267, opened 2024-08-30
Split sanitize and normalize to be separate operations

janette:
21640

WIP: Fixes applying date formats when using data dictionaries and the streamed CSV download API

## Describe your changes
We only want to sanitize expressions, not when adding fields to the query.

## QA Steps

- [ ] TBD

### DIFF
--- a/modules/common/src/Storage/SelectFactory.php
+++ b/modules/common/src/Storage/SelectFactory.php
@@ -104,10 +104,8 @@ class SelectFactory {
   private function addDateExpressions($db_query, $fields, $meta_data) {
     foreach ($meta_data as $definition) {
       // Confirm definition name is in the fields list.
-      $name = $this->dbQuery->escapeField($definition['name']);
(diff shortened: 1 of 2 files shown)


dmundra:
@janette is the issue that the branch this is targeting is 'update-addDateExpressions' instead of '2.x'?

dmundra:
Your also might need to rebase this branch of 2.x

modules/common/src/Storage/SelectFactory.php
```diff
@@ -152,6 +150,10 @@ private function normalizeProperty(mixed $property): object {
     }
     // Throw exception if obviously unsafe property name.
     self::safeProperty($property->property);
+    return $property;
+  }
+
+  private function sanitizeProperty(object $property) {
```
dmundra, line 156: @janette I think splitting this function potentially causing the `$alias` error but I am not sure. It is one place to debug.

PR #4269, opened 2024-09-01
Fixes some HarvestCommand logging

paul-m:
Fixes [issue#]

## Describe your changes

- Removes logger as an injected dependency from `HarvestCommands`.
- Converts all commands in `HarvestCommands` to use `$this->logger()` instead of alternatives.
- Converts all message strings to use concatenation rather than interpolation.
- Adds minimal automated test of commands.

## QA Steps

Use sample_content module to create a harvest.

Try out the commands.

### DIFF
--- a/modules/harvest/drush.services.yml
+++ b/modules/harvest/drush.services.yml
@@ -3,7 +3,6 @@ services:
     class: \Drupal\harvest\Commands\HarvestCommands
     arguments:
       - '@dkan.harvest.service'
-      - '@dkan.harvest.logger_channel'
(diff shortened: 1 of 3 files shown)


dmundra (approved):
Changes look solid to me and the commands work in my sandbox. @paul-m anything to be done about code climate https://codeclimate.com/github/GetDKAN/dkan/pull/4269 issue about shared code between `archive` and `publish`?

dmundra (approved):
Code climate changes look good as well. Approved and merging.

PR #4272, opened 2024-09-03
Drush deregister improvements

paul-m:
Fixes [issue#]

## Describe your changes

Modifies `drush dkan:harvest:deregister`:
- Adds a confirmation step ('Are you sure?'). Can be bypassed with `-y`, `--yes`, or `--no-interaction`.
- Adds a flag to the command line, `--revert`. This will revert all datasets in the harvest before deregistering, and is the equivalent of `drush dkan:harvest:revert ID && drush dkan:harvest:deregister ID`

## QA Steps

### DIFF
--- a/docs/source/drush_commands.rst
+++ b/docs/source/drush_commands.rst
@@ -118,7 +118,13 @@ dkan:harvest:archive
 dkan:harvest:deregister
 -----------------------
 
-    Deregister a harvest.
+    Deregister a harvest. This action removes the harvest plan from the system,
+    but will leave the datasets in place. This is good if you are using harvest
+    to migrate content and you do not plan on using the harvest feature to
+    update the datasets.
+
+    If you want to remove the harvest plan AND the datasets associated with it,
+    use the --revert option.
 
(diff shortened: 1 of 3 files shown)


janette (approved):

docs/source/drush_commands.rst
```diff
@@ -123,6 +123,11 @@ dkan:harvest:deregister
     **Arguments**
```
janette, line 129: Let's define this a little more on line 121.

Deregister a harvest. This action removes the harvest plan from the system, but will leave the datasets in place. This is good if you are using harvest to migrate content and you do not plan on using the harvest feature to update the datasets. 
If you want to remove the harvest plan AND the datasets associated with it, use the `--revert` option.

janette: Also update the dkan:harvest:revert description

Revert a harvest, i.e. remove harvested entities and unpublish orhpaned keywords, themes, and distributions. The harvest plan will remain and can be run again to generate the datasets after any issues have been resolved.

PR #4276, opened 2024-09-05
Add tests for ResourceImporter and sample_content

### LINKED ISSUE #4205: Don't require file_public_base_url to be set manually
### Feature Description

Every new DKAN site requires file_public_base_url to be set with the full URL of the site if sample content is to be created.

If I understand the issue, it is because we use a harvest to create the sample content, and that harvest works by downloading remotely, but sites don't always know their own URL when called from the command line?

### Problem and Motivation

This is a paper cut that can trip up users and adds steps to the setup process.

### Potential Benefits

Makes it easier to test and deploy new DKAN sites.

### Target Audience

- [x] Administrators

### Possible Implementation

Not sure what the easiest approach, but I think for a full fix we need to enable sites to harvest from a local file path/uri (without redownloading it), at least when called programmatically.

### Additional Context

_No response_

paul-m:
Fixes [#4205] Be sure and see the notes there.

## NOTE:

This PR adds tests to support a change to the DKAN DDev add-on. See https://github.com/GetDKAN/ddev-dkan/pull/52

## Describe your changes

- Adds tests for `ResourceImporter`, kernel and functional.
- Performs minor refactoring of `ResourceImporter`:
  - Short-circuits unneeded else statements.
  - Injects services explicitly.
- Adds steps to `SampleContentCommandsTest` to prove that the Drush commands work without setting `file_public_base_url`.

## Discoveries

- `Drupal\common\Util\DrupalFiles` uses the deprecated `system_retrieve_file()` function, which goes away in Drupal 11.
- `DrupalFiles` also does not handle file wrappers properly, leading to an inability to write a kernel test. (Kernel tests use in-memory vfsStream file wrappers.)

Follow-up recommendations:

- Refactor `DrupalFiles` to play nice with vfsStream and also to not use `system_retrieve_file()`.
- Turn `Drupal\Tests\harvest\Functional\Transform\ResourceImporterTest` into a kernel test.
- Investigate whether we actually need `DrupalFiles`, since it's only used in `Drupal\harvest\Transform\ResourceImporter` and `ResourceLocalizer` service.

## QA Steps

- [ ] Stand up a local build with the special settings config
- [ ] Create sample content
- [ ] Remove the config
- [ ] Remove sample content
- [ ] Create sample content again

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ commands:
         type: string
       addon_branch:
         description: 'Repo branch name for the dkan-ddev-addon you want to test against.'
-        default: 'main'
+        default: 'whither-file-public-base-url'
         type: string
       dkan_recommended_branch:
         description: 'Branch of getdkan/recommended-project to use.'
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,7 +29,7 @@ commands:
         default: '8.2'
         type: string
       addon_branch:
-        description: 'Repo branch name for the dkan-ddev-addon you want to test against.'
+        description: 'Repo branch name for the DKAN DDev add-on you want to test against.'
         default: 'whither-file-public-base-url'
         type: string
       dkan_recommended_branch:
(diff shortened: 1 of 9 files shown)


dafeder:
@janette is trying this on some existing sites to see if there is any affect we might be missing in tests. We'll create a new ticket for the follow-up recommendations as this PR as is satisfies #4205

janette (approved):
This is good to merge, no issues for existing sites.

PR #4286, opened 2024-09-13
DKAN-4285 Finish migration to harvest_runs.

### LINKED ISSUE #4285: harvest_update_8007 did not execute harvestRunsUpdate
### Current Behavior

The site on upgrading to 2.19 from lower versions does not populate the new `harvest_runs` table.

[harvest_update_8007](https://github.com/GetDKAN/dkan/blob/1a993fbefc9ddd437fecb561f36f64757cb90102/modules/harvest/harvest.install#L86) calls `harvestUtility->harvestHashUpdate()` but did not finish the update by calling `harvestRunsUpdate`

This prevents using HarvestService->getLastHarvestRunId() or HarvestService->bulkUpdateStatus()

An new update hook should be added to finish the process so it matches the two step method used by `dkan:harvest:update`

### Expected Behavior

I should be able to use the bulkUpdateStatus to publish datasets.

### Steps To Reproduce

1. Update from 2.18 to 2.19
2. run drush updb
3. see that harvest_update_8007 runs successfully
4. check harvest_runs table and see that it is empty, because the runs were never migrated.

_No response_

### Anything else?

_No response_

swirtSJW:
Fixes #4285

## QA Steps

- [ ] run drush updb
- [ ] validate that it proceeds with out error.
- [ ] check your database and validate that you have no harvest_ID_runs tables

### DIFF
--- a/modules/harvest/harvest.install
+++ b/modules/harvest/harvest.install
@@ -88,3 +88,15 @@ function harvest_update_8007(&$sandbox) {
   $harvest_utility = \Drupal::service('dkan.harvest.utility');
   $harvest_utility->harvestHashUpdate();
 }
+
+/**
+ * Move entries from harvest_[ID]_runs to harvest_runs.
+ *
+ * This finishes the process started by harvest_update_8007.
+*/
+function harvest_update_8008(&$sandbox) {
+  /** @var \Drupal\harvest\HarvestUtility $harvest_utility */
+  $harvest_utility = \Drupal::service('dkan.harvest.utility');
+  $harvest_utility->harvestRunsUpdate();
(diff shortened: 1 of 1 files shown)


swirtSJW:
If you get errors trying to run this code, it is due to this https://github.com/GetDKAN/dkan/issues/4287

swirtSJW:
@janette I was going to have the review steps be to review this on PDC.... but due to this [timestamp id issue](https://github.com/GetDKAN/dkan/issues/4287) it is not runable on PDC.

I would however not wait on this fix to the 2.19 update as the issue of not running harvestRunsUpdate() can leave people's systems in pretty bad shape and not being able to detect it until too late.

janette (approved):

PR #4289, opened 2024-09-16
Adding code comments.

dmundra:
re: 20743

## Describe your changes

Adding code comments for #4185, also:
* Removed redundant code.
* Added TODOs.
* Moved code to match order in other functions.
* Removed unused parameters.
* Updated quotes.

## QA Steps

- [ ] Tests are passing

### DIFF
--- a/modules/data_dictionary_widget/src/Fields/FieldCreation.php
+++ b/modules/data_dictionary_widget/src/Fields/FieldCreation.php
@@ -61,7 +61,6 @@ class FieldCreation {
         '#type' => 'textfield',
         '#required' => TRUE,
         '#title' => t('Identifier'),
-        '#default_value' => $field_json_metadata['identifier'] ?? '',
         '#attributes' => ['readonly' => 'readonly'],
         '#default_value' => $identifier_uuid ?? '',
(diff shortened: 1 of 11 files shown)


modules/data_dictionary_widget/src/Fields/FieldCreation.php
```diff
@@ -61,7 +61,6 @@ protected static function createField(string $field, array $field_json_metadata,
         '#type' => 'textfield',
         '#required' => TRUE,
         '#title' => t('Identifier'),
-        '#default_value' => $field_json_metadata['identifier'] ?? '',
```
rovcase, line 64: Just curious as to why this was removed, possibly it wasn't necessary?

rovcase (approved):
Thanks for adding the comments for better code clarity. It's hard to tell                        where some items are removed whether or not they we're leftover/unused variables but I think QA of the feature branch will unearth that if necessary, merging this to the feature branch with respect to the fact that we have a QA ticket for feature functionality coverage. Thanks!

modules/data_dictionary_widget/src/Fields/FieldCreation.php
dmundra: Ya it was duplicate as there is another line in the same array setting the `#default_value` to the variable, see https://github.com/GetDKAN/dkan/pull/4289/files/beb8bbad16647ffa8fc03efef5b8782d297ae26b#diff-3d012ae0c7bc8097d5d36c410d4df6d57c3c89124b4ddc1ee1bd5015afd08e7aL66

PR #4307, opened 2024-10-04
#4306: properties of schema array/items do not respect property specs regarding element type

### LINKED ISSUE #4306: properties of schema array/items do not respect property specs regarding element type
### Current Behavior

Some property specs in the schema bring up special form element types in the JSON form widget.

For example:
specifying property spec "format" with a value of "uri", changes the input type to "url" instead of default type textfield (text).

This is done in the [StringHelper](https://github.com/GetDKAN/dkan/blob/2.x/modules/json_form_widget/src/StringHelper.php#L103) class.

This mechanism does not apply to property of array/items elements in the schema. These properties will be textfield (text) regardless of property spec.

### Expected Behavior

Same validation behavior as with single value properties by changing the form element type.

### Steps To Reproduce

- Create a dataset
- add a "Related document"
- enter an invalid URL for the "Related Document" like "test"
- Save dataset
- You will notice a validation error message "The attribute should match 'uri' format." All fields in the form are marked red, so you do not know which field is affected.
- This contrasts to entering a invalid URL in the "Data Dictionary" field. Here the validation error will mark the affected field with corresponding message (HTML5 validation I suppose).

_No response_

### Anything else?

_No response_

stefan-korn:
Fixes #4306 

## QA Steps

### DIFF
--- a/modules/json_form_widget/json_form_widget.services.yml
+++ b/modules/json_form_widget/json_form_widget.services.yml
@@ -16,7 +16,7 @@ services:
     class: \Drupal\json_form_widget\ObjectHelper
   json_form.array_helper:
     class: \Drupal\json_form_widget\ArrayHelper
-    arguments: ['@json_form.object_helper']
+    arguments: ['@json_form.object_helper', '@json_form.string_helper']
   json_form.widget_router:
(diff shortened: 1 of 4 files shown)


dafeder:
Closing in favor of #4382

PR #4311, opened 2024-10-08
Add option to drop datastore if there is an error on post_import

janette:
RE: 22023

## Describe your changes
Adds a config option to drop the datastore table if the post_import queue reports an error

## QA Steps

- [ ] See Jira

### DIFF
--- a/modules/datastore/src/Form/ResourceSettingsForm.php
+++ b/modules/datastore/src/Form/ResourceSettingsForm.php
@@ -54,6 +54,16 @@ class ResourceSettingsForm extends ConfigFormBase {
       '#default_value' => $this->config('datastore.settings')->get('delete_local_resource'),
       '#description' => $this->t('Delete local copy of remote files after the datastore import is complete'),
     ];
+    $form['drop_datastore_on_post_import_error'] = [
+      '#type' => 'checkbox',
+      '#title' => $this->t('Drop the datastore table if the post import queue reports an error.'),
+      '#default_value' => $this->config('datastore.settings')->get('drop_datastore_on_post_import_error'),
+      '#description' => $this->t('The datastore import queue brings in all columns as strings. The post import
+      queue will alter the table according to the data dictionary, if there is a problem during this step the
+      error will be posted to the Datastore Import Status dashboard, and the datastore table will keep all
+      data typed as strings. Check this box if you prefer that the table be dropped if there is a problem
+      in the post import stage.'),
+    ];
     $form['resource_perspective_display'] = [
       '#type' => 'select',
       '#title' => $this->t('Resource download url display'),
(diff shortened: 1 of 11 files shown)


chri5tia:
Test drop logic on post-import error.

1. Verify that datastoreService->drop() is called when a post-import error occurs, and drop_config is enabled.
Test log messages on success.

2. Verify that a notice is logged when the datastore is successfully dropped.
Test exception handling.

3. Verify that errors are logged if datastoreService->drop() throws an exception.
Test no drop when drop_config is disabled.

4. Verify that the drop() method is not called when drop_config is disabled, even if there's a post-import error.

@janette I had to change the way 'datastore.settings' is retrieved so that it comes from the config factory so it can be stored as a config object for the test. Let me know if this is wrong and/or messes anything up.

modules/datastore/src/Form/ResourceSettingsForm.php
```diff
@@ -54,6 +54,16 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
     ];
+    $form['drop_datastore_on_post_import_error'] = [
+      '#type' => 'checkbox',
+      '#title' => $this->t('Drop the datastore table if the post import queue reports an error.'),
+      '#default_value' => $this->config('datastore.settings')->get('drop_datastore_on_post_import_error'),
+      '#description' => $this->t('The datastore import queue brings in all columns as strings. The post import
+      queue will alter the table according to the data dictionary, if there is a problem during this step the
+      error will be posted to the Datastore Import Status dashboard, and the datastore table will keep all
+      data typed as strings. Check this box if you prefer that the table be dropped if there is a problem
+      in the post import stage.'),
```
rovcase, line 65: Just a thought here while it could be redundant (and as someone who is not familiar at all with this process), would it make sense and be accurate here to have this last sentence reconfirm what would happen otherwise, like: "Check this box if you prefer that the table be dropped if there is a problem, rather than have it keep all data typed as strings."

janette: The config is not concerned with the type of error at this time, it could be that there is no data typing happening so strings would be fine, but there was some other issue, like a parsing error, file encoding...

modules/datastore/tests/src/Unit/Plugin/QueueWorker/PostImportResourceProcessorTest.php
```diff
@@ -226,19 +236,130 @@ public function testPostImportProcessItemProcessorError() {
...
+    $this->assertEquals($resource->getIdentifier(), $postImportResult->getResourceIdentifier());
+    $this->assertEquals($resource->getVersion(), $postImportResult->getResourceVersion());
+  }
+
+  /**
+   * Verify Exception Handling When Datastore Drop Fails
+   *
+   * @covers ::postImportProcessItem
+   */
+  public function testExceptionHandlingWhenDatastoreDropFails() {
```
rovcase: testExceptionHandlingWhenDatastoreDropFails() is identical 1:1 to testDatastoreDropOnPostImportError(), should one be removed in favor of another?

rovcase (changes requested):
Hello wowee good stuff (very helpful to look into the test files too). Just one minor possibly redundant question as well as one piece of code cleanup. Thanks!

modules/datastore/tests/src/Unit/Plugin/QueueWorker/PostImportResourceProcessorTest.php
janette: Yes I think we can remove that second one.

rovcase (approved):
LGTM, thanks!

PR #4320, opened 2024-10-21
Add Drush command to apply a dictionary to a resource

paul-m:
Fixes [issue#]

## TO DO

While testing we noticed that applying the data dictionary via the `dictionaryEnforcer `would apply the dictionary but would not update the status under the post import column of the dashboard (`/admin/dkan/datastore/status`). A better approach might be to utilize the `PostImport` service and call `processResource`. By using `PostImport` instead of the `dictionaryEnforcer` we are able capture the post import status.

## Describe your changes

Adds a drush command: `dkan:datastore:apply-dictionary`

This command applies the data dictionary to the specified dataset, as configured in the site, and as specified by `describedBy`.

## QA Steps

- Set the site to not use data dictionaries.
- Add a data dictionary to a dataset.
- Set the site to use the data dictionary.
- Run `drush dkan:datastore:apply-dictionary` for the dataset.
- Verify that the data dictionary is displayed for the dataset.
- If indexing was specified in the dictionary, verify that the indexing was applied to the database table for the resource.


paul-m:
This is postponed until we get more clarity on how data dictionaries will operate in terms of type enforcing, indexing, and key definitions.

paul-m:
Marked as ready for review because we decided to go forward with this regardless of scope on data dictionaries.

dafeder:
@paul-m wondering if you're able to point to some good sample data to test this with

dafeder:
Also a little unsure what "Verify that the data dictionary is displayed for the dataset" means exactly

modules/datastore/src/Commands/DictionaryCommands.php
```diff
@@ -0,0 +1,81 @@
...
+   * @command dkan:datastore:apply-dictionary
+   */
+  public function applyDictionary(string $resource_identifier) {
+    if ($this->dataDictionaryDiscovery->getDataDictionaryMode() === DataDictionaryDiscoveryInterface::MODE_NONE) {
+      $this->logger()->notice('This site is not configured to use data dictionaries.');
+      return DrushCommand::SUCCESS;
+    }
+    try {
+      $this->dictionaryEnforcer->process(
+        $this->resourceMapper->get($resource_identifier)
```
kaise-lafrai, line 69: I believe we need to provide the $perspective as an argument because the table name generation is based on the resource_identifier, perspective and version. Perspective if not provided would default to source. In my testing the applyDictionary failed with the following:

```
  SQLSTATE[42S02]: Base table or view not found: 1146 Table 'db.datastore_9683ce1b537c79c9ef21c7f998819433' doe  
  sn't exist: SHOW FULL COLUMNS FROM "datastore_9683ce1b537c79c9ef21c7f998819433"; Array
```
  
  Once I passed in the perspective ie: ResourceLocalizer::LOCAL_FILE_PERSPECTIVE the data dictionary applied correctly.

kaise-lafrai: example:

```
    try {
      $this->dictionaryEnforcer->process(
        $this->resourceMapper->get($resource_identifier, ResourceLocalizer::LOCAL_FILE_PERSPECTIVE)
      );
    }
```

kaise-lafrai:
@paul-m @dafeder @janette Not sure of the original scope of this ticket and how in originated, when this new drush command is executed, I would assume the dashboard [post import] would reflect the new status. Right now as is this PR only applies the DD and it's results are displayed as a logger notice in the command line (not the dashboard). Should this PR/ISSUE expand to capture post import results and display it on the dashboard?

kaise-lafrai (commented):
A few things need to be addressed. Thank you!

TheETupper:
Can we close this out as a close import instead of a data dictionary command?

PR #4333, opened 2024-11-07
#4332: Fix import status dashboard pagination

### LINKED ISSUE #4332: Import status dashboard pagination issue - follow-up to #4317
### Current Behavior

since #4317 the pagination of the import status dashboard is not correct anymore, if there are datasets filtered out due to non-importable distributions. Pagination is still pointing to all datasets irrespective of the filter. This may lead to empty pages in the pager and pages with less items than pagination items per page setting.

### Expected Behavior

Pager should reflect the actual state of filtered datasets.

### Steps To Reproduce

Have more than 10 datasets, some of them having no importable distributions (like PDF, DOC, etc.).
See that pagination of import status dashboard is not correct.

_No response_

### Anything else?

_No response_

stefan-korn:
Fixes #4332 

## Describe your changes

Solution is maybe impacting performance, since all the datasets are requested before paging. But did not see how to fix this issue otherwise.

## QA Steps

### DIFF
--- a/modules/datastore/src/Form/DashboardForm.php
+++ b/modules/datastore/src/Form/DashboardForm.php
@@ -239,6 +239,31 @@ class DashboardForm extends FormBase {
   }
 
   /**
+   * Filter datasets with importable distributions
+   *
+   * @param array $dataset_uuids
+   *   Datasets to be filtered.
+   *
+   * @return array
+   *   Filtered datasets.
+   */
+  protected function filterImportableDatasets($dataset_uuids) {
+    $datasets_filtered = [];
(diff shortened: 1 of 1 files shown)


paul-m:
Thanks!

At minimum, this needs a unit test of `filterImportableDatasets()` to prove that it does the right thing. This could just mock `datasetInfo->gather()` to return an info array that would be `array_filter()`'d to nothing.

Bonus after that is to create a dataset with nothing to import, and then show that it does not display on the status page.

dafeder:
I think we're going to close this in favor of #4429. Thanks for surfacing @stefan-korn

PR #4338, opened 2024-11-08
Lifecycle cleanup

paul-m:
Random musings and @todos while trying to hunt down an event-related issue.

## QA Steps


modules/metastore/metastore.module
```diff
@@ -60,22 +60,23 @@ function resource_mapper_new_revision() {
...
  */
-function metastore_data_lifecycle(EntityInterface $entity, string $stage) {
-
-  if (metastore_entity_is_valid_item($entity)) {
-    $itemFactory = \Drupal::service('dkan.metastore.metastore_item_factory');
-    $metastoreItem = $itemFactory->wrap($entity);
-    \Drupal::service('dkan.metastore.lifecycle')->go($stage, $metastoreItem);
+function metastore_data_lifecycle(array $entities, string $stage): void {
+  /** @var \Drupal\metastore\LifeCycle\LifeCycle $life_cycle */
+  $life_cycle = \Drupal::service('dkan.metastore.lifecycle');
```
dafeder: "lifecycle" is one word. Should not have been camelcase in the first place. But `life_cycle` looks wrong

modules/metastore/src/LifeCycle/LifeCycle.php
```diff
@@ -115,9 +117,57 @@ public function __construct(
...
+
+      // If the type and bundle are correct, return true.
+      if ($entity->getEntityTypeId() != $storageEntityType) {
+        return FALSE;
+      }
+      if (in_array($entity->bundle(), $storageBundles)) {
+        return TRUE;
+      }
+    }
+    return FALSE;
```
dafeder:
```suggestion
    if ($entity->getEntityTypeId() !== 'node' || $entity->bundle() !== 'data') {
      return FALSE;
    }
    return TRUE;
```
If we're just going to hard code "node" and "data" anyway all the other stuff is redundant I think?

OTOH the whole function is kind of redundant because the `wrap()` method will instantiate a new `Data` object, which validates the entity in its constructor with exactly the same criteria. The issue is that that validation function doesn't return true or false, it throws an exception if invalid and does nothing if valid. Using a try/catch around `wrap()` in those entity hook functions is a icky, so an idea I had was:

a) Make `Data::validate()` public and static, and have it return `bool` rather than `void`/throw.
b) throw the exception in the constructor if validate fails.
c) Call Data::validate in the .module file rather than your new `LifeCycle::entityIsValidItem()`. 

It's not exactly a slam-dunk. This whole area is not great. Maybe I'm missing another cleaner approach. But at minimum let's simplify that validation function.

dafeder (changes requested):

dafeder (approved):

PR #4341, opened 2024-11-15
4336: Return 400 error on invalid JSON sent to the datastore API

### LINKED ISSUE #4336: Invalid JSON causes datastore API to throw 500
### Current Behavior

When we submit valid JSON data to the Datastore API but in a form that does not validate against the query schema, as expected we get a 400 error and some error output. However, submitting completely invalid JSON results in a 500 error.

This issue is specific to the "resource query" datastore endpoint. Submitting invalid JSON to the Metastore API, for instance, will return a 400 "Invalid JSON" response. The datastore  should have the same behavior.

### Expected Behavior

Submitting completely invalid JSON, such as 
```json
{
  "foo": bar,
}
```
should result in a 400 error and some output.

### Steps To Reproduce

```http
POST https://demo.getdkan.org/api/1/datastore/query/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/0 HTTP/1.1

{
  "conditions": [
}
```

```shell
```

### Anything else?

_No response_

janette:
Fixes #4336

## Describe your changes
Better response for invalid JSON

## QA Steps

- [ ] Run the following POST
```
POST https://demo.getdkan.org/api/1/datastore/query/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/0 HTTP/1.1

{
  "conditions": [
}
```
- [ ] Confirm you get a 400 error and "Invalid JSON" message
- [ ] Run the following POST

```
POST https://demo.getdkan.org/api/1/datastore/query/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/0 HTTP/1.1

{
    "format": "csv
    "limit": 10
}
```
- [ ] Confirm you get a 400 error and "Invalid JSON" message


modules/datastore/src/Controller/AbstractQueryController.php
```diff
@@ -221,6 +221,9 @@ protected function extractMetastoreDependencies(DatastoreQuery $datastoreQuery):
   protected function buildDatastoreQuery(Request $request, mixed $identifier = NULL) {
```
paul-m: `\Drupal\datastore\Controller\QueryDownloadController::buildDatastoreQuery()` overrides this... Both call `getPayloadJson()` which is probably where we should throw the exception.

So: Move this checking to either `getPayloadJson()` or `fixTypes()`, and add a bad JSON test for both query controller and download query controller.

paul-m (changes requested):

modules/datastore/src/Controller/AbstractQueryController.php
kaise-lafrai: Done

modules/datastore/src/Controller/AbstractQueryController.php
```diff
@@ -329,6 +329,11 @@ public static function fixTypes($json, $schema) {
     $data = json_decode($json);
     $validator = new Validator();
     $validator->coerce($data, json_decode($schema));
+
+    if ($data === NULL) {
```
paul-m: I'd suggest
```
  public static function fixTypes($json, $schema) {
    if ($data = json_decode($json) !== NULL) {
      $validator = new Validator();
      $validator->coerce($data, json_decode($schema));
      return json_encode($data, JSON_PRETTY_PRINT);
    }
    throw new \InvalidArgumentException("Invalid JSON");
  }
```
...or similar so that we don't create a `Validator` unless we need to.

paul-m (changes requested):

paul-m (approved):

PR #4342, opened 2024-11-15
Stop passing empty values to retrieveDownloadUrlFromResourceMapper

janette:
This is annoying
Screenshot 2024-11-14 at 7 53 43 PM

When removing old revisions, many nodes without downloadUrls were getting passed to LifeCycle::distributionLoad
which then calls retrieveDownloadUrlFromResourceMapper on non-existent $resourceIdentifier values.

## Describe your changes
Changing the check from `isset` to `!empty` avoids sending non-distribution nodes through the process

## Steps to reproduce

- [ ] Add a logger in `distributionLoad` to confirm empty $resourceIdentifier values are being sent through retrieveDownloadUrlFromResourceMapper
Screenshot 2024-11-14 at 8 01 33 PM
 

### DIFF
--- a/modules/metastore/src/LifeCycle/LifeCycle.php
+++ b/modules/metastore/src/LifeCycle/LifeCycle.php
@@ -189,11 +189,10 @@ class LifeCycle {
 
     $downloadUrl = $metadata->data->downloadURL;
 
-    if (!empty($downloadUrl) && !filter_var($downloadUrl, FILTER_VALIDATE_URL)) {
-      $resourceIdentifier = $downloadUrl;
+    if (!empty($downloadUrl) && filter_var($downloadUrl, FILTER_VALIDATE_URL) === FALSE) {
       $ref = NULL;
       $original = NULL;
-      [$ref, $original] = $this->retrieveDownloadUrlFromResourceMapper($resourceIdentifier);
+      [$ref, $original] = $this->retrieveDownloadUrlFromResourceMapper($downloadUrl);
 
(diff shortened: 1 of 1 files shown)


modules/metastore/src/LifeCycle/LifeCycle.php
```diff
@@ -189,7 +189,7 @@ protected function distributionLoad(MetastoreItemInterface $data) {
 
     $downloadUrl = $metadata->data->downloadURL;
 
-    if (isset($downloadUrl) && !filter_var($downloadUrl, FILTER_VALIDATE_URL)) {
+    if (!empty($downloadUrl) && !filter_var($downloadUrl, FILTER_VALIDATE_URL)) {
```
paul-m: How about explicitly checking the return value of `filter_var()` for the same reason? So:
```
if (!empty($downloadUrl) && (filter_var($downloadUrl, FILTER_VALIDATE_URL) === FALSE)) {
```

modules/metastore/src/LifeCycle/LifeCycle.php
```diff
@@ -189,7 +189,7 @@ protected function distributionLoad(MetastoreItemInterface $data) {
...
       $resourceIdentifier = $downloadUrl;
```
paul-m: Also, $resourceIdentifier is only used once within that if block, so we could remove it and just pass $downloadUrl.

paul-m (changes requested):

paul-m (approved):

PR #4177, opened 2024-05-04
Reduce dependency on getdkan/contracts (part 1)

paul-m:
Partially fixes https://github.com/GetDKAN/dkan/issues/3821

|Old interface|New interface|
|---|---|
|`\Contracts\CountableInterface`|`\Countable`|
|`\Contracts\HydratableInterface`|`\Procrastinator\HydratableInterface`|
|`\Contracts\ParserInterface`|`\CsvParser\Parser\ParserInterface`|

`\Contracts\RetrieverInterface` and `\Contracts\StorerInterface` remain in `getdkan/contracts` because there is no clear place to move them. They're needed within multiple libraries. Refactor this in a followup.

We also need to keep `getdkan/contracts` as a dependency in many places because some libraries use `Contracts\Mock\Storage\Memory` for tests. Refactor this in a followup.

Our changes to `Drupal\common\Storage\Query`... See the @todo which asks if we should just make a QueryInterface.

## Merge Steps
In order to make all these changes, we have to modify the following libraries, maybe in this order. 

Each PR is a branch named `no-contracts`, some of which have interdependencies on each other. These PRs will need to be updated to require the release of each of these dependencies.

- [x] GetDKAN/contracts#16 - Deprecates all the interfaces changed above (for now)
- [x] GetDKAN/csv-parser#20 - Has no dependencies on `getdkan/contracts` after PR.
- [x] GetDKAN/procrastinator#23 - Uses `StorerInterface` and `RetrieverInterface`. We need to figure out where to put storage-related interfaces in round 2 of this refactor.
- [x] GetDKAN/file-fetcher#39 - Uses `Contracts\Mock\Storage\Memory` for testing. Refactor in round 2.
- [x] GetDKAN/harvest#20 - Uses `Contracts\Mock\Storage\Memory` for testing. Refactor in round 2.

PR #4346, opened 2024-11-20
DKAN-4287 Make harvest_run id not a primary key.

### LINKED ISSUE #4287: Converting runs from old harvest_ID_runs to harvest_runs fails if duplicate time stamps
### Current Behavior

When updating to DKAN 2.19 from a lower version, if two different harvest_ID_runs tables happen to have the same timestamp a sql error is thrown because the timestamp is treated as the unique identifier.  This is unlikely since it is only a span of 1 second, but it is possible to encounter in the wild.

```
> >  [notice] Converting runs for home_health__data
> >  [error]  Drupal\Core\Database\IntegrityConstraintViolationException: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '1599570670' for key 'PRIMARY': INSERT INTO "harvest_runs" ("id", "harvest_plan_id", "data", "extract_status") VALUES (:db_insert_placeholder_0, :db_insert_placeholder_1, :db_insert_placeholder_2, :db_insert_placeholder_3); Array
> > (
> >     [:db_insert_placeholder_0] => 1599570670
> >     [:db_insert_placeholder_1] => home_health__data
> >     [:db_insert_placeholder_2] => {"plan":"{\"identifier\":\"home_health__data\",\"extract\":{\"type\":\"\\\\Drupal\\\\pqdc\\\\Harvest\\\\ETL\\\\Extract\\\\DataJson\",\"uri\":\"file:\\\/\\\/\\\/mnt\\\/tmp\\\/data.json\"},\"transforms\":[],\"load\":{\"type\":\"\\\\Drupal\\\\harvest\\\\Load\\\\Dataset\"}}","status":[],"errors":{"extract":"Error decoding JSON."}}
> >     [:db_insert_placeholder_3] => FAILURE
> > )
> >  in Drupal\mysql\Driver\Database\mysql\ExceptionHandler->handleExecutionException() (line 45 of /var/www/html/docroot/core/modules/mysql/src/Driver/Database/mysql/ExceptionHandler.php). 
> >  [error]  SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '1599570670' for key 'PRIMARY': INSERT INTO "harvest_runs" ("id", "harvest_plan_id", "data", "extract_status") VALUES (:db_insert_placeholder_0, :db_insert_placeholder_1, :db_insert_placeholder_2, :db_insert_placeholder_3); Array
> > (
> >     [:db_insert_placeholder_0] => 1599570670
> >     [:db_insert_placeholder_1] => home_health__data
> >     [:db_insert_placeholder_2] => {"plan":"{\"identifier\":\"home_health__data\",\"extract\":{\"type\":\"\\\\Drupal\\\\pqdc\\\\Harvest\\\\ETL\\\\Extract\\\\DataJson\",\"uri\":\"file:\\\/\\\/\\\/mnt\\\/tmp\\\/data.json\"},\"transforms\":[],\"load\":{\"type\":\"\\\\Drupal\\\\harvest\\\\Load\\\\Dataset\"}}","status":[],"errors":{"extract":"Error decoding JSON."}}
> >     [:db_insert_placeholder_3] => FAILURE
> > )
> >  
> >  [error]  Update failed: harvest_update_8008 
> >  [notice] Update started: metastore_update_8009
> >  [notice] Updated 0 dictionaries. If you have overridden DKAN's core schemas,
> >     you must update your site's data dictionary schema after this update. Copy
> >     modules/contrib/dkan/schema/collections/data-dictionary.json over you local
> >     site version before attempting to read or write any data dictionaries.
> >  [notice] Update completed: metastore_update_8009
> >  [notice] Update started: metastore_admin_update_8012
> >  [notice] Update completed: metastore_admin_update_8012
>  [error]  Update aborted by: harvest_update_8008 
>  [error]  Finished performing updates. 
```

### Expected Behavior

The migration of data from one table to another should happen without error.

### Steps To Reproduce

1. Have a data setup where you have harvest_ID_runs that contain the same timestamps.
2. run update.php, or `drush upddb` or `drush dkan:harvest:update`
3. See errors

_No response_

### Anything else?

This may be too unlikely a scenerio to add a try catch block `HarvestUtility::convertRunTable()` but I will at least provide a drush sqlc command to undo any duplicated IDs.

Discussion in [CA Slack](https://civicactions.slack.com/archives/C0257EEDZ1U/p1726244779586979?thread_ts=1726242709.074259&cid=C0257EEDZ1U)

swirtSJW:
Fixes #4287

## Describe your changes
This PR does three major things

1. Converts harvest runs from using a timestamp to using the harvest_plan and id as the unique combination to reverence a harvest run.  This gets rid of the problem of multiple harvest runs possibly having the same 1 second increment timestamp
2. Converts any previous run data to use the new format
3. Creates new harvests in the new format.

### From existing data
- git checkout 2.18.3 (the tag before runs became an entity)
- init the site
- ddev drush dkan:sample-content:create
- ddev drush cron
- [ ]  validate no errors show in terminal from these commands.
- git checkout DKAN-4287-de-key-harvest-run 
- ddev drush deploy
- [ ] validate no related errors - should look similar to https://github.com/GetDKAN/dkan/pull/4346#issuecomment-2568024882
- visit /admin/dkan/harvest
- [ ] validate harvests show as success
- Drill into import status
- [ ] validate import status is successful
- Drill into datasets 
- [ ] validate datasets appear to function correctly.
- view entries in the 'harvest_runs' database table.  
- [ ] validate that they have uuids
- [ ] validate all columns have content in them
- [x]  Further proof of this working: [PDC has been using the patch](https://github.cms.gov/PQDC/pqdc-dkan/blob/develop/src/patches/dkan_4287_primary_key.patch) from the PR for the past 2 months without issue.

### From new data
- git checkout DKAN-4287-de-key-harvest-run 
- init the site
- ddev drush deploy
- ddev drush dkan:sample-content:create
- [ ] validate no related errors - should look similar to https://github.com/GetDKAN/dkan/pull/4346#issuecomment-2568024882
- visit /admin/dkan/harvest
- [ ] validate harvests show as success
- Drill into import status
- [ ] validate import status is successful
- Drill into datasets 
- [ ] validate datasets appear to function correctly.
- view entries in the 'harvest_runs' database table.  
- [ ] validate that they have uuids
- [ ] validate all columns have content in them

###


modules/harvest/src/Entity/HarvestRunRepository.php
```diff
@@ -243,16 +243,19 @@ public function getExtractedUuids(string $plan_id, string $run_id): array {
...
    *   Plan ID.
-   * @param string $run_id
-   *   Run ID, which is a timestamp.
+   * @param string $timestamp
+   *   The timestamp for the run. Formerly the id.
    *
    * @return \Drupal\harvest\HarvestRunInterface|\Drupal\Core\Entity\EntityInterface|null
    *   The loaded entity or NULL if none could be loaded.
+   *
+   * @deprecated in dkan:2.19.11 and is removed from dkan:3.0.0 Use HarvestService::load().
```
swirtSJW: It is possible that this can't be deprecated.   Everywhere it is still called, there is no access to the  actual ID.  Mainly in tests where it is pulling previous specific run instances.

swirtSJW:
There are still two test failures that are not immediately clear to me why they are failing 
image

swirtSJW:
Got it down to just one test failure 
testGetExtractedUuids
Failed asserting that two arrays are equal.

dafeder:
@swirtSJW I don't see any test failures at the moment?

swirtSJW:
@paul-m Can you take a first pass at code review on this?  
Tests are passing but I will hold off on adding QA steps until I perform this whole upgrade on PDC.  Right now this is successful on vanilla but has not been fully run on an existing site.

modules/harvest/harvest.install
```diff
@@ -28,6 +33,64 @@ function harvest_requirements($phase): array {
...
+ *
+ * @return array
+ *   Elements from the row ['id', 'harvest_plan_id', 'data', 'extract_status'].
+ */
+function harvest_read_harvest_run(string $table_name_temp, string $time_id): array {
+  $connection = Database::getConnection();
+  $query = $connection->select($table_name_temp, 'hrt')
+    ->fields('hrt', ['id', 'harvest_plan_id', 'data', 'extract_status'])
+    ->condition('id', $time_id, '=');
+  //  ->where($time_id, 'id', '=');
```
swirtSJW: removing this.

paul-m (commented):
Under the 2.x branch, I used sample_content to generate a harvest to run (`drush dkan:sample-content:create`), and then ran it (using cron).
Then I switched to this branch and ran `drush updb`.
Everything seemed to work... I did notice that the one harvest run in my `harvest_runs` table didn't have a UUID.

So I went ahead and did a new installation with this branch, and then ran the sample content harvest, and it did have a UUID.

So perhaps the update process doesn't generate UUIDs the way just creating an entity does...

Also I'd suggest that the utility functions, such as `harvest_get_temp_run_ids()` in `harvest.install` should be in `HarvestUtility.` That way it's all in one place and labeled as part of the update process.

modules/harvest/src/Entity/HarvestRun.php
```diff
@@ -46,15 +46,12 @@
  *   admin_permission = "administer harvest_run",
  *   entity_keys = {
  *     "id" = "id",
- *     "label" = "id",
+ *     "label" = "ID",
```
paul-m: 'label' here tells the entity system which field to use *as* a label, not what the label is. We don't really have one, so we should just leave it as id (lower case).

modules/harvest/src/HarvestPlanListBuilder.php
```diff
@@ -74,9 +74,9 @@ public function buildRow(EntityInterface $entity) {
     $harvest_plan_id = $entity->get('id')->getString();
     $run_entity = NULL;
 
-    if ($run_id = $this->harvestService->getLastHarvestRunId($harvest_plan_id)) {
+    if ($run_id = $this->harvestService->runRepository->getLastHarvestRunId($harvest_plan_id)) {
```
paul-m: We have a repository in scope here already, right?

modules/harvest/src/HarvestPlanListBuilder.php
```diff
@@ -74,9 +74,9 @@ public function buildRow(EntityInterface $entity) {
...
       // There is a run identifier, so we should get that info.
-      $run_entity = $this->harvestRunRepository->loadEntity($harvest_plan_id, $run_id);
+      $run_entity = $this->harvestRunRepository->runStorage->load($run_id);
```
paul-m: Inject the harvestrun storage to the list builder rather than making runstorage public and accessing it here.

modules/harvest/src/Entity/HarvestRunRepository.php
```diff
@@ -24,7 +24,7 @@ class HarvestRunRepository {
    *
    * @var \Drupal\Core\Entity\EntityStorageInterface
    */
-  private EntityStorageInterface $runStorage;
+  public EntityStorageInterface $runStorage;
```
paul-m: I can see making this protected for some reason, but public so you can access it is a bad pattern. Inject it elsewhere.

modules/harvest/src/Entity/HarvestRun.php
swirtSJW: Good catch.  Do we even need to specify a label?   Lots of entities don't.

modules/harvest/src/Entity/HarvestRunRepository.php
swirtSJW: This pattern strikes me as odd though that HarvertRunRepository's concern is " Convenient queries for harvest_run entities." and yet it offers no way to load a single harvest run.  If the pattern of exposing ->runStorage is bad, should we add a public method to HarvestRunRepository that is a  public passthrough for  ->runStorage->load()?

Otherwise it is kind of like giving you a set of keys that has every key you'll need to get into things, except your own front door.

paul-m: `$runstorage` is a service. If you need it somewhere else, inject it there. The reason is that if `HarvestRunRepository` promises to provide the service then it will always have that dependency, and will always have to have that public property, regardless of whether it needs it internally. Further down the line, if we deprecate more of the repository class, we're left providing a service that really should just be an injected service for those depending on it.

So, you're using the storage service in `HarvestPlanListBuilder`, where you could inject it in order to load a run record. Alternately, you could just say `HarvestRun::load($id)` (or `::loadMultiple()`), but the injected storage service is probably more of a best practice so the entity class doesn't have to look up its own storage every time.

modules/harvest/src/Entity/HarvestRun.php
swirtSJW: Resolved.  I changed it back to id.

modules/harvest/src/HarvestPlanListBuilder.php
swirtSJW: Good catch... I got too far down the rabbit hole and forgot to look out.  
This is resolved.   I also ended up removing the harvestService because it was no longer used in the listbuilder.

modules/harvest/src/HarvestPlanListBuilder.php
swirtSJW: Done.

modules/harvest/src/Entity/HarvestRunRepository.php
swirtSJW: Good explanation.  This has been resolved.

swirtSJW:
> Everything seemed to work... I did notice that the one harvest run in my `harvest_runs` table didn't have a UUID

So it looks like a simple database write did not kick off the magic that would add a uuid that a normal drupal save would.  So I intentionally added it.

swirtSJW:
I am stumped by this test failure. 
```
There was 1 error:

1) Drupal\Tests\datastore\Unit\Plugin\QueueWorker\ImportJobTest::testBasics
TypeError: Drupal\datastore\Plugin\QueueWorker\ImportJob::getParser(): Return value must be of type Contracts\ParserInterface, CsvParser\Parser\Csv returned
```

It seems completely unrelated to the code in this PR.  

Beyond this test @paul-m I made all the changes you requested.

dafeder:
Oof we must have messed something up here with the contracts stuff @paul-m and released a breaking change (or we have too open a constraint on some dependency). I'm seeing this in other builds as well.

paul-m:
Over in https://github.com/GetDKAN/dkan/pull/4177 we deprecated some interfaces, which led to the fails.

swirtSJW:
Here is the run from PDC with existing data
```
 [notice] Database updates start.
 ----------------- ------------------------------- --------------- ----------------------------------------------------- 
  Module            Update ID                       Type            Description                                          
 ----------------- ------------------------------- --------------- ----------------------------------------------------- 
  dkan              9002                            hook_update_n   9002 - Enable data dictionary widget.                
  harvest           8006                            hook_update_n   8006 - Ensure the entity manager knows about         
                                                                    harvest_run entities.                                
  harvest           8007                            hook_update_n   8007 - Move data from harvest_ID_hashes tables to    
                                                                    harvest_hash entity. This will move all harvest      
                                                                    hash information to the updated schema, including    
                                                                    data which does not have a corresponding hash plan   
                                                                    ID. Outdated tables will be removed.                 
  harvest           8008                            hook_update_n   8008 - Move entries from harvest_[ID]_runs to        
                                                                    harvest_runs. This finishes the process started by   
                                                                    harvest_update_8007.                                 
  harvest           8009                            hook_update_n   8009 - Update harvest_run schema to add timestamp,   
                                                                    uuid, and true id. @see                              
                                                                    https://github.com/GetDKAN/dkan/issues/4287          
  harvest           8010                            hook_update_n   8010 - Move data from temp table back into           
                                                                    harvest_run. @see                                    
                                                                    https://github.com/GetDKAN/dkan/issues/4287          
  harvest           8011                            hook_update_n   8011 - Move entries from harvest_[ID]_runs to        
                                                                    harvest_runs. This finishes the process started by   
                                                                    harvest_update_8007 and re-runs 8008.                
  metastore         8009                            hook_update_n   8009 - Update existing data dictionary nodes to use  
                                                                    corrected schema.                                    
  metastore_admin   8012                            hook_update_n   8012 - Reinstall DKAN Metastore Admin configuration  
                                                                    to include new dkan menu items.                      
  search_api        fix_index_dependencies_orders   post-update     Re-save Search API index configurations to fix       
                                                                    dependencies order.                                  
 ----------------- ------------------------------- --------------- ----------------------------------------------------- 


 // Do you wish to run the specified pending updates?: yes.                                                             

> >  [notice] Update started: harvest_update_8006
> >  [notice] Update completed: harvest_update_8006
> >  [notice] Update started: harvest_update_8007
> >  [notice] Converting hashes for dialysis__data
> >  [notice] Converting hashes for dialysis__data_umkecc
> >  [notice] Converting hashes for home_health__data
> >  [notice] Converting hashes for home_health__data_two
> >  [notice] Converting hashes for home_health__HomeHealthCAHPS
> >  [notice] Converting hashes for hospice__data
> >  [notice] Converting hashes for hospital__data
> >  [notice] Converting hashes for inpatient__data
> >  [notice] Converting hashes for long_term_care_hospital__data
> >  [notice] Converting hashes for nursing_home__data
> >  [notice] Converting hashes for office_visit_cost__data
> >  [notice] Converting hashes for physician__data
> >  [notice] Converting hashes for supplier__data
> >  [notice] Update completed: harvest_update_8007
> >  [notice] Update started: harvest_update_8008
> >  [notice] Update completed: harvest_update_8008
> >  [notice] Update started: harvest_update_8009
> >  [notice] Table harvest_runs moved to harvest_runs_temp. 
> > Old harvest_run entity removed. 
> > New harvest_run entity installed. 
> > 
> >  [notice] Update completed: harvest_update_8009
> >  [notice] Update started: harvest_update_8010
> >  [notice] Processed: 0/0.
> > Data in harvest_runs updated to new schema:
> > Temporary table dropped.
> > 
> >  [notice] Update completed: harvest_update_8010
> >  [notice] Update started: dkan_update_9002
> >  [notice] Update completed: dkan_update_9002
> >  [notice] Update started: harvest_update_8011
> >  [notice] Converting runs for dialysis__data
> >  [notice] Converting runs for dialysis__data_umkecc
> >  [notice] Converting runs for home_health__data
> >  [notice] Converting runs for home_health__data_two
> >  [notice] Converting runs for home_health__HomeHealthCAHPS
> >  [notice] Converting runs for hospice__data
> >  [notice] Converting runs for hospital__data
> >  [notice] Converting runs for inpatient__data
> >  [notice] Converting runs for long_term_care_hospital__data
> >  [notice] Converting runs for nursing_home__data
> >  [notice] Converting runs for office_visit_cost__data
> >  [notice] Converting runs for physician__data
> >  [notice] Converting runs for supplier__data
> >  [notice] Harvest plan specific run tables coalesced into table harvest_runs.
> >  [notice] Update completed: harvest_update_8011
> >  [notice] Update started: metastore_update_8009
> >  [notice] Updated 0 dictionaries. If you have overridden DKAN's core schemas,
> >     you must update your site's data dictionary schema after this update. Copy
> >     modules/contrib/dkan/schema/collections/data-dictionary.json over you local
> >     site version before attempting to read or write any data dictionaries.
> >  [notice] Update completed: metastore_update_8009
> >  [notice] Update started: metastore_admin_update_8012
> >  [notice] Update completed: metastore_admin_update_8012
> >  [notice] Update started: search_api_post_update_fix_index_dependencies_orders
> >  [notice] Update completed: search_api_post_update_fix_index_dependencies_orders
>  [success] Finished performing updates.
```

swirtSJW:
harvest_runs table after the updates of existing data.
image

swirtSJW:
I thought I had this buttoned up but just discovered the list of imports is empty
image

swirtSJW:
@paul-m this has been a long time coming but this is now ready for review.

paul-m (approved):
Just one hiccup: Running `ddev drush deploy` results in a bunch of config dependency errors, when Drush tries to do a `cim`. Other than that, no issues with the testing steps.

Tests pass, test steps work. Great work @swirtSJW !

PR #4347, opened 2024-11-22
Move and remove elements from JSON arrays in form

### LINKED ISSUE #4335: Remove or Reorder Objects in An Array
More discovery work is needed to determine the approach we want to take to solve this.

--

If there is an array in the schema you can only add and remove the last item in the dataset form.

For instance, if there are 3 distributions, you cannot change their order or remove just the first or middle one. 

A user should be able to remove or reorder any object in the list

dafeder:
Fixes #4335 

## Describe your changes

Fairly extensive changes to how array properties are handled in JSON Form Widget. Arrays, whether "simple" (an array of strings or numbers) or "complex" (arrays of objects) will contain one more level of nested fieldsets. This is because each array item now has its own action buttons, so that array items can be removed or moved in the order. Previously we only had action buttons at the end of the array, allowing us to "add one" or "remove one".

Because the autoselect widgets expect a built array fieldset and then rewrite it, there are also a number of changes to that logic. A risk - we do not really have full functional test coverage for these elements. Also, now that I understand the code flow a bit better, it does make me wonder if there's a better approach that avoids building all these nested fieldsets on the first pass only to flatten and replace them on the second pass with WidgetRouter.

## QA Steps

- [ ] Create a new Dataset
- [ ] Add several distributions
- [ ] Re-order them, remove distributions from the beginning or middle of the list
- [ ] Do the same for "References"
- [ ] Confirm that the keyword and topic fields continue to work as expected
- [ ] Save the dataset, and view the JSON in the API to confirm all works as expected
- [ ] Edit an existing dataset, and try similar tests

You may want to remove some fields from the distribution schema to make that form element shorter - Keep it to just title, description and downloadURL for instance.

**Note** on codeclimate failure - ArrayHelper is pretty big now but I don't see an obvious way to split it up that would make the code much more readable. Open to suggestions but IMO let's just let it slide for now and revisit if the method count gets any higher.


stefan-korn:
@dafeder : Sorry, didn't want to create noise here, i just wanted to test this against latest 2.x and therefore created the PR in my fork.

dafeder:
No worries @stefan-korn. I've been meaning to get back to this but I got kind of stumped on the file element, it behaves very differently from the other field elements.

dafeder:
Closed in favor of #4472

PR #4348, opened 2024-11-26
Make response stream caching max-age a configuration

zedmonds96:
Fixes see 21638

## Describe your changes

- Added `datastore_response_stream_max_age` to `datastore.settings.yml` with a default of 3600 seconds.
- Updated `DatastoreSettingsForm` to include a configurable field for response stream max-age.
- Modified `QueryDownloadController` to use the configured `response_stream_max_age` instead of a hardcoded value.

## QA Steps

- [ ] Access the Datastore Settings form.
- [ ] Modify the Response Stream Max-Age value and save the form.
- [ ] Confirm that the new value is saved in the configuration
- [ ] If no configuration is set, ensure the default 3600 seconds is applied.

- [N/A] I have updated or added documentation


dafeder:
@zedmonds96 set this back to draft as tests are failing. Are you still working on this?

modules/datastore/src/Controller/QueryDownloadController.php
```diff
@@ -31,9 +26,8 @@ public function __construct(QueryService $queryService, DatasetInfo $datasetInfo
     // We do not want to cache streaming CSV content internally in Drupal,
     // because datasets can be very large. However, we do want CDNs to be able
     // to cache the CSV stream for a reasonable amount of time.
-    // @todo Replace this constant with some form of customizable caching
-    //   strategy.
-    $this->cacheMaxAge = static::RESPONSE_STREAM_MAX_AGE;
+    $config = $configFactory->get('datastore.settings');
+    $this->cacheMaxAge = $config->get('response_stream_max_age') ?: self::DEFAULT_RESPONSE_STREAM_MAX_AGE;
```
paul-m: The default value for the response stream max age is defined in `datastore.settings.yml`, so we don't need the constant here as a default value. The principle is that it's defined in only one place.

However, that also means we need an update hook to set the value in the site config.

Here's an example in the metastore module: https://github.com/GetDKAN/dkan/blob/2.x/modules/metastore/metastore.install#L113

Drupal docs here: https://www.drupal.org/docs/drupal-apis/update-api/updating-configuration#s-example-1-adding-a-configuration-key

Thanks!

modules/datastore/tests/src/Unit/Controller/QueryDownloadControllerTest.php
```diff
@@ -346,7 +347,7 @@ public function testStreamedBadSchema() {
    * @return \MockChain\Chain
    *   MockChain chain object.
    */
-  private function getQueryContainer(int $rowLimit) {
+  private function getQueryContainer(int $rowLimit, ?int $responseStreamMaxAge = NULL) {
```
paul-m, line 352: This method needs a docblock update, but not because of your changes. :-)

It needs two `@params` and also `@returns` a `MockObject` and not a `Chain`.

paul-m (changes requested):

modules/datastore/src/Controller/QueryDownloadController.php
zedmonds96: Added this to datastore.install and removed the constant.

```
/**
 * Set the default value for the response stream max age.
 */
function datastore_update_9002(&$sandbox) {
  $config = \Drupal::service('config.factory')->getEditable('datastore.settings');
  if ($config->get('response_stream_max_age') === NULL) {
    $config->set('response_stream_max_age', 3600)->save();
  }
}
```

modules/datastore/tests/src/Unit/Controller/QueryDownloadControllerTest.php
zedmonds96: Docblock now looks like this: 

```
/**
   * Create a mock object for the main container passed to the controller.
   *
   * @param int $rowLimit
   *    The row limit for a query.
   * @param int|null $responseStreamMaxAge
   *    The max age for the response stream in cache, or NULL to use the default.
   * @param array $info
   *   Dataset info array mock to be returned by DatasetInfo::gather().
   *
   * @return \PHPUnit\Framework\MockObject\MockObject
   *   MockChain mock object.
   */
```

paul-m: Thanks, but there's no `$info` parameter.

modules/datastore/src/Controller/AbstractQueryController.php
```diff
@@ -49,6 +49,7 @@ abstract class AbstractQueryController implements ContainerInjectionInterface {
    * @var int
    */
   protected const DEFAULT_ROWS_LIMIT = 500;
+  protected const DEFAULT_RESPONSE_STREAM_MAX_AGE = 3600;
```
paul-m: This is extraneous.

modules/datastore/src/Form/DatastoreSettingsForm.php
```diff
@@ -87,6 +95,7 @@ public function buildForm(array $form, FormStateInterface $form_state) {
   public function submitForm(array &$form, FormStateInterface $form_state) {
     $this->config('datastore.settings')
       ->set('rows_limit', $form_state->getValue('rows_limit') ?: QueryController::DEFAULT_ROWS_LIMIT)
+      ->set('response_stream_max_age', $form_state->getValue('response_stream_max_age') ?: QueryController::DEFAULT_RESPONSE_STREAM_MAX_AGE)
```
paul-m: Don't need the constant.

paul-m (changes requested):

modules/datastore/tests/src/Unit/Controller/QueryDownloadControllerTest.php
zedmonds96: That was there before, I didn't add it. Should I remove it?

modules/datastore/src/Controller/AbstractQueryController.php
zedmonds96: Removed.

modules/datastore/src/Form/DatastoreSettingsForm.php
zedmonds96: Removed.

modules/datastore/tests/src/Unit/Controller/QueryDownloadControllerTest.php
paul-m: Yes please.

zedmonds96: Removed. Thanks for the feedback.

paul-m (approved):

PR #4365, opened 2024-12-20
Add test coverage for Drupal 10.4

dafeder:
## Describe your changes

Drupal 10.4.0 is now out, we should add it to our test matrix.

## QA Steps

- I think just passing tests should be enough to consider this done?


paul-m (approved):
Looks like tests are passing with only the addition of 10.4 to the matrix.

PR #4367, opened 2024-12-20
Code quality review using Rector

dafeder:
New run of rector, updating config and removing some file and rule skips. Attempted to bump PHP target to 8.1 but created too much delta, will do that in another pass. 

Also made some manual fixes of things that jumped out at my local linter.

### DIFF
--- a/modules/datastore/src/DataDictionary/AlterTableQueryBase.php
+++ b/modules/datastore/src/DataDictionary/AlterTableQueryBase.php
@@ -14,43 +14,31 @@ abstract class AlterTableQueryBase implements AlterTableQueryInterface {
 
   /**
    * Keep track of whether this query has already been executed or not.
-   *
-   * @var bool
    */
   private bool $executed = FALSE;
 
   /**
    * Database connection.
-   *
-   * @var \Drupal\Core\Database\Connection
    */
   protected Connection $connection;
 
   /**
    * Strptime-to-MySQL date format converter.
-   *
-   * @var \PDLT\ConverterInterface
    */
   protected ConverterInterface $dateFormatConverter;
 
   /**
    * Query table.
-   *
-   * @var string
    */
   protected string $table;
 
   /**
    * Query fields.
-   *
-   * @var array
    */
   protected array $fields;
 
   /**
    * Query indexes.
-   *
-   * @var array
    */
   protected array $indexes;
 
--- a/modules/datastore/src/DataDictionary/AlterTableQueryBase.php
+++ b/modules/datastore/src/DataDictionary/AlterTableQueryBase.php
@@ -61,7 +61,7 @@ abstract class AlterTableQueryBase implements AlterTableQueryInterface {
     ConverterInterface $date_format_converter,
     string $table,
     array $fields,
-    array $indexes
+    array $indexes,
   ) {
     $this->connection = $connection;
     $this->dateFormatConverter = $date_format_converter;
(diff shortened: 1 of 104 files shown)


modules/datastore/src/Drush.php
```diff
@@ -186,13 +182,23 @@ public function list($options = [
    * Private.
    */
   private function createRow($uuid, $item) {
+    // Using deprecation helper.
     return [
       'uuid' => $uuid,
       'fileName' => $item->fileName,
       'fileFetcherStatus' => $item->fileFetcherStatus,
-      'fileFetcherBytes' => \format_size($item->fileFetcherBytes) . " ($item->fileFetcherPercentDone%)",
+      'fileFetcherBytes' => DeprecationHelper::backwardsCompatibleCall(
```
dafeder, line 190: A little on the fence about whether to use this Deprecation method, just jump to ByteSizeMarkup and break Drupal 10.1 (which is EOL), or skip the change completely for now.

modules/datastore/src/DataDictionary/AlterTableQueryBase.php
```diff
@@ -14,43 +14,31 @@ abstract class AlterTableQueryBase implements AlterTableQueryInterface {
...
    * Query table.
-   *
-   * @var string
    */
   protected string $table;
 
   /**
    * Query fields.
-   *
-   * @var array
```
paul-m, line 46: It's fine that we remove `@var` here, but we also don't have inline documentation of how to define fields or indexes. We should add that in the `__construct()` docblock, but that can be a follow-up.

modules/datastore/src/PostImportResult.php
```diff
@@ -43,29 +43,21 @@ class PostImportResult {
...
    * The PostImport service.
-   *
-   * @var \Drupal\datastore\Service\PostImport
    */
   protected PostImport $postImport;
 
   /**
    * The PostImportResult.
-   *
-   * @var array
```
paul-m, line 68: `$postImportResult` and `$resourceMapper` seem to be unused in this class. Let's do a followup to remove them.

modules/datastore/tests/src/Kernel/DatastoreServiceEventsTest.php
```diff
@@ -40,15 +40,13 @@ class DatastoreServiceEventsTest extends KernelTestBase implements EventSubscrib
 
   /**
    * Store the events we receive.
-   *
-   * @var array
```
paul-m, line 44: should be `@var \Drupal\datastore\Events\DatastoreEventBase[]`

modules/datastore/tests/src/Kernel/Service/ImportServiceEventsTest.php
```diff
@@ -37,15 +37,13 @@ class ImportServiceEventsTest extends KernelTestBase implements EventSubscriberI
 
   /**
    * Store the events we receive.
-   *
-   * @var array
```
paul-m, line 41: Should be `@var \Drupal\datastore\Events\DatastoreImportedEvent[]`

paul-m (changes requested):

modules/datastore/src/DataDictionary/AlterTableQueryBase.php
dafeder: OK will do

dafeder: See #4369

modules/datastore/src/PostImportResult.php
dafeder: Addressing this in #4370

paul-m (approved):

PR #4370, opened 2024-12-23
Refactor Post Import Service

kaise-lafrai:
### Run the following tests.

- [ ] `ddev dkan-phpunit --filter PostImportTest`
- [ ] `ddev dkan-phpunit --filter DictionaryEnforcerTest`
- [ ] `ddev dkan-phpunit --filter PostImportResultTest`
- [ ] `ddev dkan-phpunit --filter DashboardFormTest`

### Successful Post Import
- [ ] Set up vanilla site
- [ ] ddev drush en sample_content
- [ ] ddev drush dkan-sample-content:create
- [ ] Lets work with the 'Florida Bike Lanes' dataset otherwise known as node::8
- [ ] Create a data dictionary at/node/add/data?schema=data-dictionary
- [ ] Select Add field
- [ ] Title OBJECTID
- [ ] Name OBJECTID
- [ ] Title OBJECTID
- [ ] Data type Integer
- [ ] Description OBJECTID
- [ ] Grab the Identifier for later and select Save
- [ ] Navigate to DD settings : /admin/dkan/data-dictionary/settings
- [ ] Select Sitewide
- [ ] Add the identifier from previous step into Sitewide Dictionary ID
- [ ] Save configuration
- [ ] Run Cron a few times
- [ ] Confirm DD applied with no issues.

### Unsuccessful Post Import (Drop Resource)
- [ ] Navigate here and select Last Update and Save /admin/dkan/datastore
- [ ] Update the florida bike lane node last update date time and save
- [ ] Navigate here /admin/dkan/resources and select Drop the datastore table and save.
- [ ] Navigate to the data dictionary created before
- [ ] Alter the field "Data type" to something like "Date" and save
- [ ] Run cron a few times
- [ ] Navigate to dashboard /admin/dkan/datastore/status
- [ ] Confirm that the dataset cedcd327-4e5d-43f9-8eb1-c11850fa7c55 has a Store value of waiting(0%) with an error in the Post Import column.
- [ ] Confirm that the dataset was dropped by trying to access via the api and getting status of 400 /api/1/datastore/query/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/0

### Data Dictionary Disabled
- [ ] Navigate to data dictionary settings here admin/dkan/data-dictionary/settings and disable dictionary mode
- [ ] Update the florida bike lane node last update date time and save
- [ ] Run cron a few times
- [ ] Navigate to dashboard /admin/dkan/datastore/status
- [ ] Confirm that the dataset cedcd327-4e5d-43f9-8eb1-c11850fa7c55 has a Post Import status of N/A Data-Dictionary Disabled

### Data Dictionary via Distribution Reference
- [ ] Visit admin/dkan/data-dictionary/settings and set the data dictionary mode to "distribution reference."
- [ ] Post the following data dictionary to your DKAN instance:

```json
{
    "title": "Bike lanes data dictionary",
    "data": {
        "fields": [
            {
                "name": "objectid",
                "title": "OBJECTID",
                "type": "integer",
                "description": "Internal feature number."
            },
            {
                "name": "roadway",
                "title": "ROADWAY",
                "type": "string",
                "description": "A unique 8-character identification number assigned to a roadway or section of a roadway either On or Off the State Highway System for which information is maintained in the Department's Roadway Characteristics Inventory (RCI)."
            },
            {
                "name": "road_side",
                "title": "ROAD_SIDE",
                "type": "string",
                "constraints": {
                    "maxLength": 1,
                    "minLength": 1,
                    "enum": ["R", "L", "C"]
                },
                "description": "Side of the road. C = Composite; L = Left side; R = Right side"
            },
            {
                "name": "lncd",
                "title": "LNCD",
                "type": "integer",
                "constraints": {
                    "maxLength": 1,
                    "minLength": 1,
                    "maximum": 5,
                    "minimum": 0
                },
                "description": "Codes 0 = UNDESIGNATED; 1 = DESIGNATED; 2 = BUFFERED; 3 = COLORED; 4 = BOTH 2 AND 3; 5 = SHARROW"
            },
            {
                "name": "descr",
                "title": "DESCR",
                "type": "string",
                "constraints": {
                  "maxLength": 30,
                  "enum": ["UNDESIGNATED", "DESIGNATED"]
                },
                "description": "Designation description."
            },
            {
                "name": "begin_post",
                "title": "BEGIN_POST",
                "type": "number",
                "description": "Denotes the lowest milepoint for the record."
            },
            {
                "name": "end_post",
                "title": "END_POST",
                "type": "number",
                "description": "Denotes the highest milepoint for the record."
            },
            {
                "name": "shape_len",
                "title": "Shape_Leng",
                "type": "number",
                "description": "Length in meters"
            }
        ]
    }
}
```

- [ ] Post the following dataset to your DKAN instance.

```json
{
    "@type": "dcat:Dataset",
    "accessLevel": "public",
    "contactPoint": {
        "fn": "Jane Doe",
        "hasEmail": "mailto:data.admin@example.com"
    },
    "description": "Data on bike lanes in Florida.",
    "distribution": [
    {
        "@type": "dcat:Distribution",
        "downloadURL": "https://demo.getdkan.org/sites/default/files/distribution/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/Bike_Lane.csv",
        "mediaType": "text\/csv",
        "format": "csv",
        "title": "Florida Bike Lanes",
        "describedBy": "dkan://metastore/schemas/data-dictionary/items/b063524c-cf33-5d06-a161-770fc688d53b",
        "describedByType": "application/vnd.tableschema+json"
    }
    ],
    "identifier": "cedcd327-4e5d-43f9-8eb1-c11850fa7c55",
    "issued": "2016-06-22",
    "license": "http://opendatacommons.org/licenses/by/1.0/",
    "modified": "2016-06-22",
    "publisher": {
        "@type": "org:Organization",
    },
    "spatial": "Florida",
    "theme": [
        "Transportation",
        "City Planning"
    ],
    "title": "Florida Bike Lanes ",
    "keyword":["bike-lanes", "streets", "infrastructure"]
}
```
- [x] Run the import and post_import queues. The dictionary should be applied.

### Resource does not have dictionary
QA:

- Set the DKAN site to use 'Distribution reference' data dictionaries.
- Import datasets with no associated data dictionaries.
- Verify that no errors are thrown during the post_import phase, but instead an informative message is displayed.

Helpful script:
```
ddev dkan-site-install
ddev drush config:set metastore.settings data_dictionary_mode reference
ddev drush cron
ddev drush cron
```
You should see messages like this:
```
 [notice] No data-dictionary found for resource with id "0310f6c2d28bbee5de1211e5b5bb0b1c" and version "1721843317".
```
You should then also see a message like this in the dataset import status dashboard: 
```
done Resource 0310f6c2d28bbee5de1211e5b5bb0b1c does not have a data dictionary.
```


dafeder:
Going through the QA, I hit an issue at Unsuccessful Post Import (Drop Resource) > Confirm that the dataset cedcd327-4e5d-43f9-8eb1-c11850fa7c55 has a Store value of waiting(0%) with an error in the Post Import column.

When I ran cron I got the following error:

```sh
$ ddev drush cron
 [notice] Localization of resource 3a187a87dc6cd47c48b6b4c4785224b7: 
 [notice] ImportService for 3a187a87dc6cd47c48b6b4c4785224b7__1735074165 completed.
 [error]  SQLSTATE[22007]: Invalid datetime format: 1292 Incorrect date value: '2049' for column 'objectid' at row 1 
 [error]  Failed to drop datastore for aba79c7a62fec6c2e49bc1cebd0d7233. No datastore storage found for 3a187a87dc6cd47c48b6b4c4785224b7:1735068191. 
```

On the dashboard, I do see the right error in the postImport column but the store says done:

image

I can confirm the datastore was not dropped and the objectId has reverted back to text (it was int after the previous QA steps):

```http
GET https://kaise.ddev.site/api/1/datastore/query/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/0?results=false&count=false
```
returns

```json
{
  "schema": {
    "699890c9-480c-5f2e-89c5-06e9723303fb": {
      "fields": {
        "objectid": {
          "type": "text",
          "mysql_type": "text",
          "description": "﻿OBJECTID"
        },
        "roadway": {
          "type": "text",
          "mysql_type": "text",
          "description": "ROADWAY"
        },
        "road_side": {
          "type": "text",
          "mysql_type": "text",
          "description": "ROAD_SIDE"
        },
        "lncd": {
          "type": "text",
          "mysql_type": "text",
          "description": "LNCD"
        },
        "descr": {
          "type": "text",
          "mysql_type": "text",
          "description": "DESCR"
        },
        "begin_post": {
          "type": "text",
          "mysql_type": "text",
          "description": "BEGIN_POST"
        },
        "end_post": {
          "type": "text",
          "mysql_type": "text",
          "description": "END_POST"
        },
        "shape_leng": {
          "type": "text",
          "mysql_type": "text",
          "description": "Shape_Leng"
        }
      }
    }
  },
  "query": {
    "results": false,
    "count": false,
    "resources": [
      {
        "id": "699890c9-480c-5f2e-89c5-06e9723303fb",
        "alias": "t"
      }
    ],
    "limit": 500,
    "offset": 0,
    "schema": true,
    "keys": true,
    "format": "json",
    "rowIds": false,
    "properties": [
      "objectid",
      "roadway",
      "road_side",
      "lncd",
      "descr",
      "begin_post",
      "end_post",
      "shape_leng"
    ]
  }
}
```

modules/datastore/src/Service/PostImport.php
```diff
@@ -116,4 +328,30 @@ public function removeJobStatus($resourceIdentifier): bool {
...
+    }
+  }
+
+  /**
+   * Invalidate all appropriate cache tags for this resource.
+   *
+   * @param mixed $resourceId
+   *   A resource ID.
+   */
+  public function invalidateCacheTags($resourceId): void {
```
dafeder: This is a redundant method. It's identical to DatastoreService::invalidateCacheTags, and we are already injecting DatastoreService. Remove this method, remove ReferenceLookup as a dependency, and call the function as $this->datastoreService->invalidateReferencerCacheTags().

dafeder: (need to make the DatastoreService method public, obviously)

modules/datastore/src/Service/PostImport.php
```diff
@@ -2,40 +2,252 @@
...
+    DataDictionaryDiscoveryInterface $dataDictionaryDiscovery,
+    ReferenceLookup $referenceLookup,
+    DatastoreService $datastoreService,
     Connection $connection,
-    ResourceMapper $resource_mapper
-    ) {
+  ) {
+    $this->configFactory = $configFactory;
+    $this->logger = $logger;
+    $this->resourceMapper = $resourceMapper;
```
dafeder: Trying to think about how to get the number of dependencies down here. ResourceMapper is also part of DatastoreService. I think it would be safe to just make $datastoreService->resourceMapper public, since there are no public properties or setter methods on resourceMapper. Or we could add a getResourceMapper() method to $datastoreService.

modules/datastore/src/Service/PostImport.php
```diff
@@ -2,40 +2,252 @@
...
+    $this->resourceProcessorCollector = $resourceProcessorCollector;
+    $this->dataDictionaryDiscovery = $dataDictionaryDiscovery;
+    $this->referenceLookup = $referenceLookup;
```
dafeder: See commnent below for a suggestion how we can eliminate the referenceLookup dependency.

modules/datastore/src/Plugin/QueueWorker/PostImportResourceProcessor.php
```diff
@@ -32,108 +22,44 @@
...
-    $this->datastoreService = $datastoreService;
     $this->postImport = $post_import;
-    $this->dataDictionaryDiscovery = $data_dictionary_discovery;
-    // Set the timeout for database connections to the queue lease time.
-    // This ensures that database connections will remain open for the
-    // duration of the time the queue is being processed.
-    $timeout = (int) $plugin_definition['cron']['lease_time'];
-    $alter_table_query_builder->setConnectionTimeout($timeout);
-    $this->referenceLookup = $referenceLookup;
+    $this->configFactory = $config_factory;
```
dafeder: I could be wrong but I don't think this is used for anything?

dafeder (changes requested):
Looking good! See the question on QA in the PR conversation.

Also, reviewing how the code works there is something off to me in the back and forth between `PostImport` and `PostImportResult`. It's just strange to me that `PostImport` puts this data into an object, which has a `storeResult()` method, but the `storeResult()` method just calls another method that's back in `PostImport()`. It's essentially a circular dependency, though Drupal won't notice because `PostImportResult` is not a service. I think the correct way to do this would be one of two things:

1. Make the entire storeResult() method on PostImportResult be essentially self-contained, so it contains the DB query and just needs to be passed a Connection object (ie the `database` service). Remove PostImport::storeJobStatus(). This would also remove Connection as a dependency for PostImport.
2. Remove the `storeResult()` method from `PostImportResult`, and simply have `PostImport::storeJobStatus` take a `PostImportResult` object as its argument, rather than all those individual arguments. This means that rather than rather than calling `$result->storeResult()`, `PostImportResourceProcessor::processItem()` would call `$this->postImport->storeJobStatus($result)`.

I'd like to see this changed but if it feels out of scope we can make a new ticket for it, tell me what you think.

kaise-lafrai:
> Going through the QA, I hit an issue at Unsuccessful Post Import (Drop Resource) > Confirm that the dataset cedcd327-4e5d-43f9-8eb1-c11850fa7c55 has a Store value of waiting(0%) with an error in the Post Import column.
> 
> When I ran cron I got the following error:
> 
> ```shell
> $ ddev drush cron
>  [notice] Localization of resource 3a187a87dc6cd47c48b6b4c4785224b7: 
>  [notice] ImportService for 3a187a87dc6cd47c48b6b4c4785224b7__1735074165 completed.
>  [error]  SQLSTATE[22007]: Invalid datetime format: 1292 Incorrect date value: '2049' for column 'objectid' at row 1 
>  [error]  Failed to drop datastore for aba79c7a62fec6c2e49bc1cebd0d7233. No datastore storage found for 3a187a87dc6cd47c48b6b4c4785224b7:1735068191. 
> ```
> 
> On the dashboard, I do see the right error in the postImport column but the store says done:
> 
> image
> 
> I can confirm the datastore was not dropped and the objectId has reverted back to text (it was int after the previous QA steps):
> 
> ```httpspec
> GET https://kaise.ddev.site/api/1/datastore/query/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/0?results=false&count=false
> ```
> 
> returns
> 
> ```json
> {
>   "schema": {
>     "699890c9-480c-5f2e-89c5-06e9723303fb": {
>       "fields": {
>         "objectid": {
>           "type": "text",
>           "mysql_type": "text",
>           "description": "﻿OBJECTID"
>         },
>         "roadway": {
>           "type": "text",
>           "mysql_type": "text",
>           "description": "ROADWAY"
>         },
>         "road_side": {
>           "type": "text",
>           "mysql_type": "text",
>           "description": "ROAD_SIDE"
>         },
>         "lncd": {
>           "type": "text",
>           "mysql_type": "text",
>           "description": "LNCD"
>         },
>         "descr": {
>           "type": "text",
>           "mysql_type": "text",
>           "description": "DESCR"
>         },
>         "begin_post": {
>           "type": "text",
>           "mysql_type": "text",
>           "description": "BEGIN_POST"
>         },
>         "end_post": {
>           "type": "text",
>           "mysql_type": "text",
>           "description": "END_POST"
>         },
>         "shape_leng": {
>           "type": "text",
>           "mysql_type": "text",
>           "description": "Shape_Leng"
>         }
>       }
>     }
>   },
>   "query": {
>     "results": false,
>     "count": false,
>     "resources": [
>       {
>         "id": "699890c9-480c-5f2e-89c5-06e9723303fb",
>         "alias": "t"
>       }
>     ],
>     "limit": 500,
>     "offset": 0,
>     "schema": true,
>     "keys": true,
>     "format": "json",
>     "rowIds": false,
>     "properties": [
>       "objectid",
>       "roadway",
>       "road_side",
>       "lncd",
>       "descr",
>       "begin_post",
>       "end_post",
>       "shape_leng"
>     ]
>   }
> }
> ```

kaise-lafrai:
@dafeder Can you confirm you selected **Drop the datastore table if the post import queue reports an error.** on the Resource settings page (admin/dkan/resources) before attempting to run a failing post import?

I attempted the QA steps and was able to see "waiting (0%)" under the store column.

kaise-lafrai:
@dafeder I'm attempting to refactor so that I am satisfying this approach.

_Make the entire storeResult() method on PostImportResult be essentially self-contained, so it contains the DB query and just needs to be passed a Connection object (ie the database service). Remove PostImport::storeJobStatus(). This would also remove Connection as a dependency for PostImport._

I would not be able to remove connection as a dependecy for PostImport because I would need to pass that in when instantiating PostImportResult. I will also need to add the connection service in the DashboardForm since I moved over all the db related code out of the PostImport class into the PostImportResult class.

modules/datastore/src/Plugin/QueueWorker/PostImportResourceProcessor.php
kaise-lafrai: You are correct, it was previously used but is now used in the postImport class. Removing.

modules/datastore/src/Service/PostImport.php
kaise-lafrai: I don't see your suggestion.

dafeder:
> @dafeder Can you confirm you selected **Drop the datastore table if the post import queue reports an error.** on the Resource settings page (admin/dkan/resources) before attempting to run a failing post import?
> 
> I attempted the QA steps and was able to see "waiting (0%)" under the store column. 

I think so but possible I missed a step. I'll try again on Monday.

dafeder:
> @dafeder I'm attempting to refactor so that I am satisfying this approach.
> 
> _Make the entire storeResult() method on PostImportResult be essentially self-contained, so it contains the DB query and just needs to be passed a Connection object (ie the database service). Remove PostImport::storeJobStatus(). This would also remove Connection as a dependency for PostImport._
> 
> I would not be able to remove connection as a dependecy for PostImport because I would need to pass that in when instantiating PostImportResult. I will also need to add the connection service in the DashboardForm since I moved over all the db related code out of the PostImport class into the PostImportResult class.

You're right, apologies for not thinking about it more carefully. I had in my head there was a way to get the connection directly from DatastoreService but it looks like there is not. Disregard.

kaise-lafrai:
> @dafeder I'm attempting to refactor so that I am satisfying this approach.
> 
> _Make the entire storeResult() method on PostImportResult be essentially self-contained, so it contains the DB query and just needs to be passed a Connection object (ie the database service). Remove PostImport::storeJobStatus(). This would also remove Connection as a dependency for PostImport._
> 
> I would not be able to remove connection as a dependecy for PostImport because I would need to pass that in when instantiating PostImportResult. I will also need to add the connection service in the DashboardForm since I moved over all the db related code out of the PostImport class into the PostImportResult class.

I was able to use the database object like this $connection = \Drupal::database(); in the PostImportResult class. This way I can avoid injecting it into in the PostImport and DashboardForm classes. Pros and Cons doing it this way but we do satisfy code climate by reducing the number of arguments in the constructor.

dafeder:
Let's stick with the dependency injection pattern, the argument limit from CC is a good guideine but not an absolute rule.

kaise-lafrai:
@janette @dafeder This is ready for review/QA.

dafeder:
QA all passes

modules/datastore/src/PostImportResult.php
```diff
@@ -51,37 +50,86 @@ class PostImportResult {
...
   protected array $postImportResult;
 
   /**
-   * PostImportResource constructor.
+   * PostImportResult constructor.
    */
   public function __construct(
     $postImportResult,
+    Connection $connection,
     ResourceMapper $resourceMapper,
```
dafeder: After some discussion we don't believe the resourceMapper is necessary here. Confirm, and then remove this plus the code that uses it to find the latest version ID.

modules/datastore/src/PostImportResult.php
```diff
@@ -51,37 +50,86 @@ class PostImportResult {
...
    * The PostImportResult.
    */
   protected array $postImportResult;
 
   /**
-   * PostImportResource constructor.
+   * PostImportResult constructor.
    */
   public function __construct(
     $postImportResult,
```
dafeder: Let's try to pass as arguments a DataResource object and two strings, rather than an arbitrary associative array.

modules/datastore/src/PostImportResultFactory.php
```diff
@@ -0,0 +1,82 @@
...
+   *   Status of the post import process.
+   * @param string $message
+   *   Error messages retrieved during the post import process.
+   * @param \Drupal\common\DataResource $resource
+   *   The DKAN resource being imported.
+   *
+   * @return \Drupal\datastore\PostImportResult
+   *   The post import result service.
+   */
+  public function createPostImportResult($status, $message, DataResource $resource): PostImportResult {
```
dafeder: Improve the names of the two create methods to make it clearer what each is for

modules/datastore/src/PostImportResultFactory.php
```diff
@@ -0,0 +1,82 @@
...
+    return new PostImportResult([
+      'resource_id' => $resource->getIdentifier(),
+      'resource_version' => $resource->getVersion(),
+      'postImportStatus' => $status,
+      'postImportMessage' => $message,
+    ],
+    $this->connection,
```
dafeder: Pass a Resource object and two strings rather than an array.

modules/datastore/src/PostImportResult.php
```diff
@@ -51,37 +50,86 @@ class PostImportResult {
...
    */
-  public function storeResult() {
-    return $this->postImport->storeJobStatus($this->resourceIdentifier, $this->resourceVersion, $this->postImportStatus, $this->postImportMessage);
+  public function removeJobStatus(): bool {
+    try {
+      $latest_resource = $this->resourceMapper->get($this->getResourceIdentifier());
+      $latest_version = $latest_resource->getVersion();
+      $this->connection->delete('dkan_post_import_job_status')
+        ->condition('resource_identifier', $this->getResourceIdentifier(), '=')
+        ->condition('resource_version', $latest_version, '=')
```
dafeder: Should not be necessary to override the version passed to the constructor with the latest version; test and confirm then remove.

dafeder (changes requested):

modules/datastore/src/PostImportResultFactory.php
```diff
@@ -0,0 +1,86 @@
...
+   *
+   * Passes the dataset info when initialized.
+   *
+   * @param array $datasetInfo
+   *   The dataset info.
+   *
+   * @return \Drupal\datastore\PostImportResult
+   *   The PostImportResult object.
+   */
+  public function initializeFromDatasetInfo(array $datasetInfo): PostImportResult {
```
dafeder: I think this and the docbloc are a little misleading because you don't actually take the full DatasetInfo array, you're looking for a specific Distribution array, right?

modules/datastore/src/PostImportResultFactory.php
```diff
@@ -0,0 +1,86 @@
...
+    $resource = $this->resourceMapper->get($datasetInfo['resource_id']);
+    return new PostImportResult(
+      '',
+      '',
```
dafeder:
```suggestion
      NULL,
      NULL,
```
I'm a little unclear what it means to create a status object with NULL status - shouldn't there be a default? Anyway since the constructor allows NULL, NULL seems more correct here as argument.

dafeder (changes requested):
One or two more things here and I think we're good to go

modules/datastore/src/PostImportResultFactory.php
kaise-lafrai: This method for initialization is used to retrieve results from the `dkan_post_import_job_status` table. Hence why we don't need to initialize with the status or message.

dafeder (approved):

kaise-lafrai:
@janette If you have time could you please review this PR?

janette (approved):
Tested 👍

PR #4371, opened 2024-12-23
Provide text_format render element in JSON Form Widget

### LINKED ISSUE #4132: Provide text_format render element for string schema properties

dafeder:
Fixes #4132 
Re-roll of #4133 by @stefan-korn 

This provides an option in the UI Schema to define a Drupal text format for a textarea property. It only affects JSON Form widget forms, not any rendering or API response. We may do a follow-up to reflect this in the twig template or accompanying theme preprocess function. For now, what this achieves is replacing the bare textarea element on a form with a text_format element.

It does not allow a selection of all text formats available in Drupal, but locks a particular property in to a particular text format (as this is only about presentation and there is nowhere to store entity-specific format information).

I also added some test coverage to the `mailto:` fix as we kept getting dinged on that in coverage reports (even though that code needs to be replaced at some point).

## QA Steps

- [ ] Add a basic HTML text format to Drupal (/admin/config/content/formats)
- [ ] Edit your dataset.ui.json so that the description property includes `"textFormat": "html"
- [ ] Add a new dataset. Confirm that you get the help text under the textarea that you would expect for the selected text format
- [ ] Now, enable the ckeditor module.
- [ ] Add another text format that uses ckeditor. Call it "ckeditor."
- [ ] Change the "textFormat" property in dataset.ui.json description to "ckeditor"
- [ ] Confirm that the dataset form now displays ckeditor for the description field

### DIFF
--- a/modules/json_form_widget/README.md
+++ b/modules/json_form_widget/README.md
@@ -154,7 +154,7 @@ UI options:
         "widget": "textarea",
         "rows": 5,
         "description": "Description (e.g., an abstract) with sufficient detail to enable a user to quickly understand whether the asset is of interest.",
-        "textFormat": "html
+        "textFormat": "html"
       }
     },
 
--- a/modules/json_form_widget/README.md
+++ b/modules/json_form_widget/README.md
@@ -162,7 +162,9 @@ UI options:
 - textFormat: html
   - This can be the machine name of any Drupal text format you have configured in your system
 
-Note that using the `textFormat` property will not affect how the field is rendered; you must handle this in your decoupled frontend or by overriding the node--data twig template or `metastore_preprocess_node__data()` method in metastore.theme.inc. See the [Theming Drupal](https://www.drupal.org/docs/develop/theming-drupal) guide for more information.
+You can use `textFormat` to add a WYSIWYG or other Drupal text editor plugin to your dataset form. Just enable the modules you need and assign the editor to your text format in Drupal at /admin/config/content/formats.
+
+ Note that using the `textFormat` property will not affect how the field is rendered; you must handle this in your decoupled frontend or by overriding the node--data twig template or `metastore_preprocess_node__data()` method in metastore.theme.inc. See the [Theming Drupal](https://www.drupal.org/docs/develop/theming-drupal) guide for more information.
 
 **Form Element:**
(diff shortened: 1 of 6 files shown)


modules/json_form_widget/README.md
```diff
@@ -136,6 +136,36 @@ UI options:
...
+    },
+
+**UI Schema File Example:**
+
+    "description": {
+      "ui:options": {
+        "widget": "textarea",
+        "rows": 5,
+        "description": "Description (e.g., an abstract) with sufficient detail to enable a user to quickly understand whether the asset is of interest.",
+        "textFormat": "html
```
janette: Add closing quote mark.

janette (changes requested):
We held off as long as we could :)
Just the closing quote then this is good to go.

modules/json_form_widget/README.md
dafeder: Oops - done!

janette (approved):

PR #4372, opened 2024-12-27
Factor out DatastoreResource class

### LINKED ISSUE #4359: Consolidate DataResource and DatastoreResource
In order to eventually sort our our circular dependency issues and get the DKAN modules organized in a more intuitive way (not to mention making the code more comprehensible) we need to collapse these two very redundant classes/objects into a single thing. The `DatastoreResource` class has very little value; it does not reference any other classes or services from the Datastore module and just adds a few more properties that are just built from the DataResource properties.

This will take a bit of trial and error to get right, but will consist of, generally

1. Figuring out how to reproduce any additional fuctionality from `DatastoreResource` in the `DataResource` class. This may involve some util-type functions that stay in Datastore and process info retrievevd from DataResource`.
2. Changing all uses of DatastoreResource throughout the Datastore module to use `DataResource` instead.  
3. Deprecating DatastoreResource but keeping for now as a shim to prevent any custom or contrib code using it from breaking.

dafeder:
Fixes #4359

## Describe your changes

Remove DatastoreResource class, use DataResource for everything. Note that any class or function that expected a DatastoreResource object now expects a DataResource object. So any custom code that, for instance, builds a datastore `DatabaseTable` object will need to be slightly refactored. 

Other changes to note:

* `AbstractDatabaseTable::setTable()` is now public. We should use this directly instead of calling `::count()` (as we do currently in some places) to indirectly create a database table.
* `DictionaryEnforcer` now requires the `dkan.datastore.database_table_factory` service

Please override the CC warnings when reviewing, they are due to CC getting confused by the ReturnTypeMayChange attribute. May be an issue with CC config.

## QA Steps

Nothing to QA, everything should still work as it did before from user's perspective.

### DIFF
--- a/modules/common/tests/src/Unit/DataResourceTest.php
+++ b/modules/common/tests/src/Unit/DataResourceTest.php
@@ -19,6 +19,21 @@ use PHPUnit\Framework\TestCase;
 class DataResourceTest extends TestCase {
 
   /**
+   * Test getTableName().
+   */
+  public function testGetTableName() {
+
+    $resource = new DataResource(
+      '/foo/bar',
+      'txt',
+      DataResource::DEFAULT_SOURCE_PERSPECTIVE
+    );
+    $tableName = $resource->getTableName();
+
+    $this->assertStringStartsWith('datastore_', $tableName);
+  }
+
+  /**
    * Test getFolder().
    */
   public function testGetFolder() {
--- a/modules/common/tests/src/Unit/DataResourceTest.php
+++ b/modules/common/tests/src/Unit/DataResourceTest.php
@@ -19,21 +19,6 @@ use PHPUnit\Framework\TestCase;
 class DataResourceTest extends TestCase {
 
   /**
-   * Test getTableName().
-   */
-  public function testGetTableName() {
-
-    $resource = new DataResource(
-      '/foo/bar',
-      'txt',
-      DataResource::DEFAULT_SOURCE_PERSPECTIVE
-    );
-    $tableName = $resource->getTableName();
-
-    $this->assertStringStartsWith('datastore_', $tableName);
-  }
-
-  /**
    * Test getFolder().
    */
   public function testGetFolder() {
--- a/modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
+++ b/modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
@@ -68,8 +68,12 @@ class MysqlImport extends ImportJob {
     }
 
     // Attempt to resolve resource file name from file path.
-    if (($file_path = \Drupal::service('file_system')->realpath($this->resource->getFilePath())) === FALSE) {
-      return $this->setResultError(sprintf('Unable to resolve file name "%s" for resource with identifier "%s".', $this->resource->getFilePath(), $this->resource->getId()));
+    if (($file_path = \Drupal::service('file_system')->realpath($this->resource->getFilePath(TRUE))) === FALSE) {
+      return $this->setResultError(sprintf(
+        'Unable to resolve file name "%s" for resource with identifier "%s".',
+        $this->resource->getFilePath(TRUE),
+        $this->resource->getUniqueIdentifier())
+      );
     }
 
     $size = @filesize($file_path);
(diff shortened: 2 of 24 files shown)


dafeder:
Currently debugging failing MySqlQueryDownloadControllerTest, which I think just happens to be the first test to run that tries to run a query. It fails at `AbstractDatabaseTable::setTable()`, I think something is making the perspective be `local_file` when it should be something else? It can't get a schema anyway.

dafeder:
Some progress. Need to redo QueryDownloadControllerTest somehow... not so easy to use fake resource names because table name is no longer based on an ID in the DatastoreResource class that we can simply control via the constructor.

modules/json_form_widget/tests/src/Unit/StringHelperTest.php
```diff
@@ -102,7 +102,6 @@ public function testHandleStringElement() {
     $data = 'mailto:john@doe.com';
 
     $result = $string_helper->handleStringElement($definition, $data);
-    print_r($result['#default_value']);
```
dafeder, line 105: Not sure how this crept in here!

paul-m (changes requested):
`Drupal\Tests\datastore\Unit\Storage\DatabaseTableFactoryTest` still creates a `DatastoreResource`.

paul-m (changes requested):
Let's add to the docblock of `Drupal\datastore\Storage\DatabaseTableFactory::getInstance()` that the resource must be a DataResource object.

paul-m (changes requested):
`Drupal\Tests\datastore\Unit\Plugin\QueueWorker\ImportJobTest` spends a lot of time making `DatastoreResource` objects.

`Drupal\Tests\datastore\Unit\Service\Info\ImportInfoTest` too.

Basically search the codebase for `DatastoreResource` and you'll see a bunch of tests.

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
```diff
@@ -68,8 +68,12 @@ protected function runIt() {
     }
 
     // Attempt to resolve resource file name from file path.
-    if (($file_path = \Drupal::service('file_system')->realpath($this->resource->getFilePath())) === FALSE) {
-      return $this->setResultError(sprintf('Unable to resolve file name "%s" for resource with identifier "%s".', $this->resource->getFilePath(), $this->resource->getId()));
+    if (($file_path = \Drupal::service('file_system')->realpath($this->resource->getFilePath(TRUE))) === FALSE) {
+      return $this->setResultError(sprintf(
```
paul-m, line 72: Needs test coverage, since we might be inadvertently modifying the logic.

modules/common/tests/src/Unit/DataResourceTest.php
```diff
@@ -18,21 +18,6 @@
  */
 class DataResourceTest extends TestCase {
 
-  /**
```
paul-m: We should leave this test because even though it covers deprecated code, the code is still there.

paul-m (changes requested):

modules/common/tests/src/Unit/DataResourceTest.php
dafeder: I was having trouble getting the test to work and thought it was maybe not worth it but I'll give it another shot

modules/datastore/modules/datastore_mysql_import/src/Service/MysqlImport.php
dafeder: After some discussion, this seems outside the scope of this PR. We will do a follow-up to cover this line, and maybe make it more consistent across the two ImportJob classes.

modules/common/tests/src/Unit/DataResourceTest.php
dafeder: Actually it just works as is, I'm not sure why this was removed tbh! Restored.

dafeder:
> Let's add to the docblock of `Drupal\datastore\Storage\DatabaseTableFactory::getInstance()` that the resource must be a DataResource object.

@paul-m see 2594e68

paul-m (approved):

dafeder:
Putting back to draft, needs a bit more testing from @janette

dafeder:
Update from @janette: we're good to go

janette (approved):

Issue #3637, opened 2021-08-25
Filters and sorts for datastore import status dashboard


dafeder:
Usually you are looking for a specific dataset when you come to the dashboard. Add:

1. A title filter (maybe fulltext content?)
2. User filter
3. UUID filter
4. Sort by last imported, title, user

dafeder:
I think we decided #3714, which adds a UUID filter, was enough for now.

dafeder:
Re-opening, we do want to add this if possible. The way we store some of this info as JSON and not columnar data makes some of this hard. Title will probably be the easiest.

paul-m:
Based on spike 19023, we split this out into the easy bits and the hard bits.

Low effort:
- Title filter 23788
- User filter 23789

High effort:
- Any table sort

dafeder:
FWIW I think this and #4332 suggest we really need a new approach to the dashboard. I think somehow it needs to be based on an array of resources that the metadata is derived from, not a list of datasets that we then inspect for resource/import info. Or at least to find some way of capturing a bit more in DB columns so that we are not dependent on data trapped inside JSON. Not saying we can't make some of these improvements on top of the existing approach but we we may be kind of at the limits of what we can do here.

paul-m:
The user and title filters aren't that hard to implement as things are now, which is why I separated them out.

In an ideal world this would be a view in Views and we could mix and match as desired. Sorts would then be a freebie if the data were represented by a views handler.

PR #4376, opened 2025-01-07
Filter by title on the dataset import status dashboard

paul-m:
Partially fixes [#3637]
JIRA: 23788

## Describe your changes

This PR adds a new filter to search the Datastore Import Status dashboard with the dataset title.

## QA Steps

- [ ] Checkout branch locally.
- [ ] run `ddev dkan-site-install`
- [ ] run `ddev dkan-sample-content`
- [ ] run `ddev drush cron` 2X
- [ ] run `ddev drush uli`
- [ ] navigate to the dashboard here: /admin/dkan/datastore/status
- [ ] Confirm you are able to search with Dataset ID, Dataset Title and Harvest ID.

### DIFF
--- a/modules/datastore/src/Form/DashboardForm.php
+++ b/modules/datastore/src/Form/DashboardForm.php
@@ -2,15 +2,18 @@
 
 namespace Drupal\datastore\Form;
 
-use Drupal\common\DataResource;
-use Drupal\Core\Pager\PagerManagerInterface;
-use Drupal\Core\StringTranslation\StringTranslationTrait;
+use Drupal\Core\Datetime\DateFormatter;
+use Drupal\Core\Entity\EntityStorageInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Form\FormBase;
 use Drupal\Core\Form\FormStateInterface;
-use Drupal\common\DatasetInfo;
-use Drupal\Core\Datetime\DateFormatter;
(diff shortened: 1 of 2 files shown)


dafeder:
This could really use an integration test, but seems to be out of scope for this ticket since there wasn't one previously.

dafeder:
QA passes, just want to note that the UX is a little weird here. There are multiple filters, but it only pays attention to one, and it's not transparent which one is going to be used. I'm approving cause this is only admin-facing and I know we want to get this in, but if it proves to be too confusing we might want to revisit.

dafeder (approved):
See PR comments

kaise-lafrai:
> QA passes, just want to note that the UX is a little weird here. There are multiple filters, but it only pays attention to one, and it's not transparent which one is going to be used. I'm approving cause this is only admin-facing and I know we want to get this in, but if it proves to be too confusing we might want to revisit.

Yes this can be confusing. To clarify the search goes from left to right, starting with Dataset ID, then Title, and then Harvest ID and it narrows down the results further and further. Correct me if I am wrong @paul-m

paul-m:
Right, that's how I had made it initially, just because that's how it worked choosing between UUID and harvest, IIRC.

I think being able to filter by title is a good feature to have at this point, even if it's a little unclear. If it's too ambiguous we can iterate again later.

janette (approved):
like it 👍

PR #4379, opened 2025-01-09
Set 404 exception if dataset ID isn't found.

beeyayjay:
Fixes an longstanding issue in the dkan_js_frontend module. Paths with /dataset/[id] returned a 202 code even when the dataset was missing. The headers have already been set so the FE cannot change the response code. This PR tries to retrieve the dataset before sending the response, so that it can send a 404 code but still allow the decoupled FE app to control the content display of the page.

If you are not using the dkan_js_fronend module this will not affect any other DKAN behavior.

WECMS-6711

## QA Steps

- [ ] Open the developer tools and view the network tab
- [ ] Go to  /dataset/{id} where id is not an existing identifier for example /dataset/test123
- [ ] Verify that the response status code is 404


modules/dkan_js_frontend/tests/src/Functional/Controller/DatasetControllerTest.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace modules\dkan_js_frontend\tests\Functional\Controller;
+
+use Drupal\Core\Url;
+use Drupal\Tests\BrowserTestBase;
+use GuzzleHttp\RequestOptions;
+
+class DatasetControllerTest extends BrowserTestBase {
```
dafeder: This isn't actually testing the dkan_js_frontend controller, because the module is not enabled for this test. It's getting a 404 because it's not a valid URL either way, but it's just coming from normal Drupal routing and node controller. That's why it's getting flagged for test coverage.

dafeder (changes requested):

modules/dkan_js_frontend/tests/src/Functional/Controller/DatasetControllerTest.php
beeyayjay: Thanks @dafeder. That hadn't occurred to me. I'll fix.

modules/dkan_js_frontend/src/Controller/Page.php
```diff
@@ -3,19 +3,94 @@
...
     ];
   }
 
+  /**
+   * If a dataset with an invalid ID is being requested, throw a 404 error.
+   */
+  protected function handleInvalidDatasetId() {
+    // Path should always have leading slash.
+    // @see \Symfony\Component\HttpFoundation\Request::getPathInfo()
+    // Dataset path is /dataset/[ID]/data.
```
janette: /dataset/[ID]/data is a custom project route that does not exist in core, can we change the match to a wildcard for anything after the dataset uuid?

janette: This works for the /data route:

```
$dataset_path_match = '/^\/dataset\/(?P<id>[^\/]+)*/';

    $path = $this->currentPath->getPath();

    if (preg_match($dataset_path_match, $path, $matches)) {
      try {
        $this->metastoreService->get('dataset', $matches['id']);
      }
      catch (MissingObjectException $exception) {
        throw new NotFoundHttpException();
      }
    }
```

janette (approved):
Thanks @beeyayjay looks good

janette:
@dafeder tested and approved ✅

modules/dkan_js_frontend/dkan_js_frontend.services.yml
```diff
@@ -3,3 +3,11 @@ services:
       class: Drupal\dkan_js_frontend\Routing\RouteProvider
       arguments:
         - '@config.factory'
+  dkan_js_frontend.controller:
```
dafeder: It's very unusual to define a controller as a service - is this necessary?

modules/dkan_js_frontend/tests/src/Functional/Controller/PageTest.php
```diff
@@ -0,0 +1,41 @@
...
+   */
+  public function test() {
+    $this->drupalGet('home');
+    $session = $this->getSession();
+    $this->assertEquals(200, $session->getStatusCode(), $session->getPage()->getHtml());
+
+    $this->drupalGet('dataset/123');
+    $session = $this->getSession();
+    $this->assertEquals(404, $session->getStatusCode(), $session->getPage()->getHtml());
+  }
```
dafeder: Does this test anything that's not tested in the kernel test? Or should I say, is there any way that this test could fail if the Kernel test is passing? If not, I don't think it's worth adding another functional test for this, as they're very expensive.

dafeder (changes requested):
This looks good @beeyayjay - see my two questions. I also have a slight worry that access logic for logged-in users might be more complex than this allows for, but if that's not an issue for any of our current sites using this module I think we can address through documentation.

modules/dkan_js_frontend/dkan_js_frontend.services.yml
beeyayjay: No it's not. Added a commit to remove it from services yaml.

modules/dkan_js_frontend/tests/src/Functional/Controller/PageTest.php
beeyayjay: I think you're right. Removed functional test.

dafeder (approved):

PR #4381, opened 2025-01-14
Add config option in metastore settings to stop forcing redirect

### LINKED ISSUE #4326: Make redirect to dataset view optional
### Feature Description

Add a config option to metastore settings to stop forcing a redirect to the datasets view on dataset form submit.

### Problem and Motivation

We have a [redirect in metastore_admin.module](https://github.com/GetDKAN/dkan/blob/30b89232e7ad49f0cc9383e9f88e3fa4f5dbb639/modules/metastore/modules/metastore_admin/metastore_admin.module#L148) that takes people back to the dataset view after saving a dataset form. This is quite difficult to disable/override, and in many cases may not be the desired behavior. For instance if using a draft workflow, the dataset author may not have permission to see that page at all.

### Potential Benefits

More flexibility in user experience for different use cases.

### Target Audience

- [X] End Users
- [X] Developers
- [X] Administrators
- [ ] Other

### Possible Implementation

Add a config option and a checkbox in one of the admin forms, check for this before adding the redirect in the submit function.

zedmonds96:
Fixes #4326 

## Describe your changes
Added a configuration option in metastore settings. If checked, the user will be redirected to the datasets view when creating/editing a dataset. If unchecked, the user will remain on the node page for the dataset instead of the redirect. 

- Added configuration key `redirect_to_dataset` to `metastore.settings.yml`. 
- Added configuration schema for `redirect_to_dataset` to `metastore.schema.yml`.
- Added `redirect_to_datasets` checkbox to the metastore settings form in `DkanDataSettingsForm.php`
- Restructured the metastore settings form for future maintainability. 
- Modified `metastore_admin_form_submit()` in `metastore_admin.module` to get the checkbox value on submission to determine redirect behavior. 
- Added update hook `metastore_update_8010()` to `metastore.install` to set the default value of the config. 
- Created functional tests for when the configuration is enabled and disabled in `RedirectToDatasetsTest.php`

## QA Steps

- [ ] Navigate to DKAN -> Metastore Settings.
- [ ] Confirm that the "Redirect to datasets view after form submit" checkbox exists.
- [ ] Verify that it is checked by default.
- [ ] Go to Content -> Add Content -> Create Dataset.
- [ ] Fill in the required fields and submit the form.
- [ ] Expected result: You should be redirected to the datasets view after submission.
- [ ] Return to Metastore Settings.
- [ ] Verify the checkbox is still checked, then uncheck it and save.
- [ ] Edit the dataset you just created, make a small change, and submit again.
- [ ] Expected result: You should stay on the dataset’s node page after submission instead of being redirected.

- [N/A] I have updated or added documentation

### DIFF
--- a/modules/metastore/modules/metastore_admin/tests/src/Functional/RedirectToDatasetsTest.php
+++ b/modules/metastore/modules/metastore_admin/tests/src/Functional/RedirectToDatasetsTest.php
@@ -99,25 +99,19 @@ class RedirectToDatasetsTest extends BrowserTestBase {
       ->set('redirect_to_datasets', FALSE)
       ->save();
 
-    $this->drupalGet('node/add/data');
+    $this->drupalGet('admin/content');
     $assert->statusCodeEquals(200);
+    $this->clickLink($dataset_title);
+    $this->drupalGet($this->getSession()->getCurrentUrl() . '/edit');
 
-    $dataset_title_no_redirect = 'No Redirect Dataset';
     $this->submitForm([
-      'edit-field-json-metadata-0-value-title' => $dataset_title,
-      'edit-field-json-metadata-0-value-description' => 'Dataset Description.',
-      'edit-field-json-metadata-0-value-accesslevel' => 'public',
-      'edit-field-json-metadata-0-value-modified-date' => '2020-02-02',
-      'edit-field-json-metadata-0-value-publisher-publisher-name' => $publisher_name,
-      'edit-field-json-metadata-0-value-contactpoint-contactpoint-fn' => 'DKANTEST Contact Name',
-      'edit-field-json-metadata-0-value-contactpoint-contactpoint-hasemail' => 'dkantest@test.com',
-      'edit-field-json-metadata-0-value-keyword-keyword-0' => $keyword_data,
+      'edit-field-json-metadata-0-value-description' => 'Updated Dataset Description.',
     ], 'Save');
 
     // Assert that the user lands on the dataset node page (not redirected).
     $assert->statusCodeEquals(200);
     $assert->addressMatches('/node\/\d+$/');
-    $assert->pageTextContains('Data ' . $dataset_title_no_redirect . ' has been created.');
+    $assert->pageTextContains('Data ' . $dataset_title . ' has been updated.');
   }
 
 }
--- a/modules/metastore/modules/metastore_admin/tests/src/Functional/RedirectToDatasetsTest.php
+++ b/modules/metastore/modules/metastore_admin/tests/src/Functional/RedirectToDatasetsTest.php
@@ -35,7 +35,7 @@ class RedirectToDatasetsTest extends BrowserTestBase {
   /**
    * Tests dataset form submission when redirect_to_datasets is enabled.
    */
-  public function testRedirectEnabled() {
+  public function testDatasetRedirect() {
     /** @var \Drupal\metastore\MetastoreService $metastore_service */
     $metastore_service = $this->container->get('dkan.metastore.service');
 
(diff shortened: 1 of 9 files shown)


zedmonds96:
Need to add a test for this, moving back to draft.

zedmonds96:
~~There is a code climate issue with the metastore settings form having more than 25 lines. It has 29. In my opinion it is readable and well structured, so we should allow it as-is.~~

Updated the form structure to satisfy codeclimate now and in the future.

modules/metastore/modules/metastore_admin/tests/src/Functional/RedirectToDatasetsTest.php
```diff
@@ -0,0 +1,161 @@
...
+    // Assert that the redirect happened.
+    $assert->statusCodeEquals(200);
+    $assert->addressEquals('admin/dkan/datasets');
+    $assert->pageTextContains('Data ' . $dataset_title . ' has been created.');
+  }
+
+  /**
+   * Tests dataset form submission when redirect_to_datasets is disabled.
+   */
+  public function testRedirectDisabled() {
```
dafeder: @zedmonds96 there's significant overhead for every functional test that's run. Rather than having a whole second test, how about if we just change the config, edit and re-submit the same dataset as part of a single test?

modules/metastore/src/Form/DkanDataSettingsForm.php
```diff
@@ -74,49 +75,104 @@ public function getFormId() {
...
+   *
+   * @return array
+   *   The form element array.
+   */
+  private function getRedirectCheckbox(Config $config) {
+    return [
+      '#type' => 'checkbox',
+      '#title' => $this->t('Redirect to datasets view after form submit'),
+      '#default_value' => $config->get('redirect_to_datasets'),
+      '#description' => $this->t('Enable this option to automatically redirect
```
dafeder: This is pretty redundant with the title - let's either remove it or change it to provide additional information.

dafeder (changes requested):
Looks good @zedmonds96 - see one request about the test approach and another about form text

modules/metastore/src/Form/DkanDataSettingsForm.php
zedmonds96: I changed it, but if you'd rather I just remove it, that's fine too.

modules/metastore/modules/metastore_admin/tests/src/Functional/RedirectToDatasetsTest.php
zedmonds96: That makes sense. I made the change.

modules/metastore/src/Form/DkanDataSettingsForm.php
dafeder: Hmm - it's not quite right, because surely the person editing these settings (an admin) does have access 😄 

How about just something like "disable this option if you want to use Drupal's default or your own custom redirect after submitting a metadata form."

zedmonds96: That's true. Doesn't that mean that a non-admin user who might want to disable the redirect wouldn't have access to the setting? Should we disable it by default instead?

dafeder: It's a site-wide setting, it's only meant to be accessible to admins

modules/metastore/modules/metastore_admin/tests/src/Functional/RedirectToDatasetsTest.php
```diff
@@ -0,0 +1,117 @@
...
+
+    // Disable redirect option.
+    $this->config('metastore.settings')
+      ->set('redirect_to_datasets', FALSE)
+      ->save();
+
+    $this->drupalGet('admin/content');
+    $assert->statusCodeEquals(200);
+    $this->clickLink($dataset_title);
+    $this->drupalGet($this->getSession()->getCurrentUrl() . '/edit');
```
zedmonds96, line 105: @dafeder I tried to use `$this->clickLink('Edit');` but it was landing on the user edit page rather than the dataset edit page. I'm not sure if this is the best approach.

dafeder (approved):

PR #4386, opened 2025-01-16
Deprecate EventDispatcherTrait

### LINKED ISSUE #4313: Deprecate, disuse EventDispatcherTrait
### Current Behavior

Many services use `Drupal\common\EventDispatcherTrait` because we needed a backwards-compatibility layer between Drupal 8 and 9. See: https://github.com/GetDKAN/dkan/issues/3592

Currently, no one should be using Drupal 8, and we don't have Drupal 9 in our testing matrix.

Furthermore, services should be injected into other services, avoiding the `\Drupal` pattern.

So let's deprecate `EventDispatcherTrait` and remove its usage in DKAN.

### Expected Behavior

- Add a `@deprecated` annotation to `EventDispatcherTrait`.
- Inject the `event_dispatcher` service into classes that currently use the trait.
- Modify those classes to use their own dispatcher instead of the one called in the trait.

### Steps To Reproduce

See that we use EventDispatcherTrait.

Decide that it's not a good pattern.

Fix it.

_No response_

### Anything else?

_No response_

zedmonds96:
Fixes #4313

## Describe your changes

- Added a `@deprecated` annotation to `EventDispatcherTrait.php`
- All usages of the trait have been removed, replaced with direct injection of the `event_dispatcher` service.
- All classes now dispatch events through explicitly injected services instead of relying on global or trait-based dispatching.
- Since this is an injected service rather than a trait, many classes and tests needed updated with new arguments, etc. 

## QA Steps

- N/A

### DIFF
--- a/modules/common/src/Storage/JobStore.php
+++ b/modules/common/src/Storage/JobStore.php
@@ -3,6 +3,7 @@
 namespace Drupal\common\Storage;
 
 use Drupal\Core\Database\Connection;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 
 /**
  * Retrieve a serialized job (datastore importer or harvest) from the database.
@@ -26,14 +27,21 @@ class JobStore extends AbstractDatabaseTable {
    *   Table name for this jobstore table.
    * @param \Drupal\Core\Database\Connection $connection
    *   Database connection.
+   * @param \Symfony\Component\EventDispatcher\EventDispatcherInterface $eventDispatcher
+   *   The event dispatcher service.
    * @param string $deprecatedTableName
    *   (Optional) Deprecated table name, if there is one.
    */
-  public function __construct(string $tableName, Connection $connection, string $deprecatedTableName = '') {
+  public function __construct(
+    string $tableName,
+    Connection $connection,
+    EventDispatcherInterface $eventDispatcher,
+    string $deprecatedTableName = ''
+  ) {
     $this->tableName = $tableName;
     $this->deprecatedTableName = $deprecatedTableName;
     $this->setOurSchema();
-    parent::__construct($connection);
+    parent::__construct($connection, $eventDispatcher);
   }
 
   /**
--- a/modules/common/src/Storage/JobStore.php
+++ b/modules/common/src/Storage/JobStore.php
@@ -27,16 +27,16 @@ class JobStore extends AbstractDatabaseTable {
    *   Table name for this jobstore table.
    * @param \Drupal\Core\Database\Connection $connection
    *   Database connection.
+   * @param \Symfony\Component\EventDispatcher\EventDispatcherInterface $eventDispatcher
+   *   The event dispatcher service.
    * @param string $deprecatedTableName
    *   (Optional) Deprecated table name, if there is one.
-   * @param \Symfony\Component\EventDispatcher\EventDispatcherInterface $eventDispatcher
-   *    The event dispatcher service.
    */
   public function __construct(
     string $tableName,
     Connection $connection,
-    string $deprecatedTableName = '',
-    EventDispatcherInterface $eventDispatcher
+    EventDispatcherInterface $eventDispatcher,
+    string $deprecatedTableName = ''
   ) {
     $this->tableName = $tableName;
     $this->deprecatedTableName = $deprecatedTableName;
--- a/modules/common/src/Storage/JobStore.php
+++ b/modules/common/src/Storage/JobStore.php
@@ -3,6 +3,7 @@
 namespace Drupal\common\Storage;
 
 use Drupal\Core\Database\Connection;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 
 /**
  * Retrieve a serialized job (datastore importer or harvest) from the database.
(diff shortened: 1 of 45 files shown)


zedmonds96:
Codeclimate has complaints about 5 arguments in constructors instead of 4. Dan F. and Paul M. were in agreement that this could be ignored for now. Also the MetastoreServices file has a few too many lines. If this is an issue, I can try to reduce it in size, but I'm not sure how that could be accomplished.

modules/metastore/src/MetastoreService.php
```diff
@@ -78,12 +84,45 @@ public function __construct(
...
+   *   The event name.
+   * @param mixed $data
+   *   The data to dispatch.
+   * @param callable|null $validate
+   *   Optional validation callback.
+   *
+   * @return mixed
+   *   The (possibly) modified data.
+   */
+  private function dispatchEvent(string $eventName, $data, callable $validate = NULL) {
```
zedmonds96: This is one thing I'm not sure about. The getAll function dispatched an event with a validator function. This was the only way I could find to replicate that.

modules/metastore/src/MetastoreService.php
```diff
@@ -179,15 +218,8 @@ public function getAll(string $schema_id, ?int $start = NULL, ?int $length = NUL
...
 
     return $this->dispatchEvent(self::EVENT_DATA_GET_ALL, $objects, function ($data) {
-      if (!is_array($data)) {
-        return FALSE;
-      }
-      if (count($data) == 0) {
-        return TRUE;
-      }
-      return reset($data) instanceof RootedJsonData;
+      return !is_array($data) ? FALSE : (count($data) === 0 || reset($data) instanceof RootedJsonData);
```
zedmonds96: I think this is logically equivalent, I hope functionality is retained. It was not returning the correct value in it's original state.

modules/metastore/src/MetastoreService.php
```diff
@@ -78,12 +84,45 @@ public function __construct(
...
+   *   The data to dispatch.
+   * @param callable|null $validate
+   *   Optional validation callback.
+   *
+   * @return mixed
+   *   The (possibly) modified data.
+   */
+  private function dispatchEvent(string $eventName, $data, callable $validate = NULL) {
+    $event = new GenericEvent($data);
+    $event = $this->eventDispatcher->dispatch($event, $eventName);
```
dafeder:
```suggestion
    $this->eventDispatcher->dispatch($event, $eventName);
```

modules/common/tests/src/Unit/Events/EventTest.php
```diff
@@ -11,28 +10,26 @@
...
+      ->add(Container::class, 'get', $eventDispatcher)
       ->getMock();
 
     \Drupal::setContainer($container);
 
-    $this->dispatchEvent('test_event', 'hello', function($data) {
-      return is_string($data);
-    });
+    $event = new Event('hello');
+    $eventDispatcher->dispatch($event, 'test_event');
```
dafeder: Pass validation function to Event constructor, not dispatch()

modules/metastore/src/MetastoreService.php
dafeder: Let's try to pass the validator function to the event constructor like before, just need to use Drupal\common\Event instead of GenericEvent. Hopefully this means you can accomplish all this with just a couple lines and not need a dispatchEvent() method here.

dafeder (changes requested):

modules/metastore/tests/src/Unit/MetastoreServiceTest.php
kaise-lafrai, line 1: @dafeder How's this for a test?

dafeder: LGTM!

modules/common/src/Storage/JobStore.php
```diff
@@ -26,14 +27,21 @@ class JobStore extends AbstractDatabaseTable {
...
    * @param string $deprecatedTableName
    *   (Optional) Deprecated table name, if there is one.
    */
-  public function __construct(string $tableName, Connection $connection, string $deprecatedTableName = '') {
+  public function __construct(
+    string $tableName,
+    Connection $connection,
+    EventDispatcherInterface $eventDispatcher,
+    string $deprecatedTableName = ''
+  ) {
```
dafeder, line 40: Wondering why you added the eventDispatcher before $deprecatedTableName? Usually when you add new arguments to any function signature, especially a class constructor, you want to add them to the end to maintain compatibility with existing code or at least produce more readable errors.

dafeder (changes requested):
Just one small question @zedmonds96

dafeder (changes requested):
@zedmonds96 I think we decided we would just change the argument order back to the one that works. Let's get the tests passing again than figure out how to mitigate breaking changes.

dafeder (approved):
OK I think this is ready. This should I think be a minor/point release as it breaks the signature of a fairly important class (JobStore)

dafeder:
> OK I think this is ready. This should I think be a minor/point release as it breaks the signature of a fairly important class (JobStore)

As @paul-m pointed out, this should be fine as JobStores are always instantiated with a Factory, so if being used correctly this shouldn't break any custom code.

PR #4403, opened 2025-02-10
Improve type translation for database table

### LINKED ISSUE #4402: Column types from DB parsed incorrectly
Blocks #4356 

While debugging test failures with Mysql 8 over on #4401 we discovered that there were problems with how the type string is being parsed and described (we are trying to return a schema in JSON that follows Drupal's schema API). 

This occurs in the [`Drupal\datastore\Storage\DatabaseTable::translateType()`](https://github.com/GetDKAN/dkan/blob/2.x/modules/datastore/src/Storage/DatabaseTable.php#L234) function. We noticed the following issues:

1. `int unsigned` will not be correctly tokenized by `strtok()` and is interpreted as `varchar`
2. `int(11)` (or any number) is parsed to interpret the number in parentheses as the column length. This is not true; for mysql `int` the number represents the display length, which is now deprecated. It should be ignored
3. `decimal(3,2)` is correctly translated to `numeric` but with a `length` of 3. In fact, `numeric` columns should not have a length, but a `precision` and a `scale`. See the [Schema API documentation](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Database%21database.api.php/group/schemaapi/11.x). 

So we need smarter logic in that function.

Let's make `translateType()` a public function and write a simple unit test for it to make sure it conforms roughly to the acceptance criteria below. Recommend writing the test first, TDD FTW 😄 

## Acceptance criteria

Arguments ($type, $extra) | Return
---------------------------------------|----------
`["int unsigned", "auto_increment"]` | `{"type": "serial", "unsigned": true, "not null": true}`
`["int(11)", "auto_increment"]` | `{"type": "serial", "unsigned": true, "not null": true}`
`["int(11)", NULL]` | `{"type": "int"}`
`["int(11) unsigned", NULL]` | `{"type": "int", "unsigned": true}`
`["int unsigned", NULL]` | `{"type": "int", "unsigned": true}`
`["varchar(255)", NULL]` | `{"type": "varchar", "length": 255}`
`["text", NULL]` | `{"type": "text"}`
`["tinyint", NULL]` | `{"type": "int", "size": "tiny"}`
`["decimal(3,2)", NULL]` | `{"type": "numeric", "precision": 3, "scale": 2}`

Note: I think the date translations are working as expected now but worth double-checking.

dafeder:
Fixes #4402 

## Describe your changes

Changes the translateType() function to be a little smarter, deal with variations in type strings especially int/auto_increment fields between MySQL versions. Also properly records decimal fields as having precision and scale values rather than length.

## QA Steps

No real human QA necessary. Please review tests to make sure type translations appear accurate and improved. See #4402 for the acceptance criteria.

### DIFF
--- a/modules/datastore/src/Storage/DatabaseTable.php
+++ b/modules/datastore/src/Storage/DatabaseTable.php
@@ -235,7 +235,7 @@ class DatabaseTable extends AbstractDatabaseTable implements \JsonSerializable {
    *
    * @see https://api.drupal.org/api/drupal/core!lib!Drupal!Core!Database!database.api.php/group/schemaapi/9.2.x
    */
-  protected function translateType(string $describe_type, $extra = NULL) {
+  protected function translateType(string $describe_type, mixed $extra = NULL) {
     // Clean up things like "int(10) unsigned".
     $db_type = strtok($describe_type, ' ()');
     $driver = $this->connection->driver() ?? 'mysql';
--- a/modules/datastore/src/Storage/DatabaseTable.php
+++ b/modules/datastore/src/Storage/DatabaseTable.php
@@ -223,7 +223,7 @@ class DatabaseTable extends AbstractDatabaseTable implements \JsonSerializable {
    *
    * @param string $rawtype
    *   Type returned from the describe query.
-   * @param mixed $extra
+   * @param null|string $extra
    *   Additional information for column.
    *
    * @return array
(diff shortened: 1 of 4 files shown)


modules/datastore/tests/src/Unit/Storage/DatabaseTableTest.php
```diff
@@ -17,12 +17,105 @@
 use Psr\Log\LoggerInterface;
 
 /**
+ * @covers \Drupal\datastore\Storage\DatabaseTable
  * @group dkan
```
paul-m, line 22: Also needs `@coversDefaultClass \Drupal\datastore\Storage\DatabaseTable` in order to get the`@covers ::method` in the test methods.

modules/datastore/src/Storage/DatabaseTable.php
```diff
@@ -221,43 +221,55 @@ protected function addIndexInfo(array &$schema): void {
   /**
    * Translate the database type into a table schema type.
    *
-   * @param string $type
+   * @param string $rawtype
```
paul-m: Can we make a more descriptive variable name?

paul-m (changes requested):
Mostly LTGM. Requesting some formal changes.

paul-m (approved):

PR #4404, opened 2025-02-12
Refactor DatasetInfo as plugin-based service that works w/o datastore

### LINKED ISSUE #4360: Make DatasetInfo work without Datastore module enabled
In the context of #3746, we want to make `common` not depend on `datastore`. One place it does heavily is in the `DatasetInfo` class. Let's make sure this service can work if datastore is not enabled, and simply return less info. 

In the future, we should either set up some kind of plugin or subscriber system so that any module can add info to the core dataset info object. OR (even better) we should maybe find a way to stop using this class altogether. It is kind of an anti-pattern.

dafeder:
Fixes  #4360 

Blocks #4412 

## Describe your changes

Rather than requiring datastore services to fully initialize the DatasetInfo service, this PR:

a) Removes all Datastore dependencies from DatasetInfo
b) Adds a plugin type for DatasetInfo to allow other modules to add information to the DatasetInfo array
c) Adds a plugin for the datastore module to add the properties that depend on it
b) Moves most DatasetInfo tests to a Kernel test. Since DatasetInfo::gather() is basically just integrates data returned from other services, unit tests are of low value and require an excessive amount of mocking. 

## Notes

- The new plugin type is not documented very extensively. We see this whole system as an anti-pattern - we do not want to encourage further dependence on the DatasetInfo object, just to make it possible to decouple the dkan submodules sensibly.
- The metastore dependencies are still injected with methods, rather than the DatasetInfo constructor. This is because of the split between common and metastore; the modules cannot be enabled in an order that satisfies the service dependency checker. As we move forward with the plan to combine dkan, common and metastore into a single module called "dkan" we can change these back to constructor arguments.
- I don't think there is a better approach to either of the two codeclimate errors. Recommend overriding them but open to suggestions.

## QA Steps

No QA - DatasetInfo will continue to behave identically.


kaise-lafrai:
### QA Notes

### 2.X BRANCH
Here is what the dataset info looks like for the sample content for `ddev drush dkan:dataset-info cedcd327-4e5d-43f9-8eb1-c11850fa7c55`

```
{
    "latest_revision": {
        "uuid": "cedcd327-4e5d-43f9-8eb1-c11850fa7c55",
        "node_id": "8",
        "revision_id": "8",
        "moderation_state": "published",
        "title": "Florida Bike Lanes ",
        "modified_date_metadata": "2016-06-22",
        "modified_date_dkan": "2025-02-19T14:35:54-0500",
        "distributions": [
            {
                "distribution_uuid": "9f336d47-b603-5b0e-875d-7940a581cffc",
                "resource_id": "3a187a87dc6cd47c48b6b4c4785224b7",
                "resource_version": "1739993755",
                "fetcher_status": "done",
                "fetcher_percent_done": 100,
                "file_path": "public:\/\/resources\/3a187a87dc6cd47c48b6b4c4785224b7_1739993755\/Bike_Lane.csv",
                "mime_type": "text\/csv",
                "source_path": "https:\/\/h-o.st\/sites\/default\/files\/distribution\/cedcd327-4e5d-43f9-8eb1-c11850fa7c55\/Bike_Lane.csv",
                "importer_percent_done": 100,
                "importer_status": "done",
                "importer_error": "",
                "table_name": "datastore_a5d2b93007325b39ec82e00ac7951b9f"
            }
        ]
    }
}
```

### THIS BRANCH
Here is what the dataset info looks like for the sample content for `ddev drush dkan:dataset-info cedcd327-4e5d-43f9-8eb1-c11850fa7c55`

```
{
    "latest_revision": {
        "uuid": "cedcd327-4e5d-43f9-8eb1-c11850fa7c55",
        "node_id": "8",
        "revision_id": "8",
        "moderation_state": "published",
        "title": "Florida Bike Lanes ",
        "modified_date_metadata": "2016-06-22",
        "modified_date_dkan": "2025-02-19T14:35:54-0500",
        "distributions": [
            {
                "distribution_uuid": "9f336d47-b603-5b0e-875d-7940a581cffc",
                "resource_id": "3a187a87dc6cd47c48b6b4c4785224b7",
                "resource_version": "1739993755",
                "mime_type": "text\/csv",
                "source_path": "https:\/\/h-o.st\/sites\/default\/files\/distribution\/cedcd327-4e5d-43f9-8eb1-c11850fa7c55\/Bike_Lane.csv",
                "fetcher_status": "done",
                "fetcher_percent_done": 100,
                "file_path": "public:\/\/resources\/3a187a87dc6cd47c48b6b4c4785224b7_1739993755\/Bike_Lane.csv",
                "importer_percent_done": 100,
                "importer_status": "done",
                "importer_error": "",
                "table_name": "datastore_a5d2b93007325b39ec82e00ac7951b9f"
            }
        ]
    }
}
```

dafeder:
@kaise-lafrai I don't see any differences in the two outputs -- am I missing something?

kaise-lafrai:
> @kaise-lafrai I don't see any differences in the two outputs -- am I missing something?

@dafeder I was just putting my QA steps and results in a comment. Nothing needed from you at the moment! Thanks

modules/datastore/src/Plugin/DatasetInfo/DatastoreInfo.php
```diff
@@ -0,0 +1,157 @@
...
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Plugin implementation of the dataset_info.
+ *
+ * @DatasetInfoPlugin(
+ *   id = "datastore_info",
+ * )
+ */
+class DatastoreInfo extends DatasetInfoPluginBase {
```
kaise-lafrai, line 21: Q: Should this class be called DatasetInfo?

modules/common/src/DatasetInfoPluginManager.php
```diff
@@ -0,0 +1,26 @@
...
+ * DatasetInfo plugin manager.
+ */
+final class DatasetInfoPluginManager extends DefaultPluginManager {
+
+  /**
+   * Constructs the object.
+   */
+  public function __construct(\Traversable $namespaces, CacheBackendInterface $cache_backend, ModuleHandlerInterface $module_handler) {
+    parent::__construct('Plugin/DatasetInfo', $namespaces, $module_handler, DatasetInfoPluginInterface::class, DatasetInfoPlugin::class);
+    $this->alterInfo('dataset_info_info');
```
kaise-lafrai, line 22: Q: Should this be 'dataset_info' instead of 'dataset_info_info'?

modules/datastore/src/Plugin/DatasetInfo/DatastoreInfo.php
dafeder: I don't think so, I think it should be close to the plugin ID. Maybe DatastoreDatasetInfo would be more precise but that seems kind of silly.

modules/common/src/DatasetInfoPluginManager.php
dafeder: This is a weird one, I don't think we are really ever going to use an alter hook on this stuff, but it seems to be included in plugins by convention. dataset_info is the plugin type and _info is the standard suffix for an info_alter hook. So, this is weird but i think it's correct.

modules/datastore/src/Plugin/DatasetInfo/DatastoreInfo.php
```diff
@@ -0,0 +1,157 @@
...
+   * @param \Drupal\metastore\ResourceMapper $resourceMapper
+   *   Resource mapper service.
+   */
+  public function __construct(
+    array $configuration,
+    $pluginId,
+    $pluginDefinition,
+    ImportInfo $importInfo,
+    DatastoreService $datastore,
+    ResourceMapper $resourceMapper,
```
kaise-lafrai: ResourceMapper can be retrieved from the DatastoreService, so you could remove` ResourceMapper $resourceMapper,`

modules/datastore/src/Plugin/DatasetInfo/DatastoreInfo.php
```diff
@@ -0,0 +1,157 @@
...
+   * Alter the distribution info.
+   *
+   * @param array $distribution
+   *   The distribution info array from a dataset info array.
+   */
+  protected function addDistributionInfo(array &$distribution): void {
+    $identifier = $distribution['resource_id'];
+    $version = $distribution['resource_version'];
+    $import_info = $this->importInfo->getItem($identifier, $version);
+    $fileMapper = $this->resourceMapper->get($identifier, ResourceLocalizer::LOCAL_FILE_PERSPECTIVE, $version);
```
kaise-lafrai, line 129: You could do `$this->datastore->getResourceMapper()->get($identifier, ResourceLocalizer::LOCAL_FILE_PERSPECTIVE, $version);`

kaise-lafrai (changes requested):
A few recommendations in the comments.

modules/datastore/src/Plugin/DatasetInfo/DatastoreInfo.php
dafeder: Set in constructor

kaise-lafrai (approved):

kaise-lafrai:
Rebased. Will merge once checks pass.

PR #4410, opened 2025-02-19
#3611: Pass progress indicator over to the managed file widget through upload_or_link widget

### LINKED ISSUE #3611: Saving a dataset before the file upload has completed silently drops the file
## Describe the bug

When uploading a file the file name is shown in the upload form control whilst the file is uploading - there isn't any progress indicator and it's not immediately apparent that the file is not uploaded yet, unless you have used this before and know that it turns the file into a link when complete. There is also not anything preventing you from saving the node while the file is uploading, which causes the file to be omitted without any notification.

## Steps To Reproduce

- Get a large file (e.g. https://data.medicaid.gov/api/views/6gk3-9bxc/rows.csv?accessType=DOWNLOAD)
- Fill in dataset required fields
- Select the file for upload and click "upload"
- Then quickly press "Save" to save the dataset

## Expected behavior

Either or both:
- There is a clearer progress indicator
- You are prevented (with a helpful error message) from saving the node while a file is uploading

stefan-korn:
Fixes #3611

### DIFF
--- a/modules/json_form_widget/src/WidgetRouter.php
+++ b/modules/json_form_widget/src/WidgetRouter.php
@@ -303,6 +303,9 @@ class WidgetRouter implements ContainerInjectionInterface {
     if (isset($spec->extensions)) {
       $element['#upload_validators']['FileExtension'] = ['extensions' => $spec->extensions];
     }
+    if (isset($spec->progress_indicator)) {
+      $element['#progress_indicator'] = $spec->progress_indicator;
+    }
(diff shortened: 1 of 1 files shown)


dafeder:
Awesome thanks for this @stefan-korn will take a look

beeyayjay (commented):
Confirmed that code change results in progress indicator appearing at start of file upload and disappearing when upload is complete.

beeyayjay:
Closing in favor of #4433 (added test coverage)

Issue #4368, opened 2024-12-20
Improve and modernize file handling in Harvest module


dafeder:
Following up on #4276, we need to make DrupalFiles D11-compatible, and secondarily to be compatible with vfsStream so that we can write kernel tests for it.
 
- [x] Remove system_retrieve_file() from common\DrupalFiles
- [ ] Refactor DrupalFiles to play nice with vfsStream
- [x] Turn Drupal\Tests\harvest\Functional\Transform\ResourceImporterTest into a kernel test.

dafeder:
Think we're good here - honestly not sure if we've technically refactored to "play nice with vfsStream" but we've removed `system_retrieve_file()` and now have kernel tests.

PR #4420, opened 2025-02-27
Refactor DrupalFiles to remove deprecated system_retrieve_file() 

dafeder:
Part of #4368

Continuing work on #4283 by @paul-m 

Note that all logic dealing with managed files is removed. In the old DrupalFiles, $managed was always passed as FALSE, so there was no way files would ever be managed.

## Describe your changes

Refactors our code to not use `system_retrieve_file()`.

Required for Drupal 11.

Change record: https://www.drupal.org/node/3223362


modules/common/src/Util/DrupalFiles.php
```diff
@@ -98,7 +123,65 @@ public function retrieveFile($url, $destination) {
       return $this->fileCreateUrl("{$destination}/{$filename}");
     }
     // Handle http(s):// URIs.
-    return system_retrieve_file($url, $destination, FALSE, FileSystemInterface::EXISTS_REPLACE);
+    return $this->retrieveRemoteFile($url, $destination);
```
dafeder, line 126: Because this is the only call to `system_retrieve_file()`, and we always pass it `false` for managed and replace for file exists, we don't need to support parameters for this. Meaning we don't need to have any support at all for managed files in this class.

paul-m (approved):

PR #4421, opened 2025-02-27
Trim string value for converting to bool.

beeyayjay:
Fixes [issue#]
20456

## Describe your changes
Trims field before comparing to 'TRUE' or 'FALSE' during conversion to boolean (actually tinyint) when applying a data dictionary.

## QA Steps


dafeder (changes requested):
@beeyayjay I know it was already missing but I think we need to add some test coverage here to prove it works and keep this maintainable. From what I see this is tested in the DictionaryEnforcerTest functional tests, where it does try to convert to a boolean (in practice, TINYINT) but the sample data is already just 1 or 0. Do you mind updating that test so that 
1. There is a second bool column in the data and data dictionary that uses true/false text vlues
2. Includes one row where the text does not need to be trimmed, and one where it does
3. Asserts that "true" and "false" are correctly converted to 1 or 0?

dafeder (approved):

PR #4427, opened 2025-03-06
Move the harvest library into the harvest module

### LINKED ISSUE #4373: Bring GetDKAN/harvest library into DKAN harvest module
Related to #3821 

We have no reason anymore to split Harvest between a module and an external library. Let's bring those library classes into the Harvest module and kill one more dependency. 

## Acceptance criteria

1. All the classes in GetDKAN/harvest (`Harvest` namespace) are moved into the DKAN harvest submodule. References throughout DKAN are changed to reflect new namespace
2. Incoming code changed to meet Drupal coding/doc standards (currently using simply PSR standards)
3. Move tests in as well. Refactor to avoid the existing tests' dependency on contracts library.

beeyayjay:
Fixes #4373
see also 24373

## Describe your changes
Moved code from harvest library (GetDKAN/harvest) into harvest module (GetDKAN/dkan/modules/harvest), and removed harvest library from DKAN build.

## QA Steps

- [ ] enable sample-content module (ddev drush en sample-content)
- [ ] create sample content (ddev drush dkan:sample-content:create)
- [ ] run various harvest commands against sample-content (e.g. `ddev drush dkan:harvest:list`, `ddev drush dkan:harvest:info sample-content`,  `ddev drush dkan:harvest:revert sample-content`,  `ddev drush dkan:harvest:run sample-content`,  `ddev drush dkan:harvest:archive sample-content`,   `ddev drush dkan:harvest:archive sample-content`)

- [x] I have updated or added tests to cover my code - added test for db update.
- [x] I have updated or added documentation - minor updates reflecting namespace changex


dafeder (changes requested):
This looks good but I'm pretty sure that an upgrade would break existing harvests that are referring to classes in the Harvest\ETL namespace. So I think we need an update hook that will fix those. I'm working on steps to demonstrate the issue. We should probably do an upgrade test for this as well.

dafeder:
@beeyayjay to add update test

modules/harvest/tests/src/Functional/HarvestCodeConsolidationTest.php
```diff
@@ -0,0 +1,66 @@
...
+
+
+    try {
+      $result = $harvest_service->runHarvest('sample_content');
+    }
+    catch (\Exception $exception) {
+      $result = $exception->getMessage();
+    }
+    $this->assertIsArray($result);
+    $this->assertSame('SUCCESS', $result['status']['extract']);
```
dafeder: @beeyayjay it's confusing to have try/catches here, because it's not clear when you read the test which one is happening. Is it throwing an exception in each case? If so, let's not set $result in the try, because it's not clear which one is in the assertion. If it's not throwing an exception, we don't need a try/catch at all.

Also not sure what this test shows, maybe I'm missing something. Why does the harvest have different results before and after the update? Aren't we trying to demonstrate that it works the same?

dafeder (changes requested):

modules/harvest/tests/src/Functional/HarvestCodeConsolidationTest.php
beeyayjay: @dafeder we're trying to demonstrate that the update works as intended. `runHarvest()` fails before the update because the extractor namespace in the db is outdated. It should succeed after the update is applied. (Note that the second assignment of `$harvest_service` on L54 is required to load the updated data.)

I think I can replace the first try/catch with an `expectException()` and leave the second catch blank. The second `runHarvest()` *shouldn't* fail, but that's really what we're testing, and a failed assertion seems preferable to an uncaught exception if the test fails.

beeyayjay: I had to leave the first try/catch block in. `expectException()` worked, but the rest of the test didn't run.

I rewrote it a bit though, so hopefully it's clearer now.

modules/harvest/tests/src/Functional/HarvestCodeConsolidationTest.php
```diff
@@ -0,0 +1,57 @@
...
+
+    // Run update and recreate harvest service with updated data.
+    $this->runUpdates();
+    $harvest_service = \Drupal::service('dkan.harvest.service');
+
+    try {
+      $result = $harvest_service->runHarvest('sample_content');
+    }
+    catch (\Exception $exception) {
+    }
```
dafeder: I must be missing something - there is no reason for a try/catch here, because no exception is expected, right?
```suggestion
    $result = $harvest_service->runHarvest('sample_content');
```

dafeder (changes requested):
@beeyayjay sorry almost there, see my last comment on the test please

modules/harvest/tests/src/Functional/HarvestCodeConsolidationTest.php
dafeder: OK sorry re-reading this the logic of the test does make sense. But I'm still failing to see why the second try/catch is necessary. Removing the `$result =` does make the first more readable and the test more valid, thanks. See my suggestion below.

modules/harvest/tests/src/Functional/HarvestCodeConsolidationTest.php
beeyayjay: We don't expect an exception, but we also don't expect the test to fail. If there's no exception the test will almost certainly pass. 

The only value to the catch block, is that in the case of failure, it will fail on a failed assertion (` $this->assertIsArray($result);`) instead of on an exception.

dafeder: Let's please remove the try/catch. We should never have a try/catch in a test unless we definitely expect an exception, as above. Having the test throw an error rather than a failure if an unexpected exception is thrown is the correct behavior.

beeyayjay: Thanks! I'm pretty new to writing unit tests, still learning the conventions.

dafeder (approved):

PR #4409, opened 2025-02-18
PHP 8.4 Compatibility

dafeder:
Fixes [issue#]

## QA Steps

- [ ] Add manual QA steps in checklist format for a reviewer to perform. Be as specific as possible, provide examples if appropriate.

## Checklist before requesting review

_If any of these are left unchecked, please provide an explanation_

- [ ] I have updated or added tests to cover my code
- [ ] I have updated or added documentation

PR #4428, opened 2025-03-07
Add PHP 8.4 compatibility/test coverage

### LINKED ISSUE #4366: Support PHP 8.4
### Feature Description

Ensure that a site with DKAN can be built and pass tests with PHP 8.4.

Druapl 10.4 and 11.1 both support PHP 8.4. Reading the release notes it doesn't look like we should expect any issues with it. Let's add it to the testing matrix and work on any issues if there are failures. At least one dependency -- getdkan/file-fetcher -- explicitly has 8.3 as the max compatible version so at minimum we need a new release of that.

### Target Audience

- [ ] End Users
- [x] Developers
- [ ] Administrators
- [ ] Other

### Dependencies

|Package|Has PHP 8.4|PR|Notes|
|--|--|--|--|
|getdkan/dkan|✅|GetDKAN/dkan#4428|DKAN itself is in here because it can have passing builds w/o all dependencies having passing builds (CI problem not actual runtime problem)|
|getdkan/contracts|✅|||
|getdkan/csv-parser|✅|GetDKAN/csv-parser#22 merged||
|getdkan/file-fetcher|✅|||
|getdkan/harvest|✅|||
|getdkan/mock-chain|✅|GetDKAN/mock-chain#24||
|getdkan/pdlt|✅|GetDKAN/pdlt#6||
|getdkan/procrastinator|✅|||
|getdkan/rooted-json-data|✅|GetDKAN/RootedJsonData#19|opis/jsonSchema has many deprecations. Needs to be updated for PHP 8.4. See: opis/json-schema#151|

dafeder:
Fixes #4366
Replaces #4409 

Drops support for Drupal 10.2.


dafeder:
Interesting, it's passing for D11 but not for 10.4. Major issue seems to be robo which does not have a 5.x release (even 5.x-dev seems hard to pull in due to other dependency issues)

dafeder:
Check 6c6983d make sure these test tweaks seem ok.

modules/common/src/Util/DrupalFiles.php
```diff
@@ -101,19 +101,19 @@ public function getStreamWrapperManager(): StreamWrapperManager {
    * the newly stored file.
    */
   public function retrieveFile($url, $destination) {
```
paul-m: Can we just cast $url as string in the function declaration and not have to cast it everywhere it's used?

modules/common/tests/src/Traits/GetDataTrait.php
```diff
@@ -29,7 +29,7 @@ private function getDownloadUrl(string $filename) {
    * @return string|false
    *   Json encoded string of this dataset's metadata, or FALSE if error.
    */
-  private function getDataset(string $identifier, string $title, array $downloadUrls, bool $localFiles = FALSE, string $describedBy = NULL) {
+  private function getDataset(string $identifier, string $title, array $downloadUrls, bool $localFiles = FALSE, ?string $describedBy = NULL) {
```
paul-m, line 35: Update the docblock.

modules/data_dictionary_widget/src/Plugin/Field/FieldWidget/DataDictionaryWidget.php
```diff
@@ -42,7 +42,7 @@ public function formElement(FieldItemListInterface $items, $delta, array $elemen
     $element = FieldCreation::createGeneralFields($element, $field_json_metadata, $current_fields, $form_state);
 
     $element['dictionary_fields']['#pre_render'] = [
-      [$this, 'preRenderForm'],
+      $this->preRenderForm(...),
```
paul-m, line 45: Is this still a callable?

modules/datastore/src/Drush.php
```diff
@@ -52,8 +52,6 @@ class Drush extends DrushCommands {
 
   /**
    * The PostImportResultFactory service.
-   *
-   * @var \Drupal\datastore\PostImportResultFactory
```
paul-m, line 56: I'm surprised CC didn't flag this and the other @var removals. Can we modify rector so it leaves 'unneeded' @vars behind? I guess we also need to talk about coding standards in general and tooling in particular.

modules/datastore/src/Service/Info/ImportInfo.php
```diff
@@ -56,13 +52,13 @@ class ImportInfo {
...
    * Datastore service.
-   *
-   * @var \Drupal\datastore\DatastoreService
    */
   protected DatastoreService $datastoreService;
 
   /**
    * Constructor.
+   *
+   * @todo Remove the $importServiceFactory argument; it's not used.
```
paul-m, line 61: Do we have a follow-up issue?

modules/datastore/src/Service/PostImport.php
```diff
@@ -219,13 +207,8 @@ protected function createPostImportResult($status, $message, DataResource $resou
...
   public function drop($resourceIdentifier): bool {
-    try {
-      $this->datastoreService->drop($resourceIdentifier, NULL, FALSE);
-      return TRUE;
-    }
-    catch (\Exception $e) {
-      throw $e;
-    }
+    $this->datastoreService->drop($resourceIdentifier, NULL, FALSE);
+    return TRUE;
```
paul-m, line 211: Boggle. How did we end up with a no-op try/catch in the first place?

modules/datastore/tests/src/Unit/Controller/QueryDownloadControllerTest.php
```diff
@@ -85,13 +83,13 @@ private function queryResultCompare($data, $resource = NULL) {
     $csv = $response->getContent() ?? '';
 
     $dController = QueryDownloadController::create($this->getQueryContainer(25));
-    ob_start([self::class, 'getBuffer']);
+    ob_start(self::getBuffer(...));
```
paul-m, line 86: Is this still a callable?

modules/datastore/src/Controller/QueryDownloadController.php
```diff
@@ -126,7 +126,7 @@ private function initStreamedCsvResponse($filename = "data.csv") {
    *   Row of data to send as CSV.
    */
   private function sendRow($handle, array $row) {
-    fputcsv($handle, $row);
+    fputcsv($handle, $row, escape: "\\");
```
paul-m, line 129: Goodbye PHP 7.4, but... Do we need to specify the default value for escape?

Also, since we're doing this 8.0+ stuff, we should specify PHP 8.0+ in composer.json.

modules/frontend/src/Page.php
```diff
@@ -12,39 +12,33 @@ class Page {
...
-   *
-   * @var string
    */
-  private $frontendPath;
+  private string $frontendPath;
 
   /**
    * Constructor.
+   *
+   * @todo Remove $nodeStorage argument; it's not used.
```
paul-m, line 36: Do we have a followup issue for this?

modules/frontend/src/Routing/RouteProvider.php
```diff
@@ -49,11 +47,13 @@ class RouteProvider {
 
   /**
    * Constructor.
+   *
+   * @todo Remove the $entityQuery argument; it's not used.
```
paul-m, line 51: Do we have a followup issue for this?

paul-m (changes requested):

modules/datastore/src/Controller/QueryDownloadController.php
dafeder: We do, default value is [deprecated](https://www.php.net/manual/en/function.str-getcsv.php). I had to do these manually, actually, and [filed an issue with rector to fix automatically in the future](https://github.com/rectorphp/rector/issues/9057). (I think it might actually be more correct to just use `''` but this seemed more conservative as it was the old default.)

Will make the change in composer.json

modules/datastore/src/Drush.php
dafeder: I think we told codeclimate to skip that or something? Last time this came up we agreed to remove these when they were redundant (when the variable is already typed), I very much prefer that.

modules/datastore/src/Service/Info/ImportInfo.php
dafeder: No, pretty low-priority. Do we need to have an issue for every @todo? If so we have a lot of catchup to do :)

modules/datastore/src/Service/PostImport.php
dafeder: lol no idea

modules/common/src/Util/DrupalFiles.php
dafeder: You can't "cast" in a signature right? That's just validation I think? But we can add `$url = (string) $url;` at the top.

paul-m: You can say `public function retrieveFile(string $url, $destination) {` and that means `$url` is already a string inside the method, and then you won't have to cast it as a string to please the linter.

modules/datastore/src/Drush.php
paul-m: ok. no biggie.

modules/datastore/src/Service/Info/ImportInfo.php
paul-m: well, yes, indeed we have a lot of catchup to do. :-) But not a blocker.

modules/datastore/src/Controller/QueryDownloadController.php
paul-m: Nice catch on the deprecation.

modules/common/src/Util/DrupalFiles.php
dafeder: Right; what I was concerned about was that `$url` could be another type that's cast-able to string, which would work as it is now but would throw an error if we added a type hint. But I was mostly thinking about `Drupal\Core\Url`, which a) I think is not actually cast-able to `string` (it has a `::toString()` but not a `::__toString()`), and b) the only place in DKAN this method is called is in `ResourceImporter`, where `$url` can only be a `string`. So I think it's ok to type hint here.

modules/datastore/src/Controller/QueryDownloadController.php
dafeder: It was very noisy, hard to miss! Re: version constraint, we don't have one in DKAN now. In our module info we require DKAN ^10.2, and that [requires PHP ^8.1](https://git.drupalcode.org/project/drupal/-/blob/10.2.x/composer.json?ref_type=heads#L58), so I don't think we need to maintain an explicit PHP requirement for our module composer.json?

paul-m (approved):

PR #4429, opened 2025-03-07
Dashboard updates to display datasets with unsupported mime_types

### LINKED ISSUE #4332: Import status dashboard pagination issue - follow-up to #4317

kaise-lafrai:
Fixes #4332 

Updates made to the dashboard to include datasets with unsupported mime_types but with a new status message 
and styling to distinguish it from other rows and status.

## QA Steps

- [x] Set up vanilla site
- [x] ddev dkan-sample-content
- [x] run cron a few times
- [x] navigate to the dashboard at (/admin/dkan/datastore/status)
- [x] confirm that resources that are not importable (like PDF, DOC, ZIP etc.) appear with a light gray background in the resource column, and a status that spans across fetch/store/post import with the following message: _Data import is not supported for this resource type_
- [x] At this point pagination should not be made available on the page because there is only 10 datasets. 
- [x] Create one more dataset to make it 11 total datasets. Be sure that this dataset has a resource file that is a pdf or zip.
- [x] navigate to the dashboard at (/admin/dkan/datastore/status)
- [x] confirm that you now see the pagination, select next page
- [x] confirm you see a single row and that the resourced, fetch, store, post import all have the light gray background and a status that spans across fetch/store/post import with the following message: _Data import is not supported for this resource type_

### DIFF
--- a/modules/datastore/src/Form/DashboardForm.php
+++ b/modules/datastore/src/Form/DashboardForm.php
@@ -454,18 +454,12 @@ class DashboardForm extends FormBase {
     // Create a row for each dataset revision (there could be both a published
     // and latest).
     foreach ($datasetInfo as $rev) {
-      // Filter out distributions whose resources are not csv or tsv.
-      $distributions = array_filter($rev['distributions'], function ($v) {
-        return !isset($v['mime_type']) || in_array($v['mime_type'], DataResource::IMPORTABLE_FILE_TYPES);
-      });
-
-      if (!empty($distributions)) {
-        // For first distribution, combine with revision information.
-        $rows[] = array_merge(
-          $this->buildRevisionRow($rev, count($distributions), $harvestStatus),
-          $this->buildResourcesRow(array_shift($distributions))
-        );
-      }
+      $distributions = $rev['distributions'];
+      // For first distribution, combine with revision information.
+      $rows[] = array_merge(
+        $this->buildRevisionRow($rev, count($distributions), $harvestStatus),
+        $this->buildResourcesRow(array_shift($distributions))
+      );
       // If there are more distributions, add additional rows for them.
       while (!empty($distributions)) {
         $rows[] = $this->buildResourcesRow(array_shift($distributions));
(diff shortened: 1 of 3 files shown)


kaise-lafrai:
@dafeder @stefan-korn this is ready for review. Stefan feel free to test this with your builds.

modules/datastore/src/Form/DashboardForm.php
```diff
@@ -537,26 +531,51 @@ protected function buildResourcesRow($dist): array {
...
+
     return ['', '', '', ''];
   }
 
+  /**
+   * Build resource data array.
+   *
+   * @param array $dist
+   *   Distribution details.
+   * @param bool $mime_type
```
dafeder: Can we call this `$importable`? It seems counter-intuitive that `$mime_type` is a bool, I'd expect it to contain the mimetype.

modules/datastore/src/Form/DashboardForm.php
```diff
@@ -537,26 +531,51 @@ protected function buildResourcesRow($dist): array {
...
     }
+
     return ['', '', '', ''];
   }
 
+  /**
+   * Build resource data array.
+   *
+   * @param array $dist
+   *   Distribution details.
```
dafeder: Specify that this is a distribution element from a datastore_info array

modules/datastore/src/Form/DashboardForm.php
```diff
@@ -537,26 +531,51 @@ protected function buildResourcesRow($dist): array {
...
+   * @param bool $mime_type
+   *   Whether the mime type is importable.
+   * @param string $postImportStatus
```
dafeder:
```suggestion
   * @param string $post_import_status
```
Use snake_case for consistency

modules/datastore/src/Form/DashboardForm.php
```diff
@@ -537,26 +531,51 @@ protected function buildResourcesRow($dist): array {
...
+      [
+        'data' => [
+          '#theme' => 'datastore_dashboard_resource_cell',
+          '#uuid' => $dist['distribution_uuid'],
+          '#file_name' => basename($dist['source_path']),
+          '#file_path' => UrlHostTokenResolver::resolve($dist['source_path']),
+        ],
+        'class' => $mime_type ? '' : 'unsupported',
+      ],
+      $mime_type ? $this->buildStatusCell($dist['fetcher_status']) : $this->buildStatusCell('unsupported'),
```
dafeder:
```suggestion
      $this->buildStatusCell($mime_type ? $dist['fetcher_status'] : 'unsupported'),
```

modules/datastore/src/Form/DashboardForm.php
```diff
@@ -537,26 +531,51 @@ protected function buildResourcesRow($dist): array {
...
+      $mime_type ? $this->buildStatusCell($dist['importer_status'], $dist['importer_percent_done'], $this->cleanUpError($dist['importer_error'])) : '',
```
dafeder: Can we split this into multiple lines?

modules/datastore/src/Form/DashboardForm.php
```diff
@@ -537,26 +531,51 @@ protected function buildResourcesRow($dist): array {
...
+      $mime_type ? $this->buildPostImportStatusCell($postImportStatus, $error) : '',
+    ];
+
+    // Remove empty cells.
+    return array_filter($data);
```
dafeder, line 579: I'm not sure why you're using `array_filter()` here. a) we would want there to always be the same number of elements in this array, since they represent columns in a table, right? Also, `array_filter` only removes cells that are `null` or `false` and I don't think your logic allows for that (if `$mime_type` is `false`, we add an empty string not `null`)? So I don't think this does anything.

modules/datastore/src/Form/DashboardForm.php
dafeder: Also, it seems wrong that this can add a structured array or else an empty string. Usually you'd want the array to contain elements of the same type, or else at least something simpler like NULL or FALSE to represent "nothing".

modules/datastore/src/Form/DashboardForm.php
```diff
@@ -571,15 +590,23 @@ protected function buildResourcesRow($dist): array {
...
+        '#status' => $status === 'unsupported' ? 'Data import is not supported for this resource type' : $status,
         '#percent' => $percentDone ?? NULL,
         '#error' => $error,
       ],
       'class' => str_replace('_', '-', $status),
     ];
+
+    // If the status is unsupported, we want to span the cell across
+    // all three columns.
+    if ($status === 'unsupported') {
```
dafeder: Use a class constant rather than an arbitrary string for comparisons

modules/datastore/src/Form/DashboardForm.php
```diff
@@ -571,15 +590,23 @@ protected function buildResourcesRow($dist): array {
    *   Renderable array.
    */
   protected function buildStatusCell(string $status, ?int $percentDone = NULL, ?string $error = NULL) {
-    return [
+    $statusCell = [
       'data' => [
         '#theme' => 'datastore_dashboard_status_cell',
-        '#status' => $status,
+        '#status' => $status === 'unsupported' ? 'Data import is not supported for this resource type' : $status,
```
dafeder: Use a class constant like `DashboardForm::TYPE_UNSUPPORTED` rather than an arbitrary string for comparisons

modules/datastore/src/Form/DashboardForm.php
```diff
@@ -537,26 +531,51 @@ protected function buildResourcesRow($dist): array {
...
+   *   Distribution details.
+   * @param bool $mime_type
+   *   Whether the mime type is importable.
+   * @param string $postImportStatus
+   *   Post import status.
+   * @param string|null $error
+   *   Error message, if any.
+   *
+   * @return array
+   *   Resource data array.
```
dafeder: Can we explain a bit more what's in the return here? I think this could be described an array of render arrays representing the last three columns of the dashboard table?

dafeder (changes requested):

dafeder:
QA is a 👍🏽

modules/datastore/src/Form/DashboardForm.php
dafeder: OK I was wrong here, array_filter is removing the items that just contain `''`. AS I said above I still think it would be more correct to add a `NULL` array element but this line is fine

dafeder (approved):

kaise-lafrai:
@janette @dafeder Here is page 1 and page 2 with the changes introduced in this PR.

image

image

PR #4438, opened 2025-03-12
Update some missing/incomplete/incorrect docs

dafeder:
Just a couple cleanup things that will prevent the codeclimate errors in #4419.

### DIFF
--- a/modules/metastore/src/Exception/AlreadyRegistered.php
+++ b/modules/metastore/src/Exception/AlreadyRegistered.php
@@ -12,7 +12,7 @@ class AlreadyRegistered extends \Exception {
   /**
    * An array of entities that have already been registered in the mapping.
    *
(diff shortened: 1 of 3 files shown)


modules/metastore/src/Exception/AlreadyRegistered.php
```diff
@@ -12,7 +12,7 @@ class AlreadyRegistered extends \Exception {
   /**
    * An array of entities that have already been registered in the mapping.
    *
-   * @var \Drupal\Core\Entity\EntityInterface[]
+   * @var \Drupal\Core\Entity\ContentEntityInterface[]
```
dafeder, line 15: We call `get()` from `$alreadyRegistered` over in `Drupal\metastore\Reference\Referencer` so general `EntityInterface` is not specific enough/

janette (approved):

PR #4442, opened 2025-03-14
Initial update path tests

### LINKED ISSUE #4393: Add upgrade tests
We don't have tests for old installs that new to be upgraded.

- Run tests against the last upgrade of the latest release tag
- This is the base template that you should use [\Drupal\FunctionalTests\Update\UpdatePathTestBase](https://git.drupalcode.org/project/drupal/-/blob/11.x/core/tests/Drupal/FunctionalTests/Update/UpdatePathTestBase.php?ref_type=heads)
- Need to determine what tests we will test, talk to the team about the approach

In our CI, we have one set of jobs that are supposed to test upgrades. However, the PHPUnit tests are no longer valid upgrade tests; we have switched to using BrowserTestBase which does a clean install for every test, so there is no way a test would fail on upgrade but not on install. Previously, we used Drupal Test Traits to test against an existing instance, so we could get the code + db in the correct state to simulate an upgrade.

Let's get something in place to restore this, as we've had a few recent bugs or regressions introduced that proper upgrade tests would have caught. We don't really want to re-add DrupalTestTraits as a dependency but could possibly use the core BuiltTest framework, or test Drush and API output via some other test framework outside of Drupal PHPUnit tests (something bash-based perhaps, or kernel-based tests that use a pre-build sqlite fixutre... there could be many approaches).

dafeder:
Adds upgrade tests for the last few hook_updates in metastore and datastore. Already found some issues with datastore update hooks (curly braces around table names not properly escaped) which have been fixed as well.

This will serve as a template for future update path tests, especially as we do some [module renaming](https://github.com/GetDKAN/dkan/issues/3746).

It might be a good idea to use [git lfs](https://git-lfs.com/) to manage the db dumps going forward, but for now I have not made that switch.


modules/datastore/tests/src/Functional/DatastoreUpdatePathTest.php
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setDatabaseDumpFiles() {
+    $this->databaseDumpFiles = [
+      __DIR__ . '/../../../../../tests/fixtures/update/update-2.20.0.php.gz',
+    ];
+  }
+
```
kaise-lafrai, line 25: Test is self explanatory but a comment would be nice here so we know what the test is doing.

modules/metastore/tests/src/Functional/MetastoreUpdatePathTest.php
```diff
@@ -0,0 +1,38 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  protected function setDatabaseDumpFiles() {
+    $this->databaseDumpFiles = [
+      __DIR__ . '/../../../../../tests/fixtures/update/update-2.20.0.php.gz',
+    ];
+  }
+
+  public function testUpdates8010on(): void {
```
kaise-lafrai, line 29: Test is self explanatory but a comment would be nice here so we know what the test is doing.

kaise-lafrai (commented):
Looks good to me! I had a small suggestion but it's not critical.

kaise-lafrai:
Took me a little longer to review this as UpdatePathTestBase.php is something I have not seen in a while and I was tinkering around with it locally. Thank you.

kaise-lafrai:
Will wait to merge once pipeline issues are resolved.

kaise-lafrai (approved):

PR #4463, opened 2025-04-08
Managed remote files

### LINKED ISSUE #4461: Use managed files correctly for both uploaded and remote on JSON form
When trying to make #4335 work, we ran into a _lot_ of problems with the file fields. This is because the custom widget we implemented on top of ManagedFile - UploadOrLink - does some kind of weird stuff to change values after the form build and doesn't work with with any ajax-y stuff modifying the form after initial build. This got me to thinking, do we really need to change the ManagedFile widget so much? What if we made a managed file entity for remote files as well, then just let managed file handle the values as it would normally, as opposed to letting it create managed files for uploaded files but bypassing all that for remote files?

dafeder:
Fixes #4461 

## Describe your changes

This is a significant rewrite of the UploadOrLink element in the JSON Form Widget. It works more in harmony with the MangedFile widget that it overwrites. Whereas previously we'd been overwriting the widget's behavior in more ways, this new approach creates a managed file for any value - whether an uploaded or remote file - and uses MangagedFile's own display/theming stuff to show the file once created. This hopefully makes the code easier to work with and more importantly will make the values play nice with #4335. 

## Known issues

* The codeclimate error - I don't see a way to make that one method better that's as readable. Open to suggestions. 
* I don't like the additional submit function that's added to the form submit button in the module form alter hook. That feels very hacky and I'm hoping we can find a better way to add that file usage after the form is submitted. That said, the form_alter was there before this PR.
* In the claro theme, you get a "( )" after the filename, as the template expects a filesize variable and doesn't check to see if it exists before adding the parenthesis. Not sure if we should override this in the theming system or make more of an attempt to get the file size from a remote file and save that to the file entity.

## QA Steps

1. Create a new dataset. In the distribution section, upload a file. Save the dataset and edit it again, confirm the file still appears as expected.
2. Add a distribution. Instead of uploading, link to a remote file like [this one](https://download.cms.gov/openpayments/PGYR2023_P01302025_01212025/OP_DTL_GNRL_PGYR2023_P01302025_01212025.csv).
3. Save, re-edit, confirm the file is there correctly
4. Remove the file, save and re-open, confirm correct behavior.
5. Otherwise try to beat up on it/try to break it

### DIFF
--- a/modules/json_form_widget/json_form_widget.module
+++ b/modules/json_form_widget/json_form_widget.module
@@ -12,7 +12,7 @@ use Drupal\file\Entity\File;
 use Drupal\file\FileInterface;
 use Drupal\json_form_widget\Plugin\Field\FieldWidget\JsonFormWidget;
 use Drupal\json_form_widget\ArrayHelper;
-
+use Drupal\json_form_widget\Element\UploadOrLink;
 
 /**
  * Submit handler for the "add-one-more" button.
@@ -35,7 +35,7 @@ function json_form_widget_remove_one(array &$form, FormStateInterface $form_stat
 /**
  * Update count property by the given offset.
  *
- * @param FormStateInterface $form_state
+ * @param \Drupal\Core\Form\FormStateInterface $form_state
  *   Form state.
  * @param int $offset
  *   Offset to change count by.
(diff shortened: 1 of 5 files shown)


kaise-lafrai (commented):
First issue encountered was that after uploading a local or remote file and saving, clicking the file opens a new tab and tries to load the file in the browser which doesn't work or takes too long to process, previously it would download the file. I think we want to continue having the file download.

modules/json_form_widget/src/Util/FileHelper.php
kaise-lafrai, line 1: This file is empty. Did we forget to commit?

kaise-lafrai:
> First issue encountered was that after uploading a local or remote file and saving, clicking the file opens a new tab and tries to load the file in the browser which doesn't work or takes too long to process, previously it would download the file. I think we want to continue having the file download.

** Correction: The file does download, but it tries to open the file in a new tab. The new tab is just blank. I can go under downloads and select the file and it then opens up in my prefered app. Maybe it's just me but worth others trying this as well to see results.

modules/json_form_widget/json_form_widget.module
```diff
@@ -188,4 +119,30 @@ function json_form_widget_entity_delete(EntityInterface $entity) {
...
+  foreach (array_keys($usage) as $type) {
+    if ($type == 'json_form_widget') {
+      $jfw_usage = TRUE;
+      break;
+    }
+  }
+  // If it does, and the URI does not start with public://, show the full URL as
+  // the link title. Otherwise, show the file name.
+  if ($jfw_usage && (strpos($file->getFileUri(), 'public://') !== 0) && isset($variables['link'])) {
+    unset($variables['file_size']);
```
kaise-lafrai, line 145: I think we need to remove more than just the file_size. I'm noticing that the parentheses are still visible. See img

image

kaise-lafrai:
This is great work and looks like a huge improvement to before. I will need more time to test this and review the remainder of the code next week. Thanks!

modules/json_form_widget/json_form_widget.module
dafeder: Yeah I should have documented this as a known issue, I'll update the description.

This is Claro theme's fault, it provides a twig template for this widget that does not surround the parentheses with a conditional to see if the file size variable is set. Let's discuss what the best way to address would be.

dafeder:
> ** Correction: The file does download, but it tries to open the file in a new tab. The new tab is just blank. I can go under downloads and select the file and it then opens up in my prefered app. Maybe it's just me but worth others trying this as well to see results.

Yeah this is weird.. I thought perhaps it was default behavior of the ManagedFile widget but when I add a new file field to an arbitrary content type it doesn't do this. I can't figure out where the target=_blank attribute is coming from. Will take another look later or see if anyone else has ideas.

modules/json_form_widget/src/Util/FileHelper.php
dafeder: Will delete

kaise-lafrai:
I also noticed that validation does not stop a user from adding a distribution without a resource and saving the node.

kaise-lafrai (changes requested):
Looks good to me Dan. There was a lot to think about for this PR and I think you did a great job with the changes. Some small changes as we discussed earlier.

dafeder:
@kaise-lafrai upon closer examination the new window behavior is exactly the same as in a normal file field added to a node form. I think we should just leave as is, its not really a problem?

modules/json_form_widget/json_form_widget.module
dafeder: There is a thread about fixing this in Claro, although it seems to be pretty slow-moving. https://www.drupal.org/project/drupal/issues/3117430#comment-16067172

It doesn't bother me that much, but we might consider making an issue to attempt to get the size for remote files with a HEAD request or something.

kaise-lafrai:
> I also noticed that validation does not stop a user from adding a distribution without a resource and saving the node.

Steps to reproduce:

- Navigate to an existing or new dataset, Select Add one button in the Distribution section.
- Enter title, descr, select Upload Data File
- Do not select a file and just save the dataset. 

You will be routed to the /content page and see the following error:

`Warning: Undefined array key "#uri" in Drupal\json_form_widget\Element\UploadOrLink::getLocalFileUrl() (line 315 of modules/contrib/dkan/modules/json_form_widget/src/Element/UploadOrLink.php).`

dafeder:
@kaise-lafrai I believe this fixed it, please confirm

modules/json_form_widget/json_form_widget.module
kaise-lafrai: Not a bad idea to request the file size from the HEAD.

kaise-lafrai (approved):
Thank you Dan. Looks good to me.

janette (approved):
LGTM

dafeder:
Reviewing with @janette we discovered a weird edge case - if you provide a URL to another Drupal site, that has /sites/default/files in the path to the resource file, the URI generation logic will break.

If you give UploadOrLink::getFileUri() "https//example.com/sites/default/files/test1.csv" you would expect to get the same thing back, but you will get "https//example.compublic://test1.csv", and the form validation will fail due to invalid URL (which is fortunate, or we might not have caught).

I just pushed a test that fails due to this problem, next commit should fix w/o modifying the test.

PR #4472, opened 2025-04-21
Move and remove elements from JSON arrays in form

### LINKED ISSUE #4335: Remove or Reorder Objects in An Array

dafeder:
Fixes #4335 

## Describe your changes

Fairly extensive changes to how array properties are handled in JSON Form Widget. Arrays, whether "simple" (an array of strings or numbers) or "complex" (arrays of objects) will contain one more level of nested fieldsets. This is because each array item now has its own action buttons, so that array items can be removed or moved in the order. Previously we only had action buttons at the end of the array, allowing us to "add one" or "remove one".

Because the autoselect widgets expect a built array fieldset and then rewrite it, there are also a number of changes to that logic. A risk - we do not really have full functional test coverage for these elements. Also, now that I understand the code flow a bit better, it does make me wonder if there's a better approach that avoids building all these nested fieldsets on the first pass only to flatten and replace them on the second pass with WidgetRouter.

Functional tests for json_field_widget have also been reorganized to reduce duplicate code.

## Known issues

1. The default behavior for managed files is to show only the filename, not the full URL. We override that if the file has usage linking it to json_form_widget (see #4463) but that doesn't apply when on a new dataset form. So if you are in node/add/data and add a remote file, then use the move up or move down button, you will see it tranformed into a managed file link with just the filename. I'm actually not sure if showing the full URL _is_ what we want so not sure whether to try to find another solution to override the theming, or whether to remove the override and just always show the filename only.
2. This does make the form a lot busier. Especially for arrays of single text fields the fields are kind of overwhelmed by all the new buttons. So there is some UX tradeoff.

## QA Steps

- [ ] Create a new Dataset
- [ ] Add several distributions
- [ ] Re-order them, remove distributions from the beginning or middle of the list
- [ ] Do the same for "References"
- [ ] Confirm that the keyword and topic fields continue to work as expected
- [ ] Save the dataset, and view the JSON in the API to confirm all works as expected
- [ ] Edit an existing dataset, and try similar tests

You may want to remove some fields from the distribution schema to make that form element shorter - Keep it to just title, description and downloadURL for instance.

## Note on codeclimate failure

WidgetRouter is pretty overloaded, I have another ticket to break it up into multiple classes #4473. 
ArrayHelper is also just inherently a complex class, and gained some functions by moving callbacks from the .module file to the class, which I think is a win for readability/maintainability. I could maybe eliminate some of those very short helper functions or move them to a Util class but I don't think that would make it more maintainable. Could revisit.


dafeder:
More tests needed

dafeder:
Current issues:

1. If you edit an existing dataset, remove the last distribution, then add one (all without saving), it will be populated with the data from the deleted distribution.
2. If you edit an existing dataset, remove a distribution that's not the last one, then add one, the new distribution will be populated with the same data as the previous last dataset. For instance, if a dataset has distributions A B and C, delete B and add another, and you will have A, C and a copy of C.
3. If you add a new distribution with a new remote file, then move it up or down, you will lose the remote file (reordering distros in existing datasets that already point to managed files is fine).

I think I'd be comfortable merging this with 3 as a known issue and maybe 1 but not 2. One and two have the same cause: when a stored value for distro 3 exists and there is no user input to override it, it uses the stored values.  I'm going to add tests for them that will fail and then work on getting them to pass.

dafeder:
I believe I've fixed all these issue.

janette:
@dafeder When editing a distribution with an existing remote url, change that distribution to point to a new remote url, and save, the downloadURL value is not saved.

dafeder:
Thanks @janette - I believe I've fixed that now. Was actually a bug related to #4463 and not specific to the reordering. However, also saw a similar issue where removing a URL, adding a new one and then reordering w/o saving didn't work. I believe I've fixed both issues, added tests.

janette:
Testing Results:
- ✅ Updating a remote downloadURL works as expected
- ✅ Re-ordering pre-existing distributions or related document items
- ❌  Re-ordering **new** distribution or related document items will cause the entire element to disappear from the form
-  ❌ Theme and Keyword elements display an unneeded "Add one" button (as well as any other custom property that can contain multiple values)

dafeder:
OK @janette I found the two bugs I introduced right before pushing and believe they're both fixed, including quite a few more tests (and refactoring of the ReorderTest to make it more readable, hopefully). Can you re-test please?

dafeder:
Regarding the low diff coverage, that should be addressed by changes in #4486.

modules/json_form_widget/src/Element/UploadOrLink.php
```diff
@@ -66,24 +66,27 @@ public static function valueCallback(&$element, $input, FormStateInterface $form
     // Detect whether the remove button was clicked.
     $remove = FALSE;
     $file_remove = $form_state->get('file_remove') ?? [];
+    // Make sure the element set for remove is the same as current.
```
janette: ...set for **removal.**..

modules/json_form_widget/src/Element/UploadOrLink.php
```diff
@@ -66,24 +66,27 @@ public static function valueCallback(&$element, $input, FormStateInterface $form
...
+      $file = static::getManagedFile(static::getFileUri($uri));
       // If remove was clicked, we need to unset the uri. If not, we need to add
       // the fid to the input array.
       if ($remove) {
         $element['#uri'] = '';
+        $input['file_url_remote'] = '';
       }
       else {
-        // Add filet to input array and update the entity.
+        // Add fil ID to input array and update the entity.
```
janette: file ID

janette (commented):
@dafeder this is working really well, the "add one" buttons are no longer showing up on other elements. Seems like you are only excluding the theme and keyword elements in the code but the other custom fields are also fixed. Moving things anywhere at anytime also works great. Overall the code looks good just two possible typos in comments.

dafeder:
> Seems like you are only excluding the theme and keyword elements in the code but the other custom fields are also fixed.

Not sure where you see theme and keyword specified in the code? The fix for that problem with the extra "add" button was here:

https://github.com/GetDKAN/dkan/pull/4472/commits/4024f3a353f132f0597ae5882e3f432c4b50861b#diff-531b9da6698423939724567eef6afa0957c1f11ca3cde6a243dbb43fd1a2de71L199

I had renamed the parent element to "array_actions" to differentiate a bit but forgot to update in a key place.

janette (approved):

stefan-korn:
@dafeder : The issue with subschema mentioned in [4335](https://github.com/GetDKAN/dkan/issues/4335#issuecomment-2684303856) still persists as far as my tests go. The issue is not prevalent with the current default schema in DKAN, because distribution does not use multi value fields. But in theory it should be possible to use multi value fields in subschema too I guess (was at least possible before).

This is the issue mentioned:
> does not work correct on subschema, i. e. you have a multivalue field in distribution and either edit this field in the dataset form or in the distribution form. The actions buttons won't work (mismatch of ajax target wrapper)

stefan-korn:
@dafeder : Using changes in #4493 I could make the "add one" button work for subschema too (in own edit form and inside dataset form like for distribution), while not breaking the functionality for dataset hopefully.

But still the remove/reorder does not work on subschema.

dafeder:
Interesting. Tbh we never edit subschema items directly and I didn't think to test for it. Can you file a bug?

stefan-korn:
Okay, will file a bug. But it is not happening only for direct edit for subschema, but also for edit subschema inside dataset, like distribution.

If the distribution has a multivalue field like this:

```
    "text_prop": {
      "title": "Text property example",
      "description": "Multivalue text.",
      "type": "array",
      "items": {
        "type": "string",
      },
    },
```

Distribution schema does currently not use such a field, therefore this problem is not visible.

It's something about the nesting inside the json form widget, which is different for subschema than dataset (ultimately probably coming from "properties" for dataset schema and "properties->data" for subschema).

One thing I came across is data dictionary. The fields part there seems to be solved differently altogether (Add field, then Add/Cancel)? Is this special kind of widget?

dafeder:
Ah I see what you mean. Yes, we created a special widget specifically for data dictionary, as it is not meant to have a customizable schema. Did you have a field like this already in distribution, IE is this a regression?

stefan-korn:
Okay, I see there is a separate module for [data dictionary](https://github.com/GetDKAN/dkan/tree/2.x/modules/data_dictionary_widget) and the widget is injected via [hook_entity_form_display_alter](https://github.com/GetDKAN/dkan/blob/2.x/modules/data_dictionary_widget/data_dictionary_widget.module#L36) - interesting approach, might use this in some other place too.

But we do not use the data dictionary actively at the moment, I was only wondering why that widget behaves so differently. The issue for us is to use schema properties of "type: array" in distribution subschema, which does not play nicely with the add/move/remove actions.

stefan-korn:
And out of curiosity, you said

> Yes, we created a special widget specifically for data dictionary, as it is not meant to have a customizable schema ...

But you can customize the Dictionary schema via data-dictionary.json and this will work as far as I can see?

dafeder:
@stefan-korn sorry I got pulled into other work and lost the thread here.

Correct, this is sort of a weak spot, technically data-dictionary is customizable like all the schemas, and in fact must be copied to a custom schema folder if you're customizing any other schema, which can cause it to diverge from the schema in core (though we do avoid making any changes to it).

I think this will be better if/when we complete #3761 which would mean you could among other things override certain schemas but fall back to core for others.

stefan-korn:
@dafeder : Oh wow, did not know #3761 until now. This reads very interesting, but also a bit scary while thinking about the customization we already did ourselves in terms of DCAT-AP.de schema ...

dafeder:
> @dafeder : Oh wow, did not know #3761 until now. This reads very interesting, but also a bit scary while thinking about the customization we already did ourselves in terms of DCAT-AP.de schema ...

Don't worry, again it's not 100% we will do this at all (although I very much hope so) but if we do I definitely plan to make it opt-in, where if you have a custom schema/collections directory it will fall back to the old method instead of looking for them in modules and yml files. I also think moving to the new system would be a fairly light lift and perhaps even automate-able. (Also, see the comments in the ticket where I walk back some of the more complex parts of the proposal)

PR #4474, opened 2025-04-23
#4469: Pass Drupal Http Client to Harvester

### LINKED ISSUE #4469: Harvester does not use proxy settings
### Current Behavior

DKAN harvester does not use proxy settings (from. $settings['http_client_config']['proxy']...).

I suppose the reason is, that [HarvestService does not provide a http client](https://github.com/GetDKAN/dkan/blob/2.x/modules/harvest/src/HarvestService.php#L431) to the [Factory](https://github.com/GetDKAN/harvest/blob/master/src/ETL/Factory.php) and thus the harvest is [using a new GuzzleClient](https://github.com/GetDKAN/harvest/blob/master/src/ETL/Extract/DataJson.php#L27) that does not take proxy settings.

### Expected Behavior

Harvester use proxy settings (from. $settings['http_client_config']['proxy']...).

### Steps To Reproduce

Define proxy settings in your environment
Configure a harvest that needs to run with proxy settings
Run the harvest and notice it fails

```shell

```

### Anything else?

_No response_

stefan-korn:
Fixes #4469 

## Describe your changes

Just passing the Drupal HTTP Client to the Harvester. If you have suggestion, like do it with dependency injection in the harvester service, let me know.

## QA Steps

### DIFF
--- a/modules/harvest/src/HarvestService.php
+++ b/modules/harvest/src/HarvestService.php
@@ -428,7 +428,7 @@ class HarvestService implements ContainerInjectionInterface {
    * Get the harvester from the harvester library.
    */
   protected function getDkanHarvesterInstance($harvestPlan, $item_store, $hash_store): Harvester {
-    return new Harvester(new Factory($harvestPlan, $item_store, $hash_store));
+    return new Harvester(new Factory($harvestPlan, $item_store, $hash_store, \Drupal::httpClient()));
   }
 
 }


stefan-korn:
okay, codeclimate says it all ;-)

> \Drupal calls should be avoided in classes, use dependency injection instead

Will take a look into this.

dafeder:
@stefan-korn yes we would need to add http client as a constructor for the service. Codeclimate will surely complain that it's too many arguments but that one we can ignore; I think the solution there is just to split this into two services or something because as it is now they're all needed.

Also BTW I don't know why we have "getHarvester" and "getDkanHarvesterInstance" they seem totally redundant. I'll probably squash those into one method at some point.

dafeder:
@stefan-korn do you plan to fix this? Thanks

dafeder:
@stefan-korn bump :)

Make sure to rebase this if you pick it back up, as you need to switch to the new qlty tooling

Issue #4465, opened 2025-04-10
Move JsonFormWidget plugin into DKAN metastore module


dafeder:
Now that the DKAN-specific stuff is in the form widget plugin, move this plugin from the json_form_widget module into DKAN itself.

Issue #4413, opened 2025-02-19
Combine dkan and metastore modules


dafeder:
Continuing the work laid out in #3746, we do not need separate a module for "metastore". There is no DKAN w/o the metastore. One #4412 is complete, merge "metastore" into "dkan" as well.

## Acceptance criteria

* Metastore services can maintain the same name. It is fine for them to be prefixed with `dkan.metastore`. 
* As appropriate, move Drupal\metastore namespaces into Drupal\dkan root or Drupal\dkan\Metastore. More discussion needed to determine what goes where here.
* Of course, upgrade path. Table names should not change but there will likely be some update hook magic needed for this to be smooth. Documentation for custom code integrations refactor a must.

More discussion needed to finalize sequence of merging/release here. Should this and #4412 merge into a 3.x branch that we maintain in parallel? Can this be released under 2.x as it doesn't technically add or rename any modules?

PR #4480, opened 2025-04-25
Merge common into dkan module

### LINKED ISSUE #4412: Combine common and dkan modules
#4360 will need to be completed before we can start this work

As a first step in #3746, let's try to move everything in "common" up one level so that we just have a "dkan" module and no "common" module.

## Acceptance criteria

* Services that have `dkan.common` as a prefix should now only have `dkan`. We can keep the old service names as well for one release cycle but note them as deprecated and internally switch to the new service names.
* No tests should need to be removed, added or significantly changed, just moved into the new appropriate folder and namespaces updated.
* Upgrade path. We must be sure that sites without any custom modules using DKAN services/classes upgrade without any extra steps, and that clear documentation is provided to refactor existing custom code to work with new namespaces and service names.

TBD:

* Can this be released as a minor version release under DKAN 2.x?
* Does it necessitate a 3.x release? Perhaps an alpha/pre-release? Or should it be kept in a 3.x branch until all of #3746 is complete?

dafeder:
Fixes #4412


dafeder:
I think we're blocked by #4465. We get circular dependencies between metastore and json-form-widget.

Also, not sure it will be feasible to do this and not #4413 at the same time. Seems inevitable we have circular dependencies.

dafeder:
I take it back i don't think we're blocked by #4465 exactly, but to get around we need a one way dependency from json_form_widget on metastore, when what we actually want is the opposite.

dafeder:
I had to update our db dump. In its current state, it has post import queue items still in the queue, which break because they point to old classes.

modules/metastore/metastore.install
```diff
@@ -161,5 +161,5 @@ function metastore_update_8010() {
     $config->set('redirect_to_datasets', TRUE)->save();
   }
 
-  drupal_flush_all_caches();
+  return t('Set the default value for the redirect_to_datasets setting to TRUE.');
```
dafeder, line 164: Flush all caches not necessary, and no return messes up the update tests apparently.

PR #4481, opened 2025-04-28
Rearrange harvest unit tests

dafeder:
The harvest unit tests are organized in a bit of a nonstandard way now due to the absorption of the harvest PHP library in #4373. This PR simply moves the tests and fixture files into some more expected places, and adds a bit of missing metadata.

- Tests in tests/src/ETL are now inside tests/src/Unit
- Fixture files from tests/src/ETL moved up to tests/files
- Some test classes that were in tests/src/ETL are now simply in tests/src (putting them inside Unit prevents them from being autoloaded)


modules/harvest/tests/src/TestExtract.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Drupal\Tests\harvest;
+
+use Drupal\harvest\ETL\Extract\Extract;
+
+/**
```
paul-m, line 7: Not a test, so remove @group and @covers

modules/harvest/tests/src/TestExtractNoObjects.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Drupal\Tests\harvest;
+
+use Drupal\harvest\ETL\Extract\Extract;
+
+/**
+ * @covers \Drupal\harvest\ETL\Extract\Extract
```
paul-m: Not a test so remove @group and @covers

modules/harvest/tests/src/Unit/ETL/Load/LoadTest.php
```diff
@@ -12,7 +12,9 @@
  * @covers \Drupal\harvest\ETL\Load\Load
  * @coversDefaultClass \Drupal\harvest\ETL\Load\Load
  *
- * @group harvest
+ * @group harvest\
```
paul-m: Extra slash...

paul-m (changes requested):

paul-m (approved):

PR #4485, opened 2025-05-06
Fix the dictionaryEnforcer by passing both the resource ID and version

janette:
Fixes 24997

## The Issue
When in reference mode, the DictionaryEnforcer::returnDataDictionaryFields method is only passing the resource ID without the version, this causes DataDictionaryDiscovery::getReferenceDictionaryId to return the wrong distribution, which can lead to a json schema validation error, which in turn breaks the csv download of filtered datastore results.

## Describe your changes
This change will pass both the resouce ID and the version so that the correct distribution is used when in the data dictionary reference mode. 

## QA steps

1. To test on a down-stream build refer to testing steps in 25439
2. Otherwise
3. Create a dataset with a distribution without a downloadURL
4. Edit the dataset and update the distribution to include a resource file
5. Run cron twice to get a datastore table
6. Visit the csv download url with a query parameter (e.g. https://dkan.ddev.site/api/1/datastore/query/{{datasetUUID}}/0/download?conditions%5B0%5D%5Bproperty%5D=state&conditions%5B0%5D%5Bvalue%5D=wi&conditions%5B0%5D%5Boperator%5D=%3D&format=csv)
7. Confirm you get a csv file with the results you expect and not the "JSON Schema validation failed" error.

### DIFF
--- a/modules/datastore/src/Service/ResourceProcessor/DictionaryEnforcer.php
+++ b/modules/datastore/src/Service/ResourceProcessor/DictionaryEnforcer.php
@@ -155,7 +155,7 @@ class DictionaryEnforcer implements ResourceProcessorInterface {
 
       case "reference":
         $resource = DataResource::getIdentifierAndVersion($identifier);
-        $dictionary_id = $this->dataDictionaryDiscovery->dictionaryIdFromResource($resource[0]);
+        $dictionary_id = $this->dataDictionaryDiscovery->dictionaryIdFromResource($resource[0], $resource[1]);
         break;
 
       default:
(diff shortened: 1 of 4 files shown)


rovcase:
Just noting that the "patch" version of this update is working as intended on my local for medicaid. Will leave this up to the DKAN core team for review/merge/release.

Great detective work!

janette:
Rather than passing a resource ID without a version, throw an exception. Add/fix tests.

chri5tia:
Instructions on https://jira.cms.gov/browse/WCMS-25439 don't seem to work for this, perhaps they are legacy commands. Can you update the testing instructions for this PR?

chri5tia (approved):
Worked as expected in local functional testing.

PR #3189, opened 2020-08-25
Run tests in parallel

dafeder:
* Executes tests in three parallel jobs, reducing total CI time by ~50%
* Provides more detailed reporting from the github side (can see which of three jobs failed before going to CircleCI
* Also includes DKAN from a local /dkan dir instead of bringing in through packagist

Fixes #3178

PR #4487, opened 2025-05-12
Split functional tests

### LINKED ISSUE #4448: Parallelize PHPUnit functional tests
Our test builds have really gotten slow as we've added more and more BrowserTestBase tests. We did have an attempt at parallelizing them in #4119 but it didn't quite work. It seems unlikely that we can make automatic test splitting work for phpunit in ddev based on previous attempts, but we can manually split up the tests into groups and run specific groups in parallel.

Requirements:

1. We have some way to check if there are any functional tests missing a group assignment, or have a default pseudo-group that tests get added to by default if someone forgets to add a group.
2. We can still re-combine test output so we get accurate coverage reports.
3. Our build time improves enough to justify the added complexity

dafeder:
Fixes #4448 

Redux of #4119 #3283 #3189 #3185

Requires functional tests to have a group of functional[1-3]. We will need to rebalance the parallel jobs ourselves if we see a big difference start to appear, or add a new group and bump up the parallelism setting. But right now it has the build time **down to ~15min from ~34min for the target job** and ~8min for the rest, which is probably about as fast as is possible. Perhaps we could shave another minute or two off with another group.

## Notes 

* This temporarily eliminates the coverage HTML report, which was previously added to the artifacts for the target build. It was much too hard to re-combine the PHPUnit results and re-generate the HTML report. This will be easier to re-implement when we switch the target to Drupal 11, as the main issue had to do with the [phpcov](https://github.com/sebastianbergmann/phpcov) tool only running on php 7 for phpunit 9 (D11 uses PHPUnit 10).
* I also eliminated the last PHPUnit build in the upgrade job. Not related to splitting exactly but as we are using 4x compute resources now for the PHPUnit stuff, I wanted to cut back where we could. As we've discussed, now that we've switched functional tests to BrowserTestBase, the upgrade job is not actually testing anything.
* Usage is still quite high now. We'll have to keep our eye on our CircleCI compute usage and see if it settles down or if we need to optimize to keep our bills (and to some extent I guess, our carbon footprint) under control
* See the penultimate commit, where I added code without coverage to make sure codeclimate was correctly receiving coverage data.


dafeder:
Tests are passing and fairly evenly split, now need to figure out how to recombine the coverage report.

dafeder:
So close. Issue is now that the .cov files refer to paths beginning with /var/www/html but now they're in /home/circleci/project because we're not in ddev in the last job anymore. So far a simple find and replace in the .cov files is not working; may have to include a small php script to load the cov files, iterate through them and re-write them (cov files are just php files containing a single giant serialized php object).

dafeder:
Note that adding a bunch of functions with no tests results in 0% diff coverage

image

modules/common/tests/src/Kernel/Storage/AbstractJobStoreFactoryTest.php
```diff
@@ -16,7 +16,7 @@
    * @group dkan
    * @group common
    * @group kernel
-   *
+      *
    * @see \Drupal\Tests\common\Kernel\Storage\JobStoreFactoryTest
```
janette: The new indentations on these three JobStoreFactoryTest files seem off

janette (approved):

PR #4492, opened 2025-05-19
Restore HTML report artifact

dafeder:
Preserving some of the work on recombining for HTML report, since we're dropping that temporarily.


dafeder:
This PR is too stale to use at this point, and we get some of this functionality from qlty.sh now. Closing.

PR #4494, opened 2025-05-20
4472 regression subschema hideactions

stefan-korn:
just for internal reference

## QA Steps


dafeder:
@stefan-korn not sure the background on this but if you wanted it supported in core, I think it could be implemented as an optional property for UI schemas, and the logic would probably move to WidgetRouter

PR #4501, opened 2025-06-17
Plan info in dkan:harvest:info

dafeder:
There is currently no way to get a harvest plan in Drush without doing a database query. It would seem logical that `drush dkan:harvest:info [plan-id]` would show you _plan_ information, but according to the docs is more for getting _run_ information. While it might make sense to have a separate plan for each, maybe we can just have this command do both for now?

This change simply adds the plan JSON to the top of the command's output. I also changed some docs to use the word "plan" more consistently, rather than the more ambiguous "harvest" (is a "harvest" a _source,_ a _plan_ or a _run_?).

It does not address the fact that the existing functionality of the command - to list all past runs for a plan, or just for a specific run if a second option is passed - is not quite working. 

Also, there were no existing tests for these commands and the command file is excluded from test coverage, so I didn't add any.

### DIFF
--- a/modules/harvest/src/Commands/HarvestCommands.php
+++ b/modules/harvest/src/Commands/HarvestCommands.php
@@ -41,7 +41,7 @@ class HarvestCommands extends DrushCommands {
   }
 
   /**
-   * List available harvests.
+   * List available harvest plans.
    *
    * @command dkan:harvest:list
    *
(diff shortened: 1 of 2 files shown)


kaise-lafrai (approved):
Putting the harvest plan info on top of the run info is a bit odd. It would be nice if it were in it's own table above the run info table. We can iterate/improve on this later.

dafeder:
I think if anything the run and plan info should be in separate commands, which are more explicit in their naming? It's hard to show the plan info as a table, and there are times when you want to just copy the JSON to paste somewhere else. But we can think about other formats.

PR #4509, opened 2025-07-03
Allow null on addDistributionInfo in DatastoreInfo

janette:
Fixes
```
TypeError: Drupal\datastore\Service\Info\ImportInfo::getItem(): Argument #1 ($identifier) must be of type string, null given, called in /var/www/html/docroot/modules/contrib/dkan/modules/datastore/src/Plugin/DatasetInfo/DatastoreInfo.php on line 128 in Drupal\datastore\Service\Info\ImportInfo->getItem() (line 86 of modules/contrib/dkan/modules/datastore/src/Service/Info/ImportInfo.php). 
```

## Describe your changes
Allow the import dashboard to load even if datasets have no distributions.

## QA Steps

- [ ] Create a dataset without a distribution.
- [ ] Visit /admin/dkan/datastore/status
- [ ] Confirm the page loads and you do not see the error above

### DIFF
--- a/modules/datastore/src/Plugin/DatasetInfo/DatastoreInfo.php
+++ b/modules/datastore/src/Plugin/DatasetInfo/DatastoreInfo.php
@@ -123,8 +123,8 @@ class DatastoreInfo extends DatasetInfoPluginBase {
    *   The distribution info from a dataset info array.
    */
   protected function addDistributionInfo(array &$distribution): void {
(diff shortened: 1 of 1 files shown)


dafeder (approved):
Technically I think we should have a test for the NULL case but it's kind of a mystery how to get there. What we know is that this does fix a real-world bug and doesn't introduce any side-effects that we can see, so I'm going to approve.

PR #4513, opened 2025-07-10
Migrate from codelimate to qlty

dafeder:
Another attempt to replace codelimate with qlty, now that codeclimate is at EOL.

See the checks at #4517 to see the checks in action.


kaise-lafrai (approved):
Changes look good to me as far as I can tell. Thanks!

PR #4520, opened 2025-07-22
Split dashboard title search terms into multiple conditions in query to get accurate results

### LINKED ISSUE #4515: 500 error on Dataset Title search on import dashboard
Steps to reproduce:

Log in and set up sample content
Visit /admin/dkan/datastore/status
In the Dataset Title field, type "Florida Lanes" and click filter

`TypeError: Drupal\datastore\Form\DashboardForm::getDatasetsByTitle(): Return value must be of type array, null returned in Drupal\datastore\Form\DashboardForm->getDatasetsByTitle() (line 340 of /bitnami/drupal/modules/contrib/dkan/modules/datastore/src/Form/DashboardForm.php).`

kaise-lafrai:
Fixes #4515 

- I set an empty array variable so that we no longer get the 500 error "Return value must be of type array, null returned".
- I refactored the getDatasetsByTitle method to split the search terms and add each word as a separate search condition. This returns accurate results. For example: search term "Florida Bike" returns a dataset, search term "Florida Lanes" does not. With the changes introduces, results for "Florida Lanes" will appear in the dashboard.

## QA Steps

- Run ddev drush en sample_content
- Run ddev drush dkan:sample-content:create
- Run ddev drush cron a couple times
- Visit /admin/dkan/datastore/status
- In the Dataset Title field, type "Florida Lanes" and click filter
- 1 result should appear

### DIFF
--- a/modules/datastore/src/Form/DashboardForm.php
+++ b/modules/datastore/src/Form/DashboardForm.php
@@ -333,9 +333,9 @@ class DashboardForm extends FormBase {
       ->accessCheck(FALSE)
       ->condition('type', 'data')
       ->condition('field_data_type', 'dataset');
-    
+
     $searchTerms = array_filter(explode(' ', trim($filters['dataset_title'])));
-    
+
     if (!empty($searchTerms)) {
       $titleGroup = $query->andConditionGroup();
       foreach ($searchTerms as $term) {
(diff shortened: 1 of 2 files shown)


janette:
@kaise-lafrai the code looks good, passes manual testing, also works when applied to client site, not sure why the phpunit tests are failing on `Error: unknown flag: --exclude-group`

janette (approved):

PR #4522, opened 2025-07-23
Remove flaky test

### LINKED ISSUE #4497: Fix or remove ImportJobTest::testMultiplePasses()
`Drupal\Tests\datastore\Unit\Plugin\QueueWorker\ImportJobTest::testMultiplePasses()` is a flaky test.

We should make it stable.

It is currently marked as skipped. Unskip the test and fix it so it works.

Alternately, remove it.

Currently, the test wants multiple import jobs to be run during an import process. The issue is that the import doesn't always take more than one job in the queue. Other than the number of queue items required, the test passes fine.

This replaces 14754.

dafeder:
Fixes #4497

This test has been disabled for years. It hasn't seemed to be missed, and is constantly showing up as a warning at the bottom of pipeline output. I can't actually reproduce the "single pass" behavior -- the test passes for me locally -- but I am sure it happens sometimes or we wouldn't have skipped it. This seems like behavior that would be better covered in a kernel or functional test anyway (and surely is).

### DIFF
--- a/modules/datastore/tests/src/Unit/Plugin/QueueWorker/ImportJobTest.php
+++ b/modules/datastore/tests/src/Unit/Plugin/QueueWorker/ImportJobTest.php
@@ -240,51 +240,6 @@ class ImportJobTest extends TestCase {
     $this->assertEquals($c, $results[10001]);
   }
 
-  /**
-   * This is the same as testLargeImport but expects more than one pass.
-   */
-  public function testMultiplePasses() {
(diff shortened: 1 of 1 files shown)


paul-m (approved):
testMultiplePasses() duplicates coverage of testLargeImport(), but expects more than one pass. Since we don't seem to be able to simulate more than one pass, we are OK for now. There's some mitigation in the fact that most dkan sites seem to be using the LOAD DATA LOCAL importer, rather than the chunked one anyway.

Issue #4510, opened 2025-07-07
Stop using separate API permissions, follow content permissions.


janette:
It is redundant to have separate permissions for API and UI , replace the API permissions and rely on the Drupal content authorship permissions.
https://www.drupal.org/node/2733435

dafeder:
This won't be completely straightforward because the routes don't use the kind of patterns normal Drupal entity routes use. I'm pretty sure we'll need to implement [custom route access checking](https://www.drupal.org/docs/8/api/routing-system/access-checking-on-routes/custom-route-access-checking) -- which, is not actually that hard and has the advantage of being more flexible if we finally move to custom entities.

dafeder:
Actually maybe just handling access correctly in the Metastore NodeData storage and bubbling-up/handling the exceptions correctly in the controller can take care of this. Trying that approach first, and updating some of our tests to help with that.

dafeder:
Nope, NodeData uses EntityStorageInterface, which does not do any access checks. Need to implement access checks in the controller.

PR #4523, opened 2025-07-24
Test coverage for API delete, and for unauthorized request

dafeder:
Related to #4510

Adds missing coverage in the functional API tests for using the delete method in the API. Also cleans up a bit of code in the API tests, and adds a check for 403 if unauthorized user.

Note: Some additional items needed in the spec so that we can validate DELETE requests.

### DIFF
--- a/modules/metastore/tests/src/Functional/Api1/DatasetItemTest.php
+++ b/modules/metastore/tests/src/Functional/Api1/DatasetItemTest.php
@@ -53,6 +53,14 @@ class DatasetItemTest extends Api1TestBase {
     $response = $this->post($dataset, FALSE);
     $this->assertEquals(409, $response->getStatusCode());
     $this->validator->validate($response, $this->endpoint, 'post');
+
+    // Now an unauthorized user.
+    $response = $this->httpClient->post($this->endpoint, [
+      RequestOptions::JSON => $dataset,
+      RequestOptions::AUTH => $this->authNoPerms,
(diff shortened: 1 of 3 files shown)


modules/metastore/tests/src/Functional/Api1/DatasetItemTest.php
```diff
@@ -113,6 +128,40 @@ public function testPut() {
...
+    // @todo Add delete to the spec so we can validate it.
+    $this->assertEquals(200, $response->getStatusCode());
+
+    // Now try to get the deleted dataset.
+    $response = $this->httpClient->get("{$this->endpoint}/{$datasetId}", [
+      RequestOptions::HTTP_ERRORS => FALSE,
+    ]);
+    $this->assertEquals(404, $response->getStatusCode());
+
+    // Now an unauthorized user.
```
paul-m: Maybe try the no-auth user assertion while the dataset is still available, so try it first.

modules/metastore/tests/src/Functional/Api1/DatasetItemTest.php
```diff
@@ -53,6 +53,14 @@ public function testPost() {
     $response = $this->post($dataset, FALSE);
```
paul-m, line 54: Do me a favor and add `@group metastore` to the class docblock?

paul-m (changes requested):

paul-m (approved):

PR #4525, opened 2025-07-29
API delete and 404 error handling

dafeder:
This PR fixes two inconsistencies in the Metastore discovered while working on #4510:

1. MissingObjectException was throwing [HTTP 412](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/412) errors. I'm not sure where this came from but there's really no use-case in DKAN I can think of where this would be an appropriate error. I've changed it to the much-more-standard 404.
2. There was not really correct error handling going on between the Metastore service and storage classes for delete methods. The service expects a result of some kind from the storage, but it doesn't give one, as Entity::delete() is not specified to return anything. This has been changed to assume that deletion was successful if no exceptions were thrown.

Note that this will also change the behavior of the PATCH method for the metastore API, which could also make a 412 response if given an invalid ID. It will also now return a 404 in that case.

### DIFF
--- a/cypress/e2e/01_metastore.cy.js
+++ b/cypress/e2e/01_metastore.cy.js
@@ -226,7 +226,7 @@ context('Metastore', () => {
           body: dkan.generateMetastore(schema_id, identifier),
           failOnStatusCode: false
         }).then((response) => {
-          expect(response.status).eql(412)
+          expect(response.status).eql(404)
         })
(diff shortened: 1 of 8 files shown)


kaise-lafrai (commented):
Code looks good to me

paul-m (approved):

PR #4526, opened 2025-07-29
Use entity access to negotiate API permissions

### LINKED ISSUE #4510: Stop using separate API permissions, follow content permissions.

dafeder:
Fixes #4510 

Adds custom access check for Metastore routes.

The permission `post put delete datasets through the api` is deprecated. All API actions now require the same permissions needed to accomplish equivalent things in the Drupal UI. They will, for now, also respect the old "post put..." permission for backward compatibility, but we will remove this in a later release. Release notes should advise people to update their roles. 

The "API User" role has been updated to simply include the appropriate content access permissions. This role should probably be removed/replaced eventually as default install config.

## QA Steps

1. Create a one user with the following permissions:

* `access content`
* `create data content`
* `edit own data content`
* `delete own data content`
* `use dkan_publishing transition publish`
* `use dkan_publishing transition archive`
* `use dkan_publishing transition hidden`
* `use dkan_publishing transition restore`
* `view data revisions`
* `view any unpublished content`

2. Now create another user with just the `access content` permission
3. Try posting a new dataset as the unprivileged user. Confirm 403 response
4. Try posting a new dataset as the priviledged user. Confirm success
5. Try to PATCH and PUT against the new dataset as as the unprivileged user. Confirm 403 response
6. Try to PATCH and PUT against the new dataset as as the privileged user. Confirm success.
7. Try to DELETE the new dataset as the unprivileged user. Confirm 403 response.
8. Optional - temporarily remove the "delete own data content" permission from the privileged user, and confirm you can not DELETE. Restore the permission.
9. Try to DELETE the new dataset as the privileged user. Confirm success.
10. Try the same set of tests for changing publication status via the revisions endpoint.

### DIFF
--- a/modules/metastore/src/Controller/MetastoreAccessManager.php
+++ b/modules/metastore/src/Controller/MetastoreAccessManager.php
@@ -2,14 +2,23 @@
 
 namespace Drupal\metastore\Controller;
 
+use Drupal\Core\Access\AccessResult;
 use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
 use Drupal\Core\Entity\EntityAccessControlHandlerInterface;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Session\AccountInterface;
+use Drupal\metastore\Exception\MissingObjectException;
 use Drupal\metastore\Factory\MetastoreEntityItemFactoryInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 
+/**
+ * Manages access control for Metastore items in API endpoints.
+ *
+ * This class provides methods to check if a user can create, update, or delete
+ * items based on their schema ID and item ID.
+ */
 class MetastoreAccessManager implements ContainerInjectionInterface {
 
   /**
@@ -44,8 +53,10 @@ class MetastoreAccessManager implements ContainerInjectionInterface {
     EntityTypeManagerInterface $entityTypeManager,
     MetastoreEntityItemFactoryInterface $itemFactory,
   ) {
+    $this->itemFactory = $itemFactory;
     $this->entityType = $itemFactory->getEntityType();
-    $this->bundle = reset($itemFactory->getBundles());
+    $bundles = $itemFactory->getBundles();
+    $this->bundle = reset($bundles);
     $this->accessControlHandler = $entityTypeManager->getAccessControlHandler($this->entityType);
   }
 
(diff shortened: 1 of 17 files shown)


dafeder:
Just a note - getting close, but I am having an issue in the DatasetRevisionTest, where a request I think should be getting 403 is getting 200. Either a caching issue or I'm building the request wrong in the test..

modules/metastore/tests/src/Kernel/Controller/MetastoreAccessManagerTest.php
```diff
@@ -0,0 +1,222 @@
...
+      ->add(NodeInterface::class, 'language', LanguageInterface::class)
+      ->add(NodeInterface::class, 'getEntityType', EntityTypeInterface::class)
+      ->add(EntityTypeInterface::class, 'id', 'node')
+      ->add(LanguageInterface::class, 'getId', 'en')
+      ->add(FieldItemListInterface::class, 'getString', 'dataset')
+      ->getMock();
+
+    $this->container->set('dkan.metastore.metastore_item_factory', $itemFactory);
+
+    $this->priviledgedUser = $this->createUser([
```
paul-m: For each individual test, I'd create a user that has the minimum permissions per test. So `testCanCreate` would create a user with only the permissions required to create, and then the other test methods would create a user based on the minim required for their access check.

modules/metastore/src/Controller/MetastoreAccessManager.php
```diff
@@ -0,0 +1,250 @@
...
+      return $this->accessControlHandler->access($entity, "update", $account, TRUE);
+    }
+    catch (MissingObjectException | \InvalidArgumentException $e) {
+      // If this is a PUT, we should check if user has CREATE permission.
+      if ($method === 'PUT' && !$this->canCreate($schema_id, $account)->isAllowed()) {
+        return AccessResult::forbidden();
+      }
+      // If the the item does not exist, assume "allowed" and let the controller
+      // handle the 404 response.
+      return AccessResult::allowed();
```
paul-m, line 138: It seems weird to see this in code, tho I know we discussed it. It seems like missing object would be a different outcome than invalid argument, but I guess not.

It looks like we have this pattern covered in `DatasetItemTest` so 👍

paul-m (changes requested):

modules/metastore/src/Controller/MetastoreAccessManager.php
```diff
@@ -0,0 +1,250 @@
...
+   */
+  protected function getEntity(string $schema_id, string $identifier): EntityInterface {
+    // Load the entity based on schema ID and item ID.
+    try {
+      $item = $this->itemFactory->getInstance($identifier, ['schema_id' => $schema_id]);
+    }
+    catch (\Throwable $e) {
+      throw new MissingObjectException("No item found for schema ID '$schema_id' and identifier '$identifier'.", 0, $e);
+    }
+    assert($item->getSchemaId() === $schema_id, \InvalidArgumentException::class);
```
dafeder: Switch this to be also MissingObjectException

modules/metastore/src/Controller/MetastoreAccessManager.php
```diff
@@ -0,0 +1,250 @@
...
+   * @param string $schema_id
+   *   The schema ID.
+   * @param \Drupal\Core\Session\AccountInterface $account
+   *   The user account.
+   *
+   * @return \Drupal\Core\Access\AccessResult
+   *   An access result object indicating whether the user can create an item.
+   */
+  public function canCreate(string $schema_id, AccountInterface $account): AccessResult {
+    return $this->accessControlHandler->createAccess($this->bundle, $account, [], TRUE);
```
dafeder, line 97: We should not just kill that perm because existing users/roles may not have updated perms. Let's check first for the 'post put delete API' permission and when we don't find it, use the new access perms.

modules/metastore/tests/src/Kernel/Controller/MetastoreAccessManagerTest.php
dafeder: Yeah, I thought about doing this and got lazy 🙃 . Will do.

dafeder: Please see new test @paul-m

modules/metastore/tests/src/Kernel/Controller/MetastoreAccessManagerTest.php
```diff
@@ -0,0 +1,268 @@
...
+      ->add(NodeInterface::class, 'getEntityTypeId', 'node')
+      ->add(NodeInterface::class, 'id', '123')
+      ->add(EntityTypeInterface::class, 'id', 'node')
+      ->add(LanguageInterface::class, 'getId', 'en')
+      ->add(FieldItemListInterface::class, 'getString', 'dataset')
+      ->getMock();
+
+    $this->container->set('dkan.metastore.metastore_item_factory', $itemFactory);
+
+    $this->createUser([], 'superuser', TRUE);
```
dafeder, line 97: I hadn't run into this before... the first user created was being made a superuser so I needed to make one first before being able to make a user with more limited permissions...

janette (approved):
Code looks good @dafeder, and passes all manual testing.

janette:
@paul-m if you are satisfied with the test updates let's merge this.

paul-m (approved):

PR #4528, opened 2025-08-01
Updates to query schema

### LINKED ISSUE #4527: DKAN datastore queries schema improvements
### Current Behavior

Invalid enums not throwing errors. 

### Expected Behavior

Invalid enum for example `='` should throw 400 error and say something like 

```

{
  "message": "JSON Schema validation failed.",
  "status": 400,
  "timestamp": "2025-08-01T20:20:33+00:00",
  "data": {
    "keyword": "enum",
    "pointer": "conditions\/1\/conditions\/1\/operator",
    "message": "The attribute must be one of the following values: \u0027=\u0027, \u0027\u003C\u003E\u0027, \u0027\u003C\u0027, \u0027\u003C=\u0027, \u0027\u003E\u0027, \u0027\u003E=\u0027, \u0027like\u0027, \u0027between\u0027, \u0027in\u0027, \u0027not in\u0027, \u0027contains\u0027, \u0027starts with\u0027, \u0027match\u0027."
  }
}
```

### Steps To Reproduce

TBD

```shell

```

### Anything else?

_No response_

kaise-lafrai:
Fixes #4527 

## Describe your changes

Query Schema not as strict as we want. Updates made to throw json schema validation errors if any of the conditions are not met. For example: "operator": "='" currently results in a status 500 when it should result is status 400, JSON Schema validation failed. 

## QA Steps

- [ ] Setup a dataset
- [ ] Create a post request to search for a column with a specific value. Request should return a 200 and results should be returned. Example request:
```
POST https://example.site/api/1/datastore/query/{datasetid}/0/ HTTP/1.1a

{
  "conditions":[
    {
      "operator":"=",
      "property":"column_name",
      "value":"any_value"
    }
  ]
}
```

- [ ] Now try to modify the same request and pass a bad operator. For example: change operator to `='`.
- [ ] Result should fail with 400 error with message: JSON Schema validation failed.

### DIFF
--- a/modules/datastore/docs/query.json
+++ b/modules/datastore/docs/query.json
@@ -213,7 +213,7 @@
                     "default": "="
                  }
             },
-            "required": [ "property", "value", "operator" ]
+            "required": [ "property", "value" ]
         },
         "expression": {
             "type": "object",
--- a/modules/datastore/docs/query.json
+++ b/modules/datastore/docs/query.json
@@ -165,7 +165,8 @@
                         ]
                     }
                 }
-            }
+            },
+            "required": ["conditions"]
         },
         "condition": {
             "type": "object",
@@ -212,7 +213,7 @@
                     "default": "="
                  }
             },
-            "required": [ "property", "value" ]
+            "required": [ "property", "value", "operator" ]
         },
         "expression": {
             "type": "object",
--- a/modules/datastore/tests/data/query/invalidConditionGroupQuery.json
+++ b/modules/datastore/tests/data/query/invalidConditionGroupQuery.json
@@ -5,12 +5,25 @@
           "alias": "t"
         }
     ],
-    "conditions": [
+  "conditions":[
+    {
+      "operator":"=",
+      "property":"column_a",
+      "value":"value"
+    },
+    {
+      "conditions":[
         {
-          "operator": "=",
-          "property": "filterable",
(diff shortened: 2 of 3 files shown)


modules/datastore/docs/query.json
```diff
@@ -212,7 +213,7 @@
                     "default": "="
                  }
             },
-            "required": [ "property", "value" ]
+            "required": [ "property", "value", "operator" ]
```
dafeder: @kaise-lafrai will this still work if we keep operator optional? We have operator default to "=" so it would ideally remain optional. I think it's just the required "conditions" that was missing?

modules/datastore/tests/data/query/invalidConditionGroupQuery.json
```diff
@@ -0,0 +1,29 @@
...
+        {
+          "operator":"=",
+          "property":"column_b",
+          "value":"value"
+        },
+        {
+          "operator":"=",
+          "property":"column_b",
+          "value":"value"
+        }
```
dafeder, line 25: Not sure how this is relevant to the changes made.. seems to just check for whether omitting a groupOperator in a conditionGroup will fail.

dafeder (changes requested):

dafeder (approved):

PR #4536, opened 2025-08-14
Override file access to allow access to remote files for UploadAndLink.

beeyayjay:
Fixes 25771

## Problem:
When using the _Link to Data File_ option in the Download URL field, the field element is hidden for other users on the edit form.

## Details:
In #4463 the UploadOrLink element was refactored to allow remote files to be stored as managed files similar to uploaded files. We have discovered that if an editor other than the original author tries to edit the dataset, the UploadOrLink element will not display due to the _checkAccess_ in the core file **FileAccessControlHandler**. The checkAccess is checking that the scheme is 'public', but for remote files the scheme is 'https', and if you are not the author then access to the field is denied.

## Describe your changes
Added a custom access control handler to allow download access to files using the 'https' scheme.

## QA Steps

- [ ] Create two data publisher accounts, user 1 and user 2
- [ ] Log in as user 1
- [ ] Create a dataset, and use the "Link to Data File" option and enter a remote url to a file.
- [ ] Create a second dataset and use the upload option, upload a local file
- [ ] Confirm the upload option works as expected
- [ ] Log in as user 2.
- [ ] Edit dataset 1 and confirm that the Download URL field displays the link and that you can edit it
- [ ] Edit dataset 2, confirm the file displays in the Download URL field, and that you can edit it

### DIFF
--- a/modules/json_form_widget/json_form_widget.module
+++ b/modules/json_form_widget/json_form_widget.module
@@ -91,3 +91,12 @@ function json_form_widget_preprocess_file_link(array &$variables) {
     $variables['link']['#title'] = $file->getFileUri();
   }
 }
+
+/**
+ * Gives linked files the same access permissions as uploaded files.
+ */
+function json_form_widget_entity_type_build(array &$entity_types) {
+  if (isset($entity_types['file'])) {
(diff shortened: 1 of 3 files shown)


janette:
@beeyayjay please fix the indentation. And we will need a test for this code as well.

janette (approved):

PR #4533, opened 2025-08-07
Move JSON form plugin to metastore, de-DKAN-ifying json_form_widget

dafeder:
Fixes #4465

Moving the Form widget plugin to metastore, leaving json_form_widget with no DKAN-specific code in src.

Ideally, this PR should also:

* Offer a generic JsonFormWidget plugin that does not rely on DKAN schema retriever
* Refactor the rest of the tests to use this generic widget

## QA Steps

Functionality should not change. Maybe try starting with a 2.x build, either of a client site or with sample content, then switching to this branch, running updatedb and making sure the dataset form still works.

PR #4540, opened 2025-08-18
Plugins for JSON Form option sources

### LINKED ISSUE #4538: Plugins for select option sources in JSON form widget
JSON Form Widget currently supports this config to pull options for select2 from another metastore schema:

```json
{
  "keyword": {
    "ui:options": {
      "hideActions": "true",
      "child": "keyword"
    },
    "items": {
      "ui:options": {
        "widget": "list",
        "type": "autocomplete",
        "allowCreate": "true",
        "multiple": "true",
        "source": {
          "metastoreSchema": "keyword"
        }
      }
    }
  }
}
```

We are trying to de-DKAN-ify the module to let it stand on its own. This included moving the actual field widget plugin to DKAN (#4533) but needs to also somehow accommodate this, which is parsed in the WidgetRouter.

I think to accomplish this we need another plugin type, for item sources. The simplest way to do this, which would support existing syntax, would be to allow a source type and an argument, which would come from the property name and value here in the source object.

However, this is kind of a bad pattern due to existing tech debt. Relying on a property _name_ makes validation tricky (what if `source` has two properties?). Also, in some cases, the source logic needs to know the value of another property, `titleProperty`, which lies outside of the `source`. We have this to support the more complex behavior in `publisher`:

```json
{
  "publisher": {
    "ui:options": {
      "widget": "list",
      "type": "autocomplete",
      "allowCreate": "true",
      "titleProperty": "name",
      "source": {
        "metastoreSchema": "publisher"
      }
    },
    "properties": {
      "@type": {
        "ui:options": {
          "widget": "hidden"
        }
      },
      "subOrganizationOf": {
        "ui:options": {
          "widget": "hidden"
        }
      }
    }
  }
}
```

I think the correct way to do this would be more like:

```json
{
  "publisher": {
    "ui:options": {
      "widget": "list",
      "type": "autocomplete",
      "allowCreate": "true",
      "source": {
        "plugin": "metastore"
        "config": {
          "schema": "publisher",
          "titleProperty": "name",
        }
      }
   }
}
```

Where `plugin` gets the plugin ID, and `config` is passed along as an associative array.

Now we implement a simple plugin type that takes the config and has an interface to return the options.

dafeder:
Fixes #4538

This PR adds another plugin type for "options sources." These plugins allow us to provide an array of options from any arbitrary source to be used for select fields, autocompletes etc when not supplied directly in the schema. This replaces the old "metastoreSchema" source property in UI schemas so that we can fully decouple JSON Form Widget from DKAN.

Backwards compatibility is provided with the old approach in the UI schema. Tests have been moved around to reflect where the code lives now, but should be equivilant.

A new Taxonomy source plugin is provided in the JSON Form Widget module, mostly as an example for other implementations.

## QA Steps

Functionality should not actually change, this is more about decoupling the code. Confirm the form still works as expected.

## Merge instructions

This should probably be rebased and merged after #4533 is merged. I tested a rebase locally and tests still passed...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -110,6 +110,7 @@ commands:
                 command: |
                   ddev composer config repositories.dkan path dkan
                   # To avoid dependency issues when installing on PHP 8.1 we need to re-require openapi-httpfoundation-testing.
+                  rm composer.lock
                   ddev composer require --dev osteel/openapi-httpfoundation-testing -W
                   ddev composer require getdkan/dkan:@dev --no-install -W
                   ddev composer update getdkan/dkan -W
--- a/modules/metastore/src/Plugin/JsonFormOptionSource/MetastoreSchema.php
+++ b/modules/metastore/src/Plugin/JsonFormOptionSource/MetastoreSchema.php
@@ -49,7 +49,10 @@ class MetastoreSchema extends JsonFormOptionSourcePluginBase implements Containe
     $this->metastore = $metastore;
   }
   
-  public function getMetastore() {
+  /**
+   * Return the metastore service.
+   */
+  public function getMetastore(): MetastoreService {
     return $this->metastore;
   }
 
@@ -78,6 +81,13 @@ class MetastoreSchema extends JsonFormOptionSourcePluginBase implements Containe
       $options[$value] = $title;
     }
     return $options;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getTargetType(array $config): string {
+    return 'node';
   }
 
   /**
--- a/modules/metastore/src/Plugin/JsonFormOptionSource/MetastoreSchema.php
+++ b/modules/metastore/src/Plugin/JsonFormOptionSource/MetastoreSchema.php
@@ -18,7 +18,7 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
  *   description = @Translation("Foo description.")
  * )
  */
-final class MetastoreSchema extends JsonFormOptionSourcePluginBase implements ContainerFactoryPluginInterface {
+class MetastoreSchema extends JsonFormOptionSourcePluginBase implements ContainerFactoryPluginInterface {
 
   /**
    * The metastore service.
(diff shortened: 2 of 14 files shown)


dafeder:
WidgetRouter and its test are complete, but still need some work on SchemaUiHandler and its tests.

modules/metastore/src/Plugin/JsonFormOptionSource/MetastoreSchema.php
```diff
@@ -0,0 +1,149 @@
...
+use Drupal\metastore\MetastoreService;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Plugin implementation of the json_form_option_source.
+ *
+ * @JsonFormOptionSource(
+ *   id = "metastoreSchema",
+ *   label = @Translation("Foo"),
+ *   description = @Translation("Foo description.")
```
kaise-lafrai: Update label and description

modules/metastore/src/Plugin/JsonFormOptionSource/MetastoreSchema.php
```diff
@@ -0,0 +1,149 @@
...
+    if ($config['titleProperty'] ?? FALSE) {
+      return is_object($item) ? $item->data->{$config['titleProperty']} : $item;
+    }
+    return $item->data;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function validateConfig(array $config): bool {
```
kaise-lafrai, line 136: I can't seem to find where this would be called.

kaise-lafrai: I think we may need to add `$this->validateConfig($config);` under the `getOptions()` method.

kaise-lafrai (changes requested):

modules/metastore/src/Plugin/JsonFormOptionSource/MetastoreSchema.php
dafeder: Ooops - correct. I did so in the TaxonomySource and forgot to here.

dafeder: See now

.circleci/config.yml
```diff
@@ -110,6 +110,7 @@ commands:
                 command: |
                   ddev composer config repositories.dkan path dkan
                   # To avoid dependency issues when installing on PHP 8.1 we need to re-require openapi-httpfoundation-testing.
+                  rm composer.lock
```
kaise-lafrai, line 113: This is new. Why are we removing composer.lock?

kaise-lafrai (approved):

Issue #3811, opened 2022-07-29
Allow \Harvest\Harvester to log errors for users and devs


paul-m:
## User Story

This issue supports WCMS-10613.

This issue is almost a bug report, but the code currently works. It's just not easy to determine what went wrong in a failed content import.

Trying to perform the Drush commands for import from sample_content module, I get a status like this:
```
% ddev drush dkan:sample-content:create
+----------------+-----------+---------+---------+--------+
| run_id         | processed | created | updated | errors |
+----------------+-----------+---------+---------+--------+
| sample_content | 10        | 0       | 0       | 10     |
+----------------+-----------+---------+---------+--------+
```
Unfortunately the errors are only counted and not logged, so there is no path forward.

This effort should probably focus on `\Harvest\Harvester::extract()`, `::transform()`, and `::load()`, which catch exceptions and store their error outputs in an array, which is subsequently discarded.

## Acceptance Criteria

Attempting to perform a harvest results in error logs for errors that occur.

PR #4541, opened 2025-08-18
HarvestService reports extraction errors

paul-m:
Addresses [#3811]

## Describe your changes

Rather than making all harvest extraction errors have the same error message based on whether anything was extracted, we throw an exception when the extract phase reports a failure, so the exception can potentially float up to the user.

## QA Steps


dafeder:
Needs update to 4.x.

PR #4544, opened 2025-08-22
Minimum of 1 for the "Limit" property in the datastore query schema

### LINKED ISSUE #4543: Stop accepting limits of -1 or 0
Passing a datastore query a limit of `-1` will produce a 400 database error. This should actually be a schema violation, so let's make the schema only allow positive numbers. Possibly enforce in SelectFactory as well.

Example with sample content

https://dkan.ddev.site/api/1/datastore/query/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/0?limit=-1

Passing zero means no limit, and actually overrides what is supposed to be the hard limit on non-download queries. We do promise this in the API but this seems like it was a bad choice and is worth a very small backwards-incompatible change. So let's just reject a limit=0 argument.

To reproduce, try

https://dkan.ddev.site/api/1/datastore/query/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/0?limit=0 and observe that the number of results returned is in the thousands, while even passing a limit of 600 is rejected.

Simply adding a minimum of 1 to the schema for the limit property should fix both these issues. The download endpoint should continue to function the same way with the exception that limit=0 would also be rejected.

Add tests to confirm this is all working as described. Update wording in schema.

kaise-lafrai:
Fixes #4543 

## Describe your changes

Added a minimum attribute set to 1 for the limit property of the query schema. This change will affect both queries and downloads when a limit is present that is not 1 or greater. This change does not affect the existing row limit setting that is found at `/admin/dkan/datastore`. The default row limit is set to 500 and this will continue to work as expected for queries. Downloads are not capped at a specific row limit and will continue to work the same.

## QA Steps

- [ ] Create sample content
- [ ] https://dkan-project.ddev.site/api/1/datastore/query/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/0/download?&format=csv&limit=0 will throw JSON Schema validation failed.
- [ ] https://dkan-project.ddev.site/api/1/datastore/query/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/0?limit=0 will throw JSON Schema validation failed.
- [ ] https://dkan-project.ddev.site/api/1/datastore/query/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/0/download?&format=csv&limit=1 will work
- [ ] https://dkan-project.ddev.site/api/1/datastore/query/cedcd327-4e5d-43f9-8eb1-c11850fa7c55/0?limit=1 will work

### DIFF
--- a/modules/datastore/docs/query.json
+++ b/modules/datastore/docs/query.json
@@ -84,7 +84,7 @@
         },
         "limit": {
             "type": "integer",
-            "description": "Limit for maximum number of records returned. Pass zero for no limit (may be restricted by user permissions).",
+            "description": "Limit for maximum number of records returned. Must be a minumum of 1.",
             "minimum": 1
         },
         "offset": {
--- a/modules/datastore/docs/query.json
+++ b/modules/datastore/docs/query.json
@@ -84,7 +84,8 @@
         },
         "limit": {
             "type": "integer",
-            "description": "Limit for maximum number of records returned. Pass zero for no limit (may be restricted by user permissions)."
+            "description": "Limit for maximum number of records returned. Pass zero for no limit (may be restricted by user permissions).",
+            "minimum": 1
         },
         "offset": {
             "type": "integer",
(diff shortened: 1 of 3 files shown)


dafeder (commented):
Looks good! Approving but holding off for a minute before merging to think about release strategy, as this is a breaking change.

dafeder (approved):

PR #4546, opened 2025-08-25
Remove select2 asset from composer require.

### LINKED ISSUE #4545: Trying to make a drupal module require dkan core fails with select2 composer issues.
### Current Behavior

When trying to add dkan to drupal module as a dependency, the pipeline and preview environments fail to build due to issues with the select2 dependency

I've seen the previous issue [here](https://github.com/GetDKAN/dkan/issues/3740)  and the latest PR that just merged [here](https://github.com/GetDKAN/dkan/pull/4479) as well as a previous internal slack thread.

I implemented the suggestion in DKAN-3740 to " add asset-packagist as a repository to your project's composer.json"   I have done that in the MR on this issue https://www.drupal.org/project/dkan_dataset_archiver/issues/3540370 but it did not resolve the problem.

I suspect other module maintainers have encountered this same blocker because none of the existing current dkan related modules on d.o reference dkan as a dependency in either the info file or the composer... and that is problematic for many reasons.

I wonder if it would make more sense to include and call select2 as a llbrary rather than a composer dependency like was done here? https://git.drupalcode.org/project/content_model_documentation/-/merge_requests/45/diffs

### Expected Behavior

I expect that any module that is created that needs to set dkan as a dependency in the module's composer.json and .info file should be able to do so and have both the gitlab ci pipeline and tugboat PR environments build appropriately.  

Setting the dependency in only the .info file causes the pipeline to fail because it can't enable dkan because it does not exist.  Setting it in the composer.yml fails due to the error shown below.

### Steps To Reproduce

1. create module on d.o
2. specifiy dkan core dependency in the composer.json
3. specify gkan for as dependency in .info file.
4. Watch CI pipeline fail catastrophically
5. Watch Tugboat preview fail catastrophically
6. If glutton for punishment, try to preview module on simplytest.me and watch that also fail to build.

```shell
Problem 1
    - Root composer.json requires drupal/dkan_dataset_archiver * -> satisfiable by drupal/dkan_dataset_archiver[dev-main].
    - drupal/dkan_dataset_archiver dev-main requires getdkan/dkan ^2.20.9 -> satisfiable by getdkan/dkan[2.20.9, 2.x-dev].
    - getdkan/dkan 2.20.9 requires npm-asset/select2 ^4.0 || dev-develop -> could not be found in any version, there may be a typo in the package name.
    - getdkan/dkan 2.x-dev requires npm-asset/select2 dev-develop#595494a -> could not be found in any version, there may be a typo in the package name.
```

### Anything else?

_No response_

swirtSJW:
This should be handled at the project level, following [guidance from the select2 module readme](https://git.drupalcode.org/project/select2#composer-recommended).

### DIFF
--- a/docs/source/installation/index.rst
+++ b/docs/source/installation/index.rst
@@ -63,7 +63,7 @@ If you already have an existing Drupal site, install DKAN with `composer <https:
    Do note that a bug in Drupal core cron may cause problems with data imports, and applying `this patch <https://www.drupal.org/project/drupal/issues/3274931>`_ is highly recommended. The patch will be applied automatically if you use the `recommended project <https://github.com/GetDKAN/recommended-project>`_.
 
 
-   warning::
+.. warning::
    DKAN requires som additional composer changes due to a dependency on the Select2 library. You should follow the steps in the `Select2 project's Readme <https://git.drupalcode.org/project/select2/-/blob/2.x/README.md?ref_type=heads#installation>`_.
 
 
--- a/docs/source/installation/index.rst
+++ b/docs/source/installation/index.rst
@@ -64,7 +64,7 @@ If you already have an existing Drupal site, install DKAN with `composer <https:
 
 
 .. warning::
-   DKAN requires som additional composer changes due to a dependency on the Select2 library. You should follow the steps in the `Select2 project's Readme <https://git.drupalcode.org/project/select2/-/blob/2.x/README.md?ref_type=heads#installation>`_.
+   DKAN requires some additional composer changes due to a dependency on the Select2 library. You should follow the steps in the `Select2 project's Readme <https://git.drupalcode.org/project/select2/-/blob/2.x/README.md?ref_type=heads#installation>`_.
 
 
 Add some example datasets to your site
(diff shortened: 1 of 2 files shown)


stefan-korn:
I suppose the whole "repositories" key could be removed? and whole extra/installer-types as well?

If your really want to include the dependency from select2, maybe [Composer merge plugin](https://github.com/wikimedia/composer-merge-plugin) could be used. Though this would also require adjusting the root composer.json and maybe then it should better be done in select2 directly.

dafeder:
Yes @stefan-korn I think that's the next step, just wanted to keep this isolated for now. There's something about bower in there too but i haven't looked into it.

docs/source/installation/index.rst
```diff
@@ -62,6 +62,11 @@ If you already have an existing Drupal site, install DKAN with `composer <https:
 .. warning::
    Do note that a bug in Drupal core cron may cause problems with data imports, and applying `this patch <https://www.drupal.org/project/drupal/issues/3274931>`_ is highly recommended. The patch will be applied automatically if you use the `recommended project <https://github.com/GetDKAN/recommended-project>`_.
 
+
+.. warning::
+   DKAN requires som additional composer changes due to a dependency on the Select2 library. You should follow the steps in the `Select2 project's Readme <https://git.drupalcode.org/project/select2/-/blob/2.x/README.md?ref_type=heads#installation>`_.
```
swirtSJW:
```suggestion
   DKAN requires some additional composer changes due to a dependency on the Select2 library. You should follow the steps in the `Select2 project's Readme <https://git.drupalcode.org/project/select2/-/blob/2.x/README.md?ref_type=heads#installation>`_.
```

dafeder (approved):
Another one that should trip 2.21 IMO. But looks good.

PR #4547, opened 2025-08-27
Remove other unneeded properties from composer.json

dafeder:
As noted by @stefan-korn in #4545 discussion, there is a lot in DKAN's composer.json that doesn't make sense there; these are properties that belong in a _project root_ composer.json and DKAN as a module will always be a dependency of another Drupal project.

Existing projects will need to ensure that they have `oomphinc/composer-installers-extender` added to their composer.json file. See https://github.com/GetDKAN/recommended-project/commit/547d534365cda9f89d62a38722542b7160042aae.

Make sure to include these instructions in release notes!

## QA Steps

* Start a new project or delete your vendor, composer.lock and docroot/libraries/select2 files/folders
* Check out this branch of DKAN
* Make sure oomphinc/composer-installers-extender (as well as select2) are in your root composer.json file.
* Run composer install
* Confirm that build completes and docroot/libraries/select2 exists

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,6 @@
         "ilbee/csv-response": "^1.2.0",
         "justinrainbow/json-schema": "^5.2",
         "m1x0n/opis-json-schema-error-presenter": "^0.5.3",
-        "oomphinc/composer-installers-extender": "^2.0",
         "ramsey/uuid": "^3.8.0 || ^4",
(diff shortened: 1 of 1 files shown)


paul-m (approved):
LGTM, but maybe update recommended-project first, and then release notes can say to get a refresh from recommended-project, or just add oomphinc to your project.

PR #4555, opened 2025-09-12
Move metastore hook_update to hook_deploy

### LINKED ISSUE #4554: 2.21.0 hook_update_N gets wiped out during drush deploy
### Current Behavior

The hook_update_n in #4533 that was released as [2.21.0](https://github.com/GetDKAN/dkan/releases/tag/2.21.0) gets undone / overwritten during Drush Deploy
Drush deploy runs the following in this order:

1. updatedb (runs that hook)
2. config:import (overwrites the settings that update just changed)
3. cache rebuild
4. deploy hooks
5. cache warm

Then a developer would do a config export, which will show no changes.   This update hook since it is modifying config, should be re-run as a deploy hook so it runs after config import rather than before.

### Expected Behavior

Updating to the latest release and running drush deploy  then followed by config export, should result in the intended configuration being captured.

### Steps To Reproduce

1. composer require 2.21.0
2. drush deploy
3. drush cex
4. Notice the intended config changes in the hook update_n do not exist.

```shell

```

### Anything else?

_No response_

swirtSJW:
Fixes #4554

## Describe your changes
Moves the metastore_update_8011 to a deploy hook so that config changes do not get overwritten during drush deploy.
## QA Steps

- [ ] git checkout 4544-move-hook-update-to-deploy
- [ ] ddev composer install
- [ ] ddev drush deploy
- [ ] validate in the output that the deploy hook ran.
- [ ] ddev drush cex
- [ ] validate that you see new settings for the dkan_json_form_widget

### DIFF
--- a/modules/metastore/metastore.deploy.php
+++ b/modules/metastore/metastore.deploy.php
@@ -1,0 +1,26 @@
+<?php
+
+/**
+ * @file
+ * Deploy hooks for the metastore module.
+ */
+
+/**
+ * Update the dkan_json_form_widget field widget settings.
+ */
+function metastore_deploy_1001(&$sandbox) {
+  // Retrieve the entity form display settings for the 'data' content type.
+  $entity_form_display = \Drupal::service('config.factory')
+    ->getEditable('core.entity_form_display.node.data.default');
+  // Update the 'dkan_json_form_widget' field widget settings.
+  $settings = $entity_form_display->get('content.field_json_metadata');
+  if (empty($settings)) {
+    return "No settings found for 'field_json_metadata'.";
+  }
+  if ($settings['type'] == 'json_form_widget') {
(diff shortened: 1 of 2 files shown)


swirtSJW:
image

swirtSJW:
Hmm the metastore deploy hook runs, but still does not change configuration 
image

I'm investigating.

swirtSJW:
Ok I added some more output to the deploy hook and I see why there are no config changes to PDC.  PDC has no config to update. 
image

So I chased this herring which I still think is a real issue, but for PDC only it turned out to be a red herring.

swirtSJW:
It did run, just did not fall into the "if" clauses due  to PDC issues and not the fault of this move.

dafeder:
Let's discuss later today if you're able @swirtSJW . To be honest I'm not very familiar with deploy hooks.

PR #4556, opened 2025-09-16
Allow case-insensitive condition operators

dafeder:
Now that a bug in query validation has been fixed (#4528) a lot of datastore queries that were working in the past using SQL-style all-caps operators (particularly, `LIKE`) are now failing, because the API expects all lowercase. After some discussion, we decided it should be more forgiving.

This switches to case-insensitive regex validation for alphanumeric operators. While the `i` regex flag is not guaranteed to work in JSON-Schema generally, it does seem to be fully respected in PHP implementations.

Tests have been updated to better test this part of the validation.


.circleci/config.yml
```diff
@@ -51,7 +51,8 @@ commands:
             ddev --version
             ddev config global --instrumentation-opt-in=false
             ddev config --project-name test-$CIRCLE_WORKFLOW_JOB_ID --project-type drupal9 --docroot docroot
-            ddev add-on get https://github.com/GetDKAN/ddev-dkan/archive/refs/heads/<< parameters.addon_branch >>.tar.gz
+            wget -P ../ https://github.com/GetDKAN/ddev-dkan/archive/refs/heads/<< parameters.addon_branch >>.tar.gz
+            ddev add-on get ../<< parameters.addon_branch >>.tar.gz
```
dafeder: For some reason, add-on get stopped working with a github url, even though curl and wget can download the file just fine. Had to switch to this.

paul-m (approved):

PR #4562, opened 2025-09-22
Use resource id not dist in dashboard

### LINKED ISSUE #4561: List resource ID instead of distribution ID in datastore import dashboard
The datastore import dashboard has a column called "resource" but shows an identifier which is actually the distribution ID. Aside from this just being a straight-up incorrect label, it's not clear that anyone needs the distribution ID; the resource ID is what you would want to copy to use with Drush commands on the datastore. SO let's swap the distribution UUID for the actual resource ID.

Would be nice if we could include a little clickable [copy] icon too.

dafeder:
Fixes #4561

## QA steps

1. Build from branch
2. Load and import sample content
3. Visit admin/dkan/datastore/status
4. Confirm resource ID for bike lanes dataset is `3a187a87dc6cd47c48b6b4c4785224b7`
5. Run `drush dkan:dataset-info cedcd327-4e5d-43f9-8eb1-c11850fa7c55`
6. Confirm resource ID is correct

### DIFF
--- a/modules/datastore/src/Form/DashboardForm.php
+++ b/modules/datastore/src/Form/DashboardForm.php
@@ -574,7 +574,7 @@ class DashboardForm extends FormBase {
       [
         'data' => [
           '#theme' => 'datastore_dashboard_resource_cell',
-          '#uuid' => $dist['distribution_uuid'],
+          '#uuid' => $dist['resource_id'],
           '#file_name' => basename((string) $dist['source_path']),
           '#file_path' => UrlHostTokenResolver::resolve($dist['source_path']),
         ],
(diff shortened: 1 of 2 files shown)


modules/datastore/src/Form/DashboardForm.php
```diff
@@ -574,7 +574,7 @@ protected function buildResourceData(array $dist, bool $importable, string $post
       [
         'data' => [
           '#theme' => 'datastore_dashboard_resource_cell',
-          '#uuid' => $dist['distribution_uuid'],
+          '#uuid' => $dist['resource_id'],
```
janette, line 577: Does it matter that the data property is still uuid?

dafeder: Well - depends what "matters" means 😄 . I did change that too at first but it required changing a lot more files, templates etc. I don't feel strongly though, if you think we should change the element/variable name too I can redo.

dafeder:
Now less sure, seems like there are other workflows where it's more useful to get the distribution ID...

dafeder:
Closing in favor of #4615

PR #4563, opened 2025-09-23
Restrict view perms, remove routing perms

### LINKED ISSUE #4531: Authorization Concern - Depositor role has access to /admin/dkan/datasets
### Current Behavior

Credit: @ncDevMan

We found an issue a few days ago where Depositors, who are our non-admin, authenticated users, had `create data content` privileges in the `user.role.depositor.yml` file.

This was giving Depositors, who are not admins, access to /admin/dkan/datastets and /admin/dkan/data-dictionaries because of the metastore.routing.yml ( https://github.com/GetDKAN/dkan/blob/2.x/modules/metastore/metastore.routing.yml ).

When we remove `create data content` permissions the Depositor, the depositor cannot access the metadata edit form because it uses the same permission.

We saw that some of the routes in the metastore.routing.yml file had administer in the permission name. 

### Expected Behavior

Depositors should not have access to /admin pages but still have access to metadata editing for datasets they have created.

### Steps To Reproduce

1. Submit a dataset as a depositor with metadata.
2. Attempt to access a dataset under the url /admin/dkan/datastets

```shell

```

### Anything else?

_No response_

dafeder:
Fixes #4531

## Describe your changes

Switches to "edit any data content" as perm in the view config, removes perm requirement from routing.

## Include in release notes

> [!NOTE] 
> This release fixes a problem reported in #4531; that the permissions to access the datasets admin page were too permissive. The change will not be automatically applied; see the section below "View Update Instructions" for how to import the new config.

### View update instructions

The changes to the datasets admin view are not automatically applied, to avoid blowing away any customizations. If you are comfortable replacing the default datasets admin view at admin/dkan/datasets, use either one of the following methods:

* Run the drush command `drush cim --partial --source=modules/contrib/dkan/modules/metastore/modules/metastore_admin/config/install -y`
* Use the configuration sync UI:
  1. Open the file modules/metastore/modules/metastore_admin/config/install/views.view.dkan_dataset_content.yml
  2. Copy the full contents
  3. Navigate to admin/config/development/configuration/single/import
  4. Set "Configuration type" to "View"
  5. Paste the contents of the yml file
  6. Click "Import"

## QA Steps

- [ ] Check out 2.x branch
- [ ] Log in as user 1 (`drush uli`)
- [ ] Create a new role, "depositor"
- [ ] Go to permissions and give depositor permission to create data content
- [ ] Create a user and give it the depositor role
- [ ] In another browser, log in as depositor user and confirm you can see /admin/dkan/datasets
- [ ] Check out branch `datasets-view-auth`
- [ ] Run `drush updatedb`, confirm update hook message makes sense
- [ ] Run `drush cim --partial --source=modules/contrib/dkan/modules/metastore/modules/metastore_admin/config/install -y`
- [ ] Confirm that the depositor can no longer access the dataset list
- [ ] Confirm that User 1 can access it, and all menu behavior is the same

### DIFF
--- a/modules/metastore/metastore.routing.yml
+++ b/modules/metastore/metastore.routing.yml
@@ -179,5 +179,4 @@ metastore.datasets:
   defaults:
     _controller: '\Drupal\system\Controller\SystemController::systemAdminMenuBlockPage'
     _title: 'Datasets'
-  requirements:
-    _permission: 'create data content'
+
--- a/modules/metastore/modules/metastore_admin/config/install/views.view.dkan_dataset_content.yml
+++ b/modules/metastore/modules/metastore_admin/config/install/views.view.dkan_dataset_content.yml
@@ -1,3 +1,4 @@
+uuid: 46a91364-3cbb-4ebd-a24c-264a2a2f4b58
 langcode: en
 status: true
 dependencies:
@@ -8,6 +9,8 @@ dependencies:
     - content_moderation
     - node
     - user
+_core:
+  default_config_hash: o0UfNa65CPMhdBKibp7p9bkwwQxjlfa-7_DFEN6FDcM
 id: dkan_dataset_content
 label: 'DKAN Datasets'
 module: node
@@ -594,7 +597,7 @@ display:
       access:
         type: perm
         options:
-          perm: 'create data content'
+          perm: 'edit any data content'
       cache:
         type: tag
       empty:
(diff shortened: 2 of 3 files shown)


janette:
After including the Use Publish transition permission as well this works as described. Can we redirect this type of user to their profile page rather than the access denied screen?

dafeder:
Ah - yes good point, I always forget this view is the default redirect after submitting a dataset. OK will think about a better way to handle this.

janette (approved):

janette:
The redirect issue is out of scope and can be addressed separately.

PR #4575, opened 2025-10-07
Fix JSON Form Widget upgrade path

dafeder:
UPDATE: This is so far behind `2.x-transition` and `3.x` that it will probably need to be discarded and replaced. However, keeping it around because it contains some tests and stuff that have been lost in those two branches since we adjusted strategy.

This is intended to test that the JSON Form widget continues working in the UI after switching from an included submodule to a [standalone module](https://www.drupal.org/project/json_form_widget).

Unfortunately I am having trouble recreating in the test what I see in my local env. When I build from 2.x, then switch to the 3.x branch and update composer, I get:

1. A broken site with an error message related to the container.
2. When I clear caches, the site is working. But when I try to load the form, I get fatal an error related to a missing plugin class.
3. When I finally run `drush updatedb`, I then get a working form.

In the test:

1. The site does not appear broken at all at the beginning of the test, even w/o a cache clear or table truncate.
2. The form will load, but the JSON Form Widget is not present; we see simply a textarea for the JSON Metadata field.
3. After running updates, the form loads as expected.

High level, in both scenarios, we don't have a working JSON Form Widget form until running the db updates, so this seems like a worthwhile test to add. It is frustrating that I can't reproduce either what I see in my browser in the test, or what the test is experiencing in my browser.


tests/src/Functional/Dkan4xTransitionalUpdatePathTest.php
```diff
@@ -88,6 +89,14 @@ public function test4xTransitionUpdates(): void {
     $mysqlImportConfig->set('strict_mode_disabled', TRUE);
     $mysqlImportConfig->save();
 
+    // Log in as admin and visit /node/add/data to ensure form works.
```
paul-m, line 88: Add some comments about why this is the first step before updating.

tests/src/Functional/Dkan4xTransitionalUpdatePathTest.php
```diff
@@ -37,7 +37,8 @@ protected function setDatabaseDumpFiles() {
    */
   public function test4xTransitionUpdates(): void {
     // Drop cache container file to avoid issues with module moving.
-    \Drupal::database()->truncate('cache_container')->execute();
+    // Clear all caches.
+    // \Drupal::database()->truncate('cache_container')->execute();
```
paul-m: Is this actually needed?

paul-m (changes requested):

dafeder:
Closing, but for now please don't delete the branch

PR #4576, opened 2025-10-07
Minimal cleanup of dkan_js_frontend

paul-m:
## Describe your changes

- Fixes erroneous variable name `$js_pcss_pathath`
- Folds `RouteProvider::routeHelper()` into `addIndexPage()` which is renamed to `addRoutesFromConfig()` which is more descriptive.
- Modifies README to be less about Create React App.
- Declares dependency on metastore module.
- Some code optimization.
- CS.

### DIFF
--- a/modules/dkan_js_frontend/dkan_js_frontend.info.yml
+++ b/modules/dkan_js_frontend/dkan_js_frontend.info.yml
@@ -4,4 +4,5 @@ type: module
 core_version_requirement: ^10.2 || ^11
 dependencies:
   - drupal:field
+  - getdkan:metastore
 package: DKAN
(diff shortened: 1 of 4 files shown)


modules/dkan_js_frontend/dkan_js_frontend.info.yml
```diff
@@ -4,4 +4,5 @@ type: module
 core_version_requirement: ^10.2 || ^11
 dependencies:
   - drupal:field
+  - getdkan:metastore
```
dafeder: Have we done this before? I think `dkan` is the correct package/namespace. Especially if we're targeting drupal.org releases in the future, I don't want to hardcode "getdkan" anywhere outside composer.json.

dafeder (changes requested):
Looks good - see one comment on info file and can you please edit the PR body so it gives some explanation of what happened?

dafeder (approved):

PR #4583, opened 2025-10-23
Add getTableName() to DatabaseTableInterface

dafeder:
Noticed in IDE that `AbstractDatabaseTable` calls `$this->getTableName()` but that method is not defined either in the abstract class or the interface. Added to the interface, and updated implementations.

### DIFF
--- a/modules/harvest/src/Storage/HarvestHashesEntityDatabaseTable.php
+++ b/modules/harvest/src/Storage/HarvestHashesEntityDatabaseTable.php
@@ -271,7 +271,7 @@ class HarvestHashesEntityDatabaseTable implements DatabaseTableInterface {
    * {@inheritdoc}
    */
   public function getTableName() {
-    return 'harvest_hashes';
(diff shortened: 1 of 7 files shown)


modules/harvest/src/Storage/HarvestHashesEntityDatabaseTable.php
```diff
@@ -267,4 +267,11 @@ protected function loadEntity(string $data_uuid): ?HarvestHashInterface {
     return NULL;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getTableName() {
+    return 'harvest_hashes';
```
paul-m: Follow the example of `get/setSchema()`:
```suggestion
    throw new \RuntimeException(__METHOD__ . ' not yet implemented.');
```

paul-m (changes requested):

paul-m (approved):

PR #4587, opened 2025-10-24
2.x-transition: Add DKAN 3.x transitional namespaced modules

dafeder:
This is a feature branch preparing what should be the last 2.x release, which adds stubs for the newly-namespaced modules. These modules will fully replace common, metastore, harvest and datastore with dkan_ namespaced equivalents in DKAN 3.x and 4.x.


dafeder:
Currently, config copying happens in this branch, let's remove it and handle it in 3.x.

modules/dkan_common/config/install/dkan_common.settings.yml
```diff
@@ -0,0 +1 @@
+always_use_existing_local_perspective: false
```
paul-m: Per conversation: Remove config from namespaced modules and re-add in 3.x

paul-m (changes requested):

paul-m (approved):

PR #4590, opened 2025-10-28
Allow streamed/download JSON data

### LINKED ISSUE #4559: Allow streamed/download JSON data
We have a /download endpoint under the datastore API that allows streaming a very large result set, but only as CSV. We should be able to handle JSON as in the streaming download controller. It looks like the [JsonStreamer component](https://symfony.com/packages/JsonStreamer) should provide most of what we need to add JSON support to the QueryDownloadController reusing most of what we already have for the CSV download.

beeyayjay:
Fixes #4559 

## Describe your changes
Added option to get download queries in json format.

## QA Steps

- [ ] Get the distribution id of an existing dataset with datastore.
- [ ] In browser url enter `[BASE URL}/api/1/datastore/query/[DATASET ID]/download?format=json` (e.g. 'https://dkan.ddev.site/api/1/datastore/query/0d89f2f7-8472-5121-96af-a43ff7017df1/download?format=json')
- [ ] Confirm that the browser responds by downloading a file named data.json with all of the data in json format.
- [ ] Go to `/api` and scroll down to GET /api/1/datastore/query/{Distribution ID}/download, and expand the section.
- [ ] Click the 'Try it out' button and confirm that the 'format' parameter description is "Response format. Either csv or json." and that both formats work

- [ ] I have updated or added documentation - not necessary

### DIFF
--- a/modules/datastore/src/Controller/QueryDownloadController.php
+++ b/modules/datastore/src/Controller/QueryDownloadController.php
@@ -38,7 +38,7 @@ class QueryDownloadController extends AbstractQueryController {
     DatastoreQuery $datastoreQuery,
     RootedJsonData $result,
     array $dependencies = [],
-    ?ParameterBag $params = NULL
+    ?ParameterBag $params = NULL,
   ) {
     return match ($datastoreQuery->{"$.format"}) {
       'csv' => $this->streamCsvResponse($datastoreQuery, $result),
@@ -138,8 +138,6 @@ class QueryDownloadController extends AbstractQueryController {
    *
    * @param \Drupal\datastore\Service\DatastoreQuery $datastoreQuery
    *   A datastore Query object.
-   * @param \RootedData\RootedJsonData $result
-   *   Query result.
    *
    * @return \Symfony\Component\HttpFoundation\StreamedJsonResponse
    *   Return the StreamedResponse object.
@@ -191,10 +189,10 @@ class QueryDownloadController extends AbstractQueryController {
   /**
    * Stream selected metadata value from json object.
    *
-   * @param RootedJsonData $data
+   * @param \RootedData\RootedJsonData $data
    *   The json object.
    * @param string $metadata_name
-   *   The name of the metadata item to stream
+   *   The name of the metadata item to stream.
    */
   private function getJsonMetadata(RootedJsonData $data, string $metadata_name) {
     yield $data->get('$.' . $metadata_name);
--- a/modules/datastore/src/Controller/QueryDownloadController.php
+++ b/modules/datastore/src/Controller/QueryDownloadController.php
@@ -148,7 +148,10 @@ class QueryDownloadController extends AbstractQueryController {
     $response = new StreamedJsonResponse(
     // JSON structure with generator which will be streamed as a list.
       [
-        'results' => $this->loadJson($datastoreQuery, $result),
+        'results' => $this->loadJson($datastoreQuery),
+        'count' => $this->getJsonMetadata($result, 'count'),
+        'schema' => $this->getJsonMetadata($result, 'schema'),
+        'query' => $this->getJsonMetadata($result, 'query')
       ],
     );
     $response->headers->set('Content-Type', 'application/json');
@@ -166,7 +169,7 @@ class QueryDownloadController extends AbstractQueryController {
    * @param \RootedData\RootedJsonData $result
    *   Query result.
    */
-  protected function loadJson(DatastoreQuery $datastoreQuery, RootedJsonData $result) {
+  protected function loadJson(DatastoreQuery $datastoreQuery) {
     $count = 0;
 
     try {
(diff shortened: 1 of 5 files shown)


modules/datastore/src/Controller/QueryDownloadController.php
```diff
@@ -131,4 +133,56 @@ private function sendRow($handle, array $row) {
...
+   *   Return the StreamedResponse object.
+   */
+  protected function streamJsonResponse(DatastoreQuery $datastoreQuery, RootedJsonData $result) {
+    $response = new StreamedJsonResponse(
+    // JSON structure with generators in which will be streamed as a list.
+      [
+        'results' => $this->loadJson($datastoreQuery, $result),
+      ],
+    );
+    $response->headers->set('Content-Type', 'application.json');
```
dafeder:
```suggestion
    $response->headers->set('Content-Type', 'application/json');
```

modules/datastore/src/Controller/QueryDownloadController.php
```diff
@@ -131,4 +133,56 @@ private function sendRow($handle, array $row) {
...
 
+  /**
+   * Set up the Stream query result as json objects.
+   *
+   * @param \Drupal\datastore\Service\DatastoreQuery $datastoreQuery
+   *   A datastore Query object.
+   * @param \RootedData\RootedJsonData $result
+   *   Query result.
+   */
+  protected function loadJson(DatastoreQuery $datastoreQuery, RootedJsonData $result) {
```
dafeder: Can we move this below the streamJsonResponse() method, to keep the order more predictable?

modules/datastore/src/Controller/QueryDownloadController.php
```diff
@@ -131,4 +133,56 @@ private function sendRow($handle, array $row) {
...
+   *   A datastore Query object.
+   * @param \RootedData\RootedJsonData $result
+   *   Query result.
+   *
+   * @return \Symfony\Component\HttpFoundation\StreamedJsonResponse
+   *   Return the StreamedResponse object.
+   */
+  protected function streamJsonResponse(DatastoreQuery $datastoreQuery, RootedJsonData $result) {
+    $response = new StreamedJsonResponse(
+    // JSON structure with generators in which will be streamed as a list.
```
dafeder: Wording here is confusing - maybe just remove the word "in"?

dafeder (changes requested):
Some initial feedback here, still need to finish testing

dafeder (changes requested):
OK finished testing. It works! But now we need to add the rest of the response back and respect the arguments passed (`schema`, `count`, etc). 

Testing this also gave me some ideas for how to better handle count queries, but those are out of scope for this issue; I'll make another one.

modules/datastore/src/Controller/QueryDownloadController.php
```diff
@@ -131,4 +133,69 @@ private function sendRow($handle, array $row) {
...
+  /**
+   * Stream selected metadata value from json object.
+   *
+   * @param \RootedData\RootedJsonData $data
+   *   The json object.
+   * @param string $metadata_name
+   *   The name of the metadata item to stream.
+   */
+  private function getJsonMetadata(RootedJsonData $data, string $metadata_name) {
+    yield $data->get('$.' . $metadata_name);
```
dafeder: I wonder if this is really necessary? I understand the loadJson needs to have a yield to return the rows sequentially, but by the time we're returning these values it's a single return for each one. Couldn't we just add `'count' => $data->{"$.count"}` and so on?

modules/datastore/tests/src/Functional/Controller/QueryDownloadControllerTest.php
```diff
@@ -124,13 +124,30 @@ public function testDownloadWithMachineName() {
...
+      ['query' => ['format' => 'json']]
+    );
+
+    $json_content = json_decode($response->getBody()->getContents())->results[0];
+    $titles = array_keys(get_object_vars($json_content));
+    $this->assertEquals('id,name,extra_long_column_name_with_tons_of_characters_that_will_ne_e872,extra_long_column_name_with_tons_of_characters_that_will_ne_5127',
+      implode(',',$titles));
+//    $this->assertEquals('1', $json_content->id);
+//    $this->assertEquals('Greg', $json_content->name);
+//    $this->assertEquals('45.6', trim($json_content->extra_long_column_name_with_tons_of_characters_that_will_ne_e872));
```
dafeder: Remove commented-out code

dafeder (changes requested):
@beeyayjay great work, amazingly simple in the end. I wonder if we could redo the csv logic as well to be more consistent with this? Just a couple issues to look at.

modules/datastore/src/Controller/QueryDownloadController.php
beeyayjay: Makes sense.

modules/datastore/tests/src/Functional/Controller/QueryDownloadControllerTest.php
beeyayjay: Done

dafeder (approved):

PR #4591, opened 2025-10-29
WIP: Adds run operation for harvest plan entities

paul-m:
Fixes [issue#]

This is basically a proof-of-concept where I teach myself how to wire up entity operations.

### Todo:

Prettify the harvest_run entity display.
Handle a bunch of error cases.

### Discoveries:

Harvests can be 'reverted', which unpublishes the dataset nodes in various ways. This is different from reverting an entity revision, which is where you go back to old content from a previous edit. Drupal's entity system wants us to use 'revert' to mean the latter rather than the former. I got it working by callling it a rollback instead, but that's probably up for discussion.

Many permissions use an '[operation] [entity]' name, such as 'edit node', but the entity access system wants permissions to have a name like '[entity].[operation]', such as 'harvest_plan.run'. We'll need to rename permissions somewhere.

## Demo:

Make a local.
Enable and create sample content.
```
ddev drush cron
ddev drush cron
```
Visit the harvest page: https://dkan.ddev.site/admin/dkan/harvest
Click 'Run' on the sample_content harvest.
Confirm that you want to run the harvest.
Be redirected to the very ugly entity page for a harvest run entity.

## QA Steps

### DIFF
--- a/modules/harvest/src/ContentAccessControlHandler.php
+++ b/modules/harvest/src/ContentAccessControlHandler.php
@@ -11,37 +11,45 @@ use Drupal\Core\Session\AccountInterface;
 
 /**
  * Defines an access control handler for content entities.
+ *
+ * @see \Drupal\harvest\Entity\HarvestPlan
  */
 class ContentAccessControlHandler extends EntityAccessControlHandler {
 
   /**
-   * {@inheritdoc}
+   * {@inheritDoc}
    */
-  protected function checkAccess(EntityInterface $entity, $operation, AccountInterface $account): AccessResult {
-    return match($operation) {
-      'view' => AccessResult::allowedIfHasPermissions($account, [
-        'view ' . $this->entityTypeId,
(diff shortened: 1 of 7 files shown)


modules/harvest/src/ContentAccessControlHandler.php
```diff
@@ -11,37 +11,45 @@
...
-      ], 'OR'),
-      'delete' => AccessResult::allowedIfHasPermissions($account, [
-        'delete ' . $this->entityTypeId,
-        'administer ' . $this->entityTypeId,
-      ], 'OR'),
-      default => AccessResult::neutral(),
-    };
+  protected function checkAccess(EntityInterface $entity, $operation, AccountInterface $account) {
+    if ($admin_permission = $this->entityType->getAdminPermission()) {
+      return AccessResult::allowedIfHasPermission($account, $admin_permission);
```
paul-m, line 24: Oof. This is clearly wrong.

PR #4595, opened 2025-11-03
Less complex AbstractDatabaseTable::setTable()

dafeder:
In another PR, AbstractDatabaseTable::setTable() was flagged by qlty.sh as too complex. Cherry-picking into its own branch/PR.

### DIFF
--- a/modules/common/src/Storage/AbstractDatabaseTable.php
+++ b/modules/common/src/Storage/AbstractDatabaseTable.php
@@ -269,19 +269,19 @@ abstract class AbstractDatabaseTable implements DatabaseTableInterface {
    *   Throws an exception if the schema was not already set.
    */
   public function setTable() {
-    if (!$this->tableExist($table_name = $this->getTableName())) {
-      if ($schema = $this->schema) {
-        try {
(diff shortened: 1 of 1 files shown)


modules/common/src/Storage/AbstractDatabaseTable.php
```diff
@@ -269,19 +269,19 @@ protected function sanitizedErrorMessage(string $unsanitizedMessage) {
...
-          // their way out to the caller.
-        }
-      }
-      else {
-        throw new \Exception('Could not instantiate the table due to a lack of schema.');
-      }
+    if ($this->tableExist($table_name = $this->getTableName())) {
+      return;
+    }
+    $schema = $this->schema;
```
paul-m: No need for `$schema` variable.

paul-m (changes requested):

modules/common/src/Storage/AbstractDatabaseTable.php
dafeder: Huh, I have zero memory of why that's there. Thanks

paul-m (approved):

PR #4611, opened 2025-12-08
Fix rerun behavior in CircleCI

dafeder:
CircleCI now automatically re-tries the flaky cypress tests up to three times if they fail, but the second and third always fail because there's already data in the cypress/tmp dir. This deletes it before running the job.

Hopefully we can get a failing build on this PR to demonstrate.


dafeder:
Looks like this was implemented in #4634

PR #4613, opened 2025-12-08
#4612: Improve sanitizing column headers and description for datastore import

### LINKED ISSUE #4612: Improve sanitizing column headers and description for datastore import
### User Story

There is already sanitizing of colum headers and column description done in the datastore import to align with MySQL requirements.

I would like to have two additional cases be handled by the sanitation:

- put in a default string for an empty column name. If there is an empty column name, creating the MySQL table will fail because empty column name is not allowed.
- the column name might be a "binary string", which is not allowed for the description. Sanitize the description from binary string.

While both cases might indicate an invalid file structure or an encoding problem, this is not necessarily the case. So imho it would be better to have the file imported in these cases, and let the users see for themselves if the datastore import is usable rather than just failing the import (users do not directly see the reason for failed import?). But is surely opinionated, so asking for your opinion.

Extend the sanitation calls in [ImportJob](https://github.com/GetDKAN/dkan/blob/2.x/modules/datastore/src/Plugin/QueueWorker/ImportJob.php)

### Problem and Motivation

_No response_

### Target Audience

- [x] End Users
- [ ] Developers
- [x] Administrators
- [ ] Other

### Possible Implementation

will create an MR.

### Acceptance Criteria

_No response_

stefan-korn:
Fixes #4612 

## QA Steps

### DIFF
--- a/modules/datastore/src/Plugin/QueueWorker/ImportJob.php
+++ b/modules/datastore/src/Plugin/QueueWorker/ImportJob.php
@@ -119,15 +119,16 @@ class ImportJob extends AbstractPersistentJob {
   }
 
   /**
-   * Transform possible multiline string to single line for description.
+   * Sanitize column description (sanitize multiline and binary string)
    *
    * @param string $column
    *   Column name.
    *
    * @return string
-   *   Column name on single line.
+   *   sanitized column name on single line.
    */
   public static function sanitizeDescription(string $column) {
+    $column = preg_replace('~[^\x20-\x7E\t\r\n]~', '_', $column);
     $trimmed = array_filter(array_map('trim', explode("\n", $column)));
     return implode(" ", $trimmed);
   }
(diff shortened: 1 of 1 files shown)


modules/datastore/src/Plugin/QueueWorker/ImportJob.php
```diff
@@ -119,15 +119,16 @@ protected function __construct(string $identifier, $storage, ?array $config = NU
...
    *
    * @param string $column
    *   Column name.
    *
    * @return string
-   *   Column name on single line.
+   *   sanitized column name on single line.
    */
   public static function sanitizeDescription(string $column) {
+    $column = preg_replace('~[^\x20-\x7E\t\r\n]~', '_', $column);
```
janette, line 131: @stefan-korn seems like it would be better to just remove the BOM rather than replace it with an underscore.

PR #4615, opened 2025-12-08
#4614: Add resource id to Datastore Import-Status Dashboard

### LINKED ISSUE #4561: List resource ID instead of distribution ID in datastore import dashboard

### LINKED ISSUE #4614: Add resource id to Datastore Import-Status Dashboard
### User Story

It would be helpful to see the resource id for any given distribution in the Datastore Import-Status Dashboard (admin/dkan/datastore/status).

Currently the UUID of the distribution and the link to the file is provided.

In some places (for example drush dkan:datastore:drop and dkan:datastore:import) the resource id is relevant instead of the UUID. So it would be convenient to have this information in the Datastore Import-Status Dashboard.

Provide the resource id below the UUID and some labels to the informations to inform the users about.

### Problem and Motivation

_No response_

### Target Audience

- [x] End Users
- [ ] Developers
- [x] Administrators
- [ ] Other

### Possible Implementation

will provide an MR.

### Acceptance Criteria

_No response_

stefan-korn:
Fixes #4614
~Fixes #4561~

## QA Steps

### DIFF
--- a/modules/datastore/datastore.module
+++ b/modules/datastore/datastore.module
@@ -13,7 +13,6 @@ function datastore_theme($existing, $type, $theme, $path) {
     'datastore_dashboard_dataset_cell' => [
       'variables' => [
         'uuid' => NULL,
-        'resource_id' => NULL,
         'title' => NULL,
         'url' => NULL,
       ],
@@ -28,6 +27,7 @@ function datastore_theme($existing, $type, $theme, $path) {
     'datastore_dashboard_resource_cell' => [
       'variables' => [
         'uuid' => NULL,
+        'resource_id' => NULL,
         'file_name' => NULL,
         'file_path' => NULL,
       ],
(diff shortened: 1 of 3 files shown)


swirtSJW:
This one is ready for review, though we may want to wait until 4.x.

swirtSJW:
Looks like some tests need tweaking. ALso I was just informe that we actually do want both.  I will fix this up soon.

swirtSJW:
@dafeder I removed my commit so this is back to having both the resource id and the distribution id

PR #4617, opened 2025-12-10
Default to Drupal 10.5 in ddev-init and allow other constraints

chri5tia:
Fixes [issue#]
0000

## QA Steps

- [ ] clone a new, temporary dkan instance using documentation: https://dkan.readthedocs.io/en/latest/developer-guide/dev_local_setup.html
- [ ] run ddev drush status and see if Drupal instance is at least 10.5.8

### DIFF
--- a/ddev-init.sh
+++ b/ddev-init.sh
@@ -5,9 +5,9 @@
 # h/t @markdorison https://github.com/ddev/ddev-drupal-contrib/issues/15
 #
 # Usage: ddev-init [DRUPAL_VERSION]
-# Example: ddev-init 10.5
+# Example: ddev-init 10.5.*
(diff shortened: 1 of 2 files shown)


dafeder (changes requested):
@chri5tia this looks good, except that if I use, say, `ddev-init.sh ~10.5.0`, I get the error message "Project type must be one of backdrop, cakephp, craftcms, drupal, drupal6, drupal7, drupal8, drupal9, drupal10, drupal11...". I imagine this is because DRUPAL_MAJOR_VERSION is now `~10` so it's trying to do `--project-type drupal~10`.

janette (approved):

dafeder (approved):

PR #4619, opened 2025-12-11
Cache the select2 library in CI

dafeder:
We get failures sometimes when we run the `ddev select2` command in CI. I'm actually not sure why but could be a rate limit from github or something. CircleCI can cache dependencies based on a key, so we can take advantage of that. 

This also modifies the select2 ddev command to exit if the library is already there.

The cache key in circleCI is a hash of the command itself, since the command targets a specific commit, if we ever update that commit ID or anything else about the logic in the script it will invalidate the cache.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,6 +59,10 @@ commands:
                   # The following should be removed once the 2.x branch has the ddev add-on config.
                   git checkout $CIRCLE_BRANCH -- .ddev composer.json
 
+      - restore_cache:
+          keys:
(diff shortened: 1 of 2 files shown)


paul-m (approved):
CircleCI says this in its cache step:
```
Found a cache from build 54378 at select2-v1-S90Jw7rUBb_o9eaUbkDFV3qVpdBP5RRaGGisuZglEhA=
Size: 2.0 MiB
Cached paths:
  * /home/circleci/project/web/libraries/select2
```

PR #4628, opened 2025-12-17
Make LifeCycle::EVENT_PRE_REFERENCE support metadata alter

### LINKED ISSUE #4627: EVENT_PRE_REFERENCE does not support altering metadata in event subscriber
### Current Behavior

If I create an event subscriber for LifeCycle::EVENT_PRE_REFERENCE and in that subscriber I alter metadata for the dataset, it does not persist.  It gets overwritten because the reference uses the metadata from before the event is dispatched, rather than after any subscribers have run.

When [referenceMetadata() fires](https://github.com/GetDKAN/dkan/blob/b3545a9fca5dafc7323022f0d4e7062a81c85cfb/modules/metastore/src/LifeCycle/LifeCycle.php#L350)  it  extracts the $metadata, then dispatches the event.   Then my event subsciber alters the metadata inside $data. However then on line 361 it essentially saves the metadata that it set aside at the beginning, which is unchanged.  So all my metadata maniplulation never got saved.

### Expected Behavior

If I alter metadata as part of an event subscriber to LifeCycle::EVENT_PRE_REFERENCE, the alteration should persist and get saved with the referenced entity.

### Steps To Reproduce

1. Create an eventsubscriber to LifeCycle::EVENT_PRE_REFERENCE
2. Alter the dataset node's metadata in the subscriber.
3. Inspect the saved dataset and notice that the alteration did not get saved.
4. Curse the PHP gods for their abject cruelty.

```shell

```

### Anything else?

This will be testable on PDC.

swirtSJW:
Fixes #4627

## Describe your changes
The moves the metadata capture to after the event dispatch, to allow for pulling in changes that might have been made by a subscriber.

## QA Steps

- [ ] There is nothing to QA here as there is only a timing change if someone wants to use this event to alter metadata prior to referencing.

- [ ] I have updated or added tests to cover my code - No additional tests needed.  Nothing functionally changed for DKAN core, only how custom evensubscribers might interact.
- [ ] I have updated or added documentation - none needed.

### DIFF
--- a/modules/metastore/src/LifeCycle/LifeCycle.php
+++ b/modules/metastore/src/LifeCycle/LifeCycle.php
@@ -347,14 +347,13 @@ class LifeCycle {
    * @throws \Exception
    */
   protected function referenceMetadata(MetastoreItemInterface $data): void {
-    $metadata = $data->getMetadata();
-
     // Trigger datastore import if applicable.
     // Needs to happen before updating references.
     if ($data instanceof MetastoreItemInterface) {
       $event = new Event($data);
       $this->eventDispatcher->dispatch($event, self::EVENT_PRE_REFERENCE);
     }
+    $metadata = $data->getMetadata();
 
     // Convert references in metadata to uuids.
     // Create new reference entities if they do not exist.


swirtSJW:
Converting this to draft until I can prove it works on upper environments

swirtSJW:
See [my comment in the issue](https://github.com/GetDKAN/dkan/issues/4627#issuecomment-3712297448) as this more comes down to a choice in philosophy.  Should the event allow modification of metadata, or only be a reactive announcement.

PR #4638, opened 2026-01-05
Refactor metastore_admin submodules

### LINKED ISSUE #4565: Re-name metastore to dkan_metastore
In order to get DKAN to play nice with the rest of the drupal.org ecosystem, its submodules really need to be namespaced correctly. Rename metastore dkan_metastore (and rename its submodules similarly). Rename it and update references everywhere, but keep all functionality the same. Do not change service names, just namespaces.

We'd still like to merge it into dkan and common modules (#4413) but for now this will help resolve immediate issues.

dafeder:
Fixes #4565

Follow-up of #4610 to include metastore submodules. Reroll of #4632 #4625

Note: To get working config for the upgrade tests, I needed to add a line to the update hooks for any module that includes its own default config. For instance, if you were upgrading, your DKAN search_api index would be lost, because the original config had a dependency on metastore_search which is now disabled. If you export your config before upgrading then import afterwards, any customizations should be restored.

Relatedly, the api_user role, which is supplied by the dkan module itself, is lost in the update. So for now i've just removed it from the cypress tests. I propose we figure out the permanent solution for that later.


.circleci/config.yml
```diff
@@ -225,18 +225,6 @@ jobs:
...
-        default: "8.3"
-        type: string
-      database_version:
-        description: "Database version for DDev to use."
-        default: "mysql:5.7"
-        type: string
-      drupal_core_constraint:
-        description: "Branch of getdkan/recommended-project to use."
-        default: "~10.5.0"
-        type: string
```
dafeder, line 239: Result of a bad merge in 3.x

paul-m (approved):

PR #4640, opened 2026-01-05
Update README to fix local sandbox ddev link

janette:
## Describe your changes

As of DKAN 2.22, we no longer recommend using the ddev-dkan addon. For local development we are now using the [DDEV Drupal Contrib add-on](https://github.com/ddev/ddev-drupal-contrib) and the README should reflect that.

The quick setup video is now out-dated so removing the link to the youtube video.

## QA Steps

- [ ] Review the module README and confirm updates are accurate.


beeyayjay (approved):
LGTM!

PR #4641, opened 2026-01-06
Factor away Data::fixDataType

paul-m:
Related to [#4627]

### TLDR

- Removes `Data::fixDataType()`, so that it 'fixes' the data type on object construction, rather than continually querying it.
- This leaves only `saveRawMetadata()` in `fix()`, so we factor away `fix()`.
- We should revisit the need for `saveRawMetadata()` in later work.

### Explainer:

Every time `Data` does anything at all with a node, it calls `fix()`. `$this->fix()` is called 9 places within `Data`. For instance, `getIdentifier()` calls `fix()` which then calls `fixDataType()` and `saveRawMetadata()`, so that's at least two attached field gets on the node entity in order to get the entity's identifier.

The original plan for this PR was to remove `Data::fix()`, but there are various mysterious requirements related to `Data::saveRawMetadata()` needing to be called at weird times.

So instead, we remove `Data::fixDataType()`, so that it 'fixes' the data type on object construction, rather than continually updating it.

## QA Steps


modules/dkan_metastore/src/NodeWrapper/Data.php
```diff
@@ -87,7 +108,6 @@ public function getCacheMaxAge() {
    * @todo Needing to call fix() on every method seems like a code smell.
    */
   protected function fix() {
```
paul-m: - Remove `fix()` in favor of calling `saveRawMetadata()` directly.
- Check git blame for fix() to understand why it's protected and not private.

paul-m: `fix()` was protected so it could be mocked in `MetastoreAccessManagerTest`, and now `saveRawMetadata()` is, too.

dafeder (approved):

PR #4644, opened 2026-01-13
Attempt CircleCI builds w/o ddev or docker-in-docker

dafeder:
WIP to try to replicate the build w/o ddev. This could help with build times and compute, and make our general CI approach more portable w/o the need for docker in docker.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,6 +80,8 @@ commands:
             - web/libraries/select2
       - run:
           name: Install site
+          environment:
+            PATH: /home/circleci/project/vendor/bin:$PATH
           command: |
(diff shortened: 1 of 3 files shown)


dafeder:
Made some progress, was able to basically recreate the ddev poser/symlink workflow, and repurpose the parameters to chose images rather than ddev config. But right now drush/dkan-site-install aren't working. It could be because i'm trying to run the ddev-command as a plain bash script but i don't think that's the issue, it's something else with the environment.

dafeder:
Closing this - hopefully sometime this year we'll be moving to Drupal gitlab for CI and will revisit there.

PR #4645, opened 2026-01-13
Refactor harvest to dkan_harvest

dafeder:
Fixes #4414

PR #4646, opened 2026-01-13
Refactor harvest module to dkan_harvest

### LINKED ISSUE #4414: Rename harvest and datastore to dkan_harvest and dkan_datastore
These generic names were a bad call. Let's rip off the bandaid and rename them under a dkan_prefix as they are submodules and only (for now) relevant to DKAN.

## Acceptance criteria

* Let's avoid any refactoring not directly related to name changes
* Service names should be unaffected
* Use `git mv` as much as possible to maintain revision history
* Best possible upgrade path. As much as possible handled in update hooks, clear documentation on any manual steps needed. 

This almost certainly requires a 3.x release. Let's continue to refine the plan on how we'll manage branches, merge order and release plan.

dafeder:
Fixes #4414. Re-roll of #4645.

Note that in addition to refactors, there is new logic in the HarvestUtility to update namespaces for harvest plan load/transform types again.

I'll see what the reviewer says, but I think that qlty coverage issue is a weird false positive. It says there were 9 lines changes, 8 of which have no coverage, so it's an 11% coverage diff, even though overall coverage is unchanged? I think all the movement just sort of breaks the "diff coverage" model.

I do sort of wonder if we should also take the opportunity to move the Load and Transform folders from harvest/src into the ETL folder, since we're updating everything anyway. We didn't do that last time around (when we moved the harvest library into the module) to minimize disruption, but since all the namespaces are changing anyway...


paul-m (approved):

paul-m:
We can put off moving Load and Transform folders to another PR. This one already touches 109 files...

PR #4653, opened 2026-02-12
UploadOrLink submit fix

### LINKED ISSUE #4637: Drupal 11.3 support
Drupal 11.3 is now released, and DKAN should support it ASAP.

The two main issues I'm seeing:

1. A change in how the SQLite Connection class is instantiated when PHP 8.4 is present (easy fix)
2. Change to the core ManagedFile form element described [here](https://www.drupal.org/about/core/blog/native-htmx-in-drupal-1130-rich-ux-with-up-to-71-less-javascript), breaking the UploadOrLink element in Json Form Widget module (harder fix)

dafeder:
Fixes #4637

The remove button in JSON Form Widget's UploadOrLink element broke in Drupal 11.3. It turns out this is because we had an existing submit() method in that class, which now conflicts with the submit() method in the parent ManagedFile widget. Our submit was unrelated, it's a _form_ submit function that is added in a form_alter function in the module file. (Note, this is very inelegant, hopefully we can handle this differently somehow). What we needed was for the Remove button to have two submit handlers - the regular ManagedFile::submit() (previously just `file_managed_file_submit()` and then our `UploadOrLink::removeSubmit()`. We were suddenly hitting `UploadOrLink::submit()` instead of `ManagedFile::submit()`.

Changing the name of `ManagedFile::submit()` fixes the issue.

Note that, as this was the last blocking issue for Drupal 11.3, I have re-added that to the testing matrix.

## QA Steps

1. Build the branch against Drupal 11.3 (`ddev core-version ~11.3.0`).
2. Import sample content (`ddev dkan-sample-content`)
3. Edit a dataset. Remove one of its files and replace it with another file.
4. Save
5. Re-open the dataset. Confirm new file is still there.

Note, there is an issue that if you save the form without replacing the file, the old file is still there. But I believe this is a pre-existing bug.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -147,7 +147,7 @@ jobs:
             if [[ $CIRCLE_NODE_INDEX == 0 ]]; then
               # Get phpunit major version
               PHPUNIT_VERSION=$(ddev composer show phpunit/phpunit | grep versions | awk '{print $4}' | cut -d. -f1)
-              
+
               if [[ $PHPUNIT_VERSION -lt 11 ]]; then
                 # For phpunit < 11, use comma-separated syntax
                 EXCLUDES="--exclude-group "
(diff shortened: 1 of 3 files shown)


janette (approved):
@dafeder @beeyayjay This checks out, I was able to replace resources with uploads and other remote urls on an 11.3 build

PR #4656, opened 2026-02-13
#4655 Add link to node/123 to dataset import status table.

### LINKED ISSUE #4655: Add node link to Datastore Import-Status Dashboard
### User Story

As a datastore admin I need a way to go from the datastore import status table to the dataset so that I can get from one to the other easily.  

I am proposing we add a link below the other items in this cell that points to the node id route

red shows additional line entry

### Problem and Motivation

Unfortunately the link in the table points to the FE page for the dataset but not the dataset node.  So the workflow is to lookup the dataset name in the import status table, then go to the dataset list and filter by name.

link points to FE page not node.

### Target Audience

- [ ] End Users
- [ ] Developers
- [x] Administrators
- [ ] Other

### Possible Implementation

Suggesting just addition of another line in that cell that shows the node id and links to it. (node/123)

### Acceptance Criteria

A link appears in the import status that takes you to the node for the dataset.

swirtSJW:
Fixes #4655 

## QA Steps

- [ ] git checkout 4655-add-node-link-to-import-status
- [ ] drush cr to pick up twig changes
- [ ] visit /dashboard/datastore
- [ ] validate that the dataset identifier comes after/below the title
- [ ] validate that the dataset title has no link
- [ ] validate the "view" links to the FE version of the dataset page
- [ ] validate the "admin" links to the node version of the dataset page
- [ ] inspect the source and validate that the "admin" link has an aria-label that reads "admin: <dataset title>"
- [ ] inspect the source and validate that the "view" link has an aria-label that reads "view: <dataset title>"

### DIFF
--- a/modules/datastore/tests/src/Unit/Form/DashboardFormTest.php
+++ b/modules/datastore/tests/src/Unit/Form/DashboardFormTest.php
@@ -155,6 +155,7 @@ class DashboardFormTest extends TestCase {
     $this->assertEquals('NEW', $form['table']['#rows'][0][2]['data']);
     $this->assertEquals('done', $form['table']['#rows'][0][6]['data']['#status']);
     $this->assertEquals(NULL, $form['table']['#rows'][0][6]['data']['#error']);
+    $this->assertEquals(1, $form['table']['#rows'][0][0]['data']['#node_id']);
   }
 
   /**
@@ -224,6 +225,7 @@ class DashboardFormTest extends TestCase {
     $this->assertEquals('NEW', $form['table']['#rows'][0][2]['data']);
     $this->assertEquals('done', $form['table']['#rows'][0][6]['data']['#status']);
     $this->assertEquals(NULL, $form['table']['#rows'][0][6]['data']['#error']);
+    $this->assertEquals(2, $form['table']['#rows'][0][0]['data']['#node_id']);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


modules/datastore/tests/src/Unit/Form/DashboardFormTest.php
```diff
@@ -2,36 +2,36 @@
 
 namespace Drupal\Tests\datastore\Unit\Form;
 
+use Drupal\common\DataResource;
+use Drupal\common\DatasetInfo;
+use Drupal\Core\Database\Connection;
```
swirtSJW: These "use" moves were all just sorts, nothing added or removed.

dafeder (changes requested):
Code itself is fine, but I'm not sure why you added the node_id to the data in all the tests but then never have assertions for it?

swirtSJW:
> Code itself is fine, but I'm not sure why you added the node_id to the data in all the tests but then never have assertions for it?

I had to add the node_id to the tests because otherwise the template errors out if the node_id does not exist in the clall to path().

I can add additional assertions to make sure that the twig path() works, but that seems excessive to test that a core twig function does its job.   That is all it would test since it blows up if you feed it something that is not a node id.

dafeder (approved):
Approved, but holding off on merging until 4.x

swirtSJW:
Closing this in favor of #4752 which is the same code, but applied against 4.x instead of 2.x.

PR #4657, opened 2026-02-15
#4621 Add revision time on presave.

### LINKED ISSUE #4621: Publishing a metastore does not change revision date
### Current Behavior

In Metastore Data.php the function setWorkflowState() does not add a revision date.  This creates a confusing situation where a draft made weeks ago and a publish today shows the two revisions with the same date.

revisions weeks apart end up with same timestamp

This comes from an odd drupalism where if a revision has not timestamp set it uses the timestamp from the previous revision, creating confusion throughout the land.

Calling 
`$entity->setRevisionCreationTime(<timestamp here>); `
right before the $entity->save() will resolve this.

Time stampery will need something like \Drupal::time()->getCurrentTime()

### Expected Behavior

- [ ] When a dataset publishes or goes through some other status change, the revision list tells the true story.

### Steps To Reproduce

1. Save a datset in draft.
2. wait 10 minutes
3. Publish it using the metastore publish function
4. View the revision log.  shows the same date and time.

swirtSJW:
Fixes #4621 

## Describe your changes
This explicitly sets data node revision time so that programatic changes like publish reflect the time of the publish, and don't repeat the time of the draft.
This results in revisions that look like
different revision dates

as opposed to the current  situation of 

confusing same revision dates

## QA Steps

- [ ] Save a a dataset as draft.
- [ ] publish it using either a drush command or a bulk operation in the UI.
- [ ] Visit the revisions tab for the dataset. validate that the first line  for each of the recent revisions has its own date and time. (rather than repeating the time.)
- [ ] validate that the revision message, which also looks like a time stamp did not change should still say something like "Created on <date>" and the date should match the message from the draft revision.

### DIFF
--- a/modules/metastore/metastore.module
+++ b/modules/metastore/metastore.module
@@ -5,12 +5,13 @@
  */
 
 use Drupal\Core\Database\Query\AlterableInterface;
-use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
 use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\Core\Form\FormStateInterface;
-use Drupal\metastore\NodeWrapper\Data;
 use Drupal\metastore\MetastoreService;
+use Drupal\metastore\NodeWrapper\Data;
+use Drupal\metastore\Factory\MetastoreEntityItemFactoryInterface;
 use Drupal\metastore\Service\OrphanNodeProcessor;
 
 /**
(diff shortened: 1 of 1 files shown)


swirtSJW:
If the timing of this is a problem, it can wait until V3.x. because I am using the patch where I need it.

swirtSJW:
Proof of life with the latest addition of using the class to the the target entity and targed bundles
revisions with different timestamps.

dafeder (changes requested):
Needs a functional test. Otherwise looks good. I can possibly add the test myself.

swirtSJW:
Closing this in favor of #4753 which is applied to the 4,x branch.

PR #4658, opened 2026-02-15
#4622 Add validation killswitch

### LINKED ISSUE #4622: Consider temporary json validation kill switch
### User Story

There are rare times when distribution validation can leave things in a very stuck state.  Can't save, can't load, can't edit.   As much as we hold data validation as a prime directive, there are times when being stuck is bad.  It would be helpful if there was a way to temporarily turn off the validation (either through the UI or a drush command) so that we could get things un-stuck.

A way to toggle validation off would provide a way to get out of jam.

### Problem and Motivation

_No response_

### Target Audience

- [ ] End Users
- [x] Developers
- [x] Administrators
- [ ] Other

### Possible Implementation

Possible approaches ranging from simple & scary to more challenging & safer.

1. A UI option in Metastore to toggle validation on or off. 
2. A drush command to toggle validation on or off
3. 1 or 2 ^^ with a cron job that runs every 30 minutes that toggles validation back on.
4. 1 or 2 ^^ with a status message that screams at admins with every page load that warns this is a "Very Very Very bad idea." to shame people into not using it for very long.
5. A UI to add a list of dataset IDs to be exempt from validation.
### Acceptance Criteria

_No response_

swirtSJW:
Fixes #4622 

## QA Steps

### DIFF
--- a/modules/metastore/src/Plugin/Validation/Constraint/ProperJsonValidator.php
+++ b/modules/metastore/src/Plugin/Validation/Constraint/ProperJsonValidator.php
@@ -2,6 +2,7 @@
 
 namespace Drupal\metastore\Plugin\Validation\Constraint;
 
+use Drupal\Core\Config\ConfigFactoryInterface;
 use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
 use Drupal\metastore\ValidMetadataFactory;
 use OpisErrorPresenter\Implementation\MessageFormatterFactory;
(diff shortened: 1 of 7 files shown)


modules/metastore/src/Plugin/Validation/Constraint/ProperJsonValidator.php
```diff
@@ -60,8 +72,22 @@ public static function create(ContainerInterface $container) {
...
+    }
+  }
+
+  /**
+   * Validates items and adds violations if needed.
+   *
+   * @param object|mixed $items
+   *   Items to validate.
+   */
+  protected function validateItems($items): void {
```
swirtSJW, line 88: I did not really want to break this out into a separate function, but Qlty had opinions.  Adding the one if wrapper brought it up to 6 layers deep.

too many layers

swirtSJW:
@dafeder raised in Slack that he has concerns about this disabling all validation and suggests that it be scoped to only disabling data read/load, not data write.  I think that sounds reasonable.

swirtSJW:
Closing this as it is replaced by #4754

PR #4662, opened 2026-02-17
#4661 Add empty, not_empty filters to dataset table

### LINKED ISSUE #4661: Dataset filter should support empty and not_empty filters
### User Story

ON PDC we support dataset filters that include "is empty" and "not empty"

Image

We should support these additional filters if they are implemented on the FE component

### Problem and Motivation

Sometime you want to look for all rows that have no value for a specific field or something you want to look for all rows that have data in some field.

### Target Audience

- [x] End Users
- [ ] Developers
- [ ] Administrators
- [ ] Other

### Possible Implementation

_No response_

### Acceptance Criteria

The api for the dataset table supports filters for empty and not empty

swirtSJW:
Fixes #4661

## Describe your changes
Needs the following
- [ ] account for 'empty'
   - [x]  empty string
   - [ ] null
- [ ] account for 'not empty' 
   - [x]  not empty string
   - [ ] not null
- [ ] Expansion of tests
## QA Steps

### DIFF
--- a/modules/datastore/docs/query.json
+++ b/modules/datastore/docs/query.json
@@ -204,13 +204,15 @@
                                 "<",
                                 "<=",
                                 ">",
-                                ">="
+                                ">=",
+                                "not_empty",
+                                "is_empty"
                             ]
                         },
                         {
                             "type": "string",
-                            "pattern": "^([lL][iI][kK][eE]|[bB][eE][tT][wW][eE][eE][nN]|[iI][nN]|[nN][oO][tT] [iI][nN]|[cC][oO][nN][tT][aA][iI][nN][sS]|[sS][tT][aA][rR][tT][sS] [wW][iI][tT][hH]|[mM][aA][tT][cC][hH])$",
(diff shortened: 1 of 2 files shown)


swirtSJW:
This can wait until after 3.x.  I have the patch I need from it.

modules/datastore/docs/query.json
```diff
@@ -204,13 +204,15 @@
                                 "<",
                                 "<=",
                                 ">",
-                                ">="
+                                ">=",
+                                "not_empty",
```
swirtSJW, line 208: These should not be here, only in regex pattern below.

swirtSJW: clarify in the ODC ticket that vlaue should be no undersores

swirtSJW: Maybe support is null and not null in a future iteraation

swirtSJW:
I am closing this in favor of a FE only approach 
WCMS-27708

PR #4667, opened 2026-03-02
3.x config event subscriber

dafeder:
Handles issues importing config that was exported before the upgrade to 3.x

## QA Steps

- [ ] Upgrade an existing site to `2.x-transition`
- [ ] Export config
- [ ] Upgrade to `config-event-subscriber`
- [ ] `drush updatedb`
- [ ] Compare staged config with existing config in UI

We expect some discrepancies no matter what but should generally be able to explain the diffs and see incoming config equivalent to what it was pre-upgrade. Any config items that were lost in the update should come back in with updated dependencies.

This is kind of an inherently fuzzy part of the upgrade path, just hoping to get it as good as possible and maybe come back for a second pass after some real world upgrade dry-runs.

### DIFF
--- a/modules/dkan_common/dkan_common.services.yml
+++ b/modules/dkan_common/dkan_common.services.yml
@@ -45,6 +45,11 @@ services:
     parent: logger.channel_base
     arguments: [ 'dkan' ]
 
+  dkan.common.config_import_rename_dependencies_subscriber:
+    class: Drupal\dkan_common\EventSubscriber\ConfigImportRenameDependenciesSubscriber
+    tags:
+      - { name: event_subscriber }
+
   # @deprecated
   common.docs:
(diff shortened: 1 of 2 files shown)


dafeder:
Thinking about how to write tests for this. Since it's only going to be included in one release and removed in the 4.x branch, not sure it's necessary?

paul-m (approved):

stefan-korn:
> Since it's only going to be included in one release and removed in the 4.x branch, not sure it's necessary?

It is still present in 4.x? https://github.com/GetDKAN/dkan/blob/4.x/modules/dkan_common/src/EventSubscriber/ConfigImportRenameDependenciesSubscriber.php - Are there plans to remove in 4.x in near future?

PR #4668, opened 2026-03-03
3.x upgrade docs

dafeder:
Upgrade instructions, plus a few tweaks to the docs while I'm at it.

## QA Steps

* Read https://dkan--4668.org.readthedocs.build/en/4668/upgrade.html
* Review other changes in diff


docs/source/upgrade.rst
```diff
@@ -0,0 +1,210 @@
...
+
+* **DKAN 2.23** will introduce the new, renamed modules (they will not yet be functional)
+* **DKAN 3.0** will move all the code from the new modules to the old, provide all 
+  the update functions to migrate existing sites, and deprecate the old modules.
+  We do not plan to release any updates to 3.0, as it is intended purely as a
+  transition to DKAN 4.0.
+* **DKAN 4.0** will remove the legacy modules from the codebase. This will also
+  be the first release on drupal.org, and the package name in composer will change
+  from ``getdkan/dkan`` to ``drupal/dkan``.
+
```
swirtSJW:
```suggestion

**Warning:** Due to limitations in how database and config updates are performed by Drupal,  it will not be possible to upgrade directly from 2.23 to 4.  It **must** be done in separate upgrades to 2.23, then to 3.0, and finally to 4.x.

```
I know you say this in much more detail farther down, but it feels like this should be establish upfront.

docs/source/upgrade.rst
```diff
@@ -0,0 +1,210 @@
...
+and new features exclusively on DKAN 4.x.
+
+Detailed instructions
+---------------------
+
+Summary
+#######
+
+We err on the side of over-explaining in the sections below, so the process may
+seem intimidating. However, the process is fairly straightforward:
```
swirtSJW:
```suggestion
Due to limitations in how database and config updates are performed by Drupal,  
it will not be possible to upgrade directly from 2.23 to 4.  It **must** be done 
in separate upgrades.   The process may seem intimidating. However, the process
 is fairly straightforward:
```

swirtSJW: This has nothing to do with caution, it simply will not work any other way. :)

docs/source/upgrade.rst
```diff
@@ -0,0 +1,210 @@
...
+of the Drupal documentation.
+
+Step 3: Get to DKAN 3.x
+#######################
+
+Now you should be ready to upgrade the DKAN module. Run:
+
+.. prompt:: bash $
+
+    composer require getdkan/dkan:~3.0.0
```
swirtSJW: Is another updb needed here?  And also a config export?

docs/source/upgrade.rst
```diff
@@ -0,0 +1,210 @@
...
+These may have references to either module names or class namespaces that need to be updated.
+
+Step 5: Run updates
+###################
+
+You now have a 3.x codebase but your database has not changed since 2.x. Run:
+
+.. prompt:: bash $
+
+    drush update:db
```
swirtSJW, line 180: And config export?

docs/source/upgrade.rst
```diff
@@ -0,0 +1,210 @@
...
+
+.. note::
+
+    The exact process by which ``drupal/dkan`` will replace ``getdkan/dkan`` is
+    a bit TBD. As of the writing of these docs, we have not released yet on
+    drupal.org, and have therefore not tested the exact composer workflow yet.
+
+While there should be no pending database updates at this point, it is always a
+good idea to run ``drush update:db`` after any change to ``composer.json``.
```
swirtSJW: This all looks good as it describes the local upgrade process.  However it does not really address the deploy process, which is not really the same.  We may want to think about suggesting in the above steps when they should create tags at each step.



```suggestion
good idea to run ``drush update:db`` after any change to ``composer.json``.

Deploying DKAN upgrades
####################

#. backup your database
#. deploy tag DKAN-2-upgrade-tag
#. drush deploy
#. deploy tag DKAN-3-upgrade-tag
#. drush deploy
#. deploy tag DKAN-4-upgrade-tag
#. drush deploy
```

docs/source/upgrade.rst
dafeder: Not sure what you're referring to here - the "caution" callout is about taking backups at each step not the multi-step upgrade. You could do this without taking backups; restoring a backup is not an actual step in the process. If you're doing this locally and pulling a db from prod you _could_ just do that every time you need to roll back.

docs/source/upgrade.rst
dafeder: I put that after the "refactor custom code piece" because I suspect the update will fail if you haven't done that yet.

docs/source/upgrade.rst
dafeder: See step 6 Your config is now almost certainly incomplete/broken, exporting will blow away things you need. You need to import the 2.x config to 3.x, which triggers the config import event subscriber, before you can safely export again.

docs/source/upgrade.rst
dafeder: Yeah I want to think about this more carefully. As you know I'm a `drush deploy` skeptic and not sure I want that in the docs, at least not without alternatives. Since this isn't being released yet, how about we note for now just that these instructions are environment-agnostic and we may add some suggestions for deployment startegies in the future before 3.x and 4.x are released?

swirtSJW: I agree it is hard to make assumptions about environment strategies, but I do think at some point we have to make it clear that you can't do all this config prep locally and then just deploy version 4.  If somebody does that they end up with a bricked site.

dafeder: Yes agreed, working on some deployment notes to pepper throughout the steps here.

docs/source/upgrade.rst
```diff
@@ -0,0 +1,260 @@
...
+
+The correct modules should now be enabled and disabled, but your site config is
+now in a weird state. Likely, several configurations have been automatically
+removed because their dependencies on legacy modules are no longer met.
+Fortunately, your original configuration is preserved in your sync directory if you followed step 2.
+
+Before you import, you can inspect the difference between your current site
+config and the sync directory by visiting the Drupal UI at ``/admin/config/development/configuration``.
+You will possibly see a number of configuration items being added, because they
+were removed the legacy modules were disabled.
```
janette: Should this last line be "...were removed **when** the legacy modules were disabled."

dafeder:
Actually, hold off on this I'm thinking maybe this should go straight to 4.x instead 🤔

dafeder:
Closing, will re-roll as PR against 4.x branch.

PR #4675, opened 2026-03-09
Fix upgrade path after 2.x rebase

dafeder:
This PR fixes some issues after the merge of the `2.x-transition` branch into `2.x` and the addition of Drupal 11.3 to the testing matrix.

The rename of all update hooks to be > 11300 turned out not to help much, but was retained because it seemed like it made sense. That's the current Drupal release as of the impending 3.x release.

We add a second dump and logic to select it depending on what version is is use, as it no longer works to upgrade directly from Drupal 10.5 to 11.3 at the same time one upgrades from DKAN 2.x to 3.x. The upgrade instructions being developed over in #4065 will be updated to reflect this.


.gitignore
```diff
@@ -11,6 +11,7 @@ tagfile
 /cypress/videos
 /cypress/screenshots
 /libraries
+/modules/sample_content/sample_content.json
```
swirtSJW, line 14: Just doublechecking that this is not going in the wrong direction, ignoring both old and new?

swirtSJW (approved):
All changes make sense to me. (not tested)

.gitignore
dafeder: Well this is mostly for me, I'm always switching branches and having to delete that file because it got created by the 2.x module but is no longer ignored. Really it's a bad pattern to have it written to the module filesystem but it is what it is for now.

PR #4679, opened 2026-03-26
Pass resource_version when re-importing a resource

janette:
## Describe your changes
Sometimes resources can fail to re-import if the resource_version is not passed to DatastoreService::import

### DIFF
--- a/modules/dkan_datastore/src/Commands/ReimportCommands.php
+++ b/modules/dkan_datastore/src/Commands/ReimportCommands.php
@@ -89,7 +89,7 @@ class ReimportCommands extends DrushCommands {
       if ($resource_id = $distribution['resource_id'] ?? FALSE) {
         $this->datastoreService->drop($resource_id);
         $this->logger()->notice('Reimporting distribution: ' . $resource_id);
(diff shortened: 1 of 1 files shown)


modules/dkan_datastore/src/Commands/ReimportCommands.php
```diff
@@ -89,7 +89,7 @@ protected function reimportDistributions(array $distributions) {
       if ($resource_id = $distribution['resource_id'] ?? FALSE) {
         $this->datastoreService->drop($resource_id);
         $this->logger()->notice('Reimporting distribution: ' . $resource_id);
-        $result = $this->datastoreService->import($resource_id);
+        $result = $this->datastoreService->import($resource_id, FALSE, $distribution['resource_version']);
```
dafeder:
```suggestion
        $result = $this->datastoreService->import($resource_id, FALSE, $distribution['resource_version'] ?? NULL);
```
Maybe just to be on the safe side here?

dafeder (changes requested):

dafeder (approved):

PR #4689, opened 2026-04-03
Use fulltext keys correctly

dafeder:
The metastore search uses a strange approach to fulltext searches... it looks at the index, gathers all the fields, and creates an OR group of conditions matching the query text against each one. Changing the code to simply use `keys()` to add fulltext conditions is much more in line with search_api and provides the option of including quoted phrases.

## QA Steps

1. Fresh build
2. `ddev dkan-sample-content`
3. Request https://dkan.ddev.site/api/1/search?fulltext=Election%20Districts%20Afghanistan (title phrase, out of order). You should get one result
4. Compare to a previous version: https://demo.getdkan.org/api/1/search?fulltext=Election%20Districts%20Afghanistan - you should see no results.
5. Now try with quotes - https://dkan.ddev.site/api/1/search?fulltext=%22Election%20Districts%20Afghanistan%22 - you should get no results.
6. With quotes and words in the right order, you should get results: https://demo.getdkan.org/api/1/search?fulltext=%22Afghanistan%20Election%20Districts%22

### DIFF
--- a/modules/dkan_metastore/modules/dkan_metastore_search/src/QueryBuilderTrait.php
+++ b/modules/dkan_metastore/modules/dkan_metastore_search/src/QueryBuilderTrait.php
@@ -58,7 +58,7 @@ trait QueryBuilderTrait {
     if (!isset($params['fulltext']) || empty($params['fulltext']) || empty($index->getFulltextFields())) {
       return [$query, FALSE];
     }
- 
+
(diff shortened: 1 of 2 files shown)


dafeder:
Wait - I think I'm calling keys() wrong, hold off on review

dafeder:
Better

janette (approved):

PR #4697, opened 2026-04-20
Handle exceptions within harvest revert.

paul-m:
Repro:

- Run a harvest to populate Data nodes for datasets.
- Delete one of the dataset nodes. Visit the datasets page and delete it using the UI.
- Try to revert the harvest: `drush dkan:harvest:revert YOUR_HARVEST`
- There will be an error that one of the datasets cannot be removed because it does not exist.
- No other datasets after that one will be removed.
- Run the harvest again: `drush dkan:harvest:run YOUR_HARVEST`
- The status will show that all datasets are unchanged.

There is no way around this, if you want to use harvests to manage the datasets after you've deleted one. The harvest system thinks you still have all the datasets, even though one is missing.

The solution is to handle the dataset not found exception, and let the harvest revert continue to run. This way all the hashes will be removed, which signals to the harvest system that the datasets have not been created.

Fixes [issue#]

## QA Steps


dafeder:
Needs test

PR #4700, opened 2026-04-28
Unset downloadURL if empty string

dafeder:
This adds a setting to the metastore to remove the downloadURL property from a distribution, if it is set to an empty string. This may be desirable because a) removing a resource referenced in downloadURL results in an empty string, and b) downloadURL is not required, but must be a URL if present. When a dataset already saved in the datastore becomes invalid, it can no longer be loaded, even to correct it.

Note this also restores the OrphanNodeProcessorTest, which appears to have been removed accidentally during the big refactor.

## QA Steps

- [ ] Build site, and load sample content
- [ ] `drush dkan:dataset-info cedcd327-4e5d-43f9-8eb1-c11850fa7c55`
- [ ] Copy resource_id (should be `3a187a87dc6cd47c48b6b4c4785224b7`)
- [ ] `drush sqlc`
- [ ] `DELETE FROM dkan_metastore_resource_mapper WHERE identifier = '3a187a87dc6cd47c48b6b4c4785224b7' AND perspective = 'source';`
- [ ] `exit`
- [ ] `drush cr`
- [ ] Load https://dkan.ddev.site/api/1/metastore/schemas/dataset/items/cedcd327-4e5d-43f9-8eb1-c11850fa7c55
- [ ] Confirm JSON validation error
- [ ] Visit https://dkan.ddev.site/admin/dkan/properties
- [ ] Check "Unset download URL if empty" and Save
- [ ] Re-load https://dkan.ddev.site/api/1/metastore/schemas/dataset/items/cedcd327-4e5d-43f9-8eb1-c11850fa7c55
- [ ] Confirm it loads correctly, and does not have a downloadURL in the distribution

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -285,13 +285,13 @@ workflows:
       - phpunit:
           matrix:
             parameters:
-              drupal_core_constraint: ["~10.6.0", "~11.2.0", "11.3.0"]
+              drupal_core_constraint: ["~10.6.0", "~11.2.0", "~11.3.0"]
               php_version: ["8.3"]
               database_version: ["mariadb:10.11"]
       - phpunit:
           matrix:
             parameters:
-              drupal_core_constraint: ["10.6.0", "~11.2.0", "~11.3.0"]
+              drupal_core_constraint: ["~10.6.0", "~11.2.0", "~11.3.0"]
               php_version: ["8.4"]
               database_version: ["mariadb:10.11"]
       - phpunit:
--- a/modules/dkan_metastore/config/schema/dkan_metastore.schema.yml
+++ b/modules/dkan_metastore/config/schema/dkan_metastore.schema.yml
@@ -42,3 +42,6 @@ dkan_metastore.settings:
         retain_for:
           type: integer
           label: 'Number of days to keep referenced content before deletion'
+    unset_download_url_if_empty:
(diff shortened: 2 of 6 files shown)


dafeder:
I still imagine we would switch to YAML-defined schemas with a bunch of options some day (#3761) that might replace this -- ie, any property in a schema could be set to `remove_if_empty` or something.

janette:
@dafeder this is great. On sites with hundreds of datasets, we'll need a way to find the dataset(s) that are missing downloadURLs. Logging a warning would be nice (enable dblog on lower envs), otherwise you would need to pull the DB to a local build, uncheck the setting, and use xdebug to stop on the validation exception. Or maybe add a sort to the resource column on the import status page.

janette (approved):

stefan-korn:
@dafeder : We are experiencing an issue where several (not all) orphaned distributions cannot be deleted, because of "empty" downloadURL (there is something in resource mapping, but it seems DKAN cannot make an URL of it). Using the setting introduced here, we are able to delete the distribution.

As far as I can see, it would not hurt to enable this setting.

Or are there actual risks you see with this setting enabled? Asking, because help text is kind of warning 

> "...Leave this unchecked unless you are experiencing fatal validation errors on dataset load."

PR #4701, opened 2026-04-30
Post import cache invalidate

dafeder:
The dataset query endpoint for the datastore api (/api/1/datastore/query/[dataset_id]/[index]) does not get a cache tag for the dataset ID, just the distribution. This causes the cache to get stuck in many situations, because the distribution IDs change over time. If a distribution ID changes in the parent dataset, invalidating caches based on the old distribution ID no longer works. Tagging URLs that contain a dataset ID with that dataset's node cache tag, plus directly invalidating both cache tags when an import is finished, result in much more predictable cache behavior.

## Notes

* Copied approach from MetastoreApiPageCacheTest but implemented local files for the tests. In a separate PR will apply that approach to the original test file. I didn't realize we still had S3 dependencies in our tests; would like to finally eliminate all of those.
* The different methods in AbstractQueryControler are similar but couldn't be condensed as much as I'd hoped, at least not in a way I could see that would preserve the signatures. I think what I did makes it a little less redundant and more readable?
* We could probably create a trait or base class for both tests since they share several functions
* I don't understand what the `::renderDatasetNodesForCache()` function does exactly, or why its needed. Is this an accurate simulation of the real-world behavior?

### Reproduce old behavior

1. Build on 4.x
2. Create a dataset with a CSV upload resource. You can use one of the csvs in modules/dkan_datastore/tests/data
3. Run queues/cron. Confirm that /api/1/datastore/query/[dataset-id]/0 shows expected results.
4. Edit dataset. Replace CSV with a new CSV.
5. Refresh the datastore API URL from 3. Confirm that it still shows the old data.
6. Run the localize_import and datastore_import queues
7. Confirm that datastore API is still stuck on the same response as #3.
8. Run the orphan_reference_processor queue. 
9. Now the datastore endpoint should finally show the new data when refreshed, because the orphan reference processor modified the old distribution, causing the distribution cache tag to invalidate.

### Demonstrate fix

1. Build on `post-import-cache-invalidate`
2. Create a dataset with a CSV upload resource. You can use one of the csvs in modules/dkan_datastore/tests/data
3. Run queues/cron. Confirm that /api/1/datastore/query/[dataset-id]/0 shows expected results.
4. Edit dataset. Replace CSV with a new CSV.
5. Refresh the datastore API URL from 3. It should now show a 400 error, as there has been a new file added but it's not yet imported
6. Run the localize_import queue
7. Refresh datastore, confirm there is still a 400 error
8. Run the datastore_import queue
9. Refresh the datastore endpoint. Confirm you now see the new expected data.

### DIFF
--- a/modules/dkan_datastore/src/Controller/AbstractQueryController.php
+++ b/modules/dkan_datastore/src/Controller/AbstractQueryController.php
@@ -12,6 +12,7 @@ use Drupal\dkan_metastore\MetastoreApiResponse;
 use JsonSchema\Validator;
 use RootedData\RootedJsonData;
 use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\ParameterBag;
 use Symfony\Component\HttpFoundation\Request;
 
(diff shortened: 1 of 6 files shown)


paul-m:
Did all the testing steps... LGTM.

paul-m (approved):

PR #4702, opened 2026-05-01
Add date to filtered csv exports

### LINKED ISSUE #4722: Download dataset file should be specifically named.
### User Story

When I download a dataset, it would be nice to have it named meaningfully.

When you download a dataset you should be able identify the dataset and the date it came from using a pattern that includes the ID and the modified date.
```
dataset_<dataset-identifier>_<index>_<modified-date>.csv
```
dataset_fgwlstr_1_2026-03-21.csv

In the event that the download is filtered, it should also include that
```
dataset_<dataset-identifier>-<index>_<modified_date>_filtered.csv
```
dataset_fgwlstr_1_2026-03-21_filtered.csv

### Problem and Motivation

Currently when you download a dataset, it gets named "dataset.csv"   Download another it become 'dataset(1).csv". It is impossible to tell what is what without opening it.

### Target Audience

- [x] End Users
- [ ] Developers
- [ ] Administrators
- [ ] Other

### Possible Implementation

See feature description

### Acceptance Criteria

When downloading a csv for a dataset, the filename has information to help me identify it.

janette:
## Describe your changes
When a user creates a filter on a dataset, they have an option to download the subset to a csv file. Currently that file is named data.csv. This will append a date and time to help differentiate it from other downloads.

## QA Steps

- [ ] Navigate to a dataset with a CSV distribution.
- [ ] Create a filter on the data
- [ ] Hit the download API and confirm the name of the file includes the date and time.

### DIFF
--- a/modules/dkan_datastore/src/Controller/QueryDownloadController.php
+++ b/modules/dkan_datastore/src/Controller/QueryDownloadController.php
@@ -11,6 +11,7 @@ use RootedData\RootedJsonData;
 use Symfony\Component\HttpFoundation\ParameterBag;
 use Symfony\Component\HttpFoundation\StreamedJsonResponse;
 use Symfony\Component\HttpFoundation\StreamedResponse;
+use Drupal\Core\Datetime\DrupalDateTime;
 
 /**
  * Controller providing functionality used to stream datastore queries.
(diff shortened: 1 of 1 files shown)


dafeder:
@swirtSJW filed an issue about this, and we discussed maybe using the resource filename (or maybe an excerpt of, say, 12 or 14 characters if it's long) to better identify it. Can we look in to whether that's viable -- not sure if we have the context there in the controller to load the resource object 🤔

PR #4704, opened 2026-05-04
Send back 503 on query requests if the database is otherwise occupied.

paul-m:
Fixes [issue#]

## Describe your changes

On a client project, we got some interesting watchdog logs, sorting them by frequency. The log represented a week of data.
The top entry is this one:
```
[HY000] [1203] User u0hFhL5uTb1qLmLI already has more than 'max_user_connections' active connections in Drupal\common\Storage\AbstractDatabaseConnectionFactory->getConnection() (line 81 of /var/www/html/docroot/modules/contrib/dkan/modules/common/src/Storage/AbstractDatabaseConnectionFactory.php).
```
This had 4,285 occurrences within a week, which is 612 a day.

This means we can handle this error better, and turn it into a useful 429 response.

This PR handles this within the `AbstractQueryController`, which is where it is most likely to occur (at least on our site).

We could place similar catch blocks on other controllers as well, or figure out a more centralized way to handle this.

## QA Steps

### DIFF
--- a/modules/dkan_common/src/JsonResponseTrait.php
+++ b/modules/dkan_common/src/JsonResponseTrait.php
@@ -9,7 +9,6 @@ use OpisErrorPresenter\Implementation\Strategies\BestMatchError;
 use OpisErrorPresenter\Implementation\ValidationErrorPresenter;
 use RootedData\Exception\ValidationException;
 use Symfony\Component\HttpKernel\Exception\HttpException;
-use Symfony\Component\HttpKernel\Exception\HttpExceptionInterface;
 
 /**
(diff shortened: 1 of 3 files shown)


paul-m:
@todo Ensure this is not locked in to be MySQL specific.

paul-m:
Research says: It turns out that \PDOException is always handled at the Drupal DBAL level and turned into DatabaseExceptionWrapper. So any DKAN query should always see DatabaseExceptionWrapper and not \PDOException. Our high rate of connection errors is likely coming from other sources.

PR #4706, opened 2026-05-07
Migrate to opis/json-schema v2

dcgoodwin2112:
Migrates DKAN from `opis/json-schema` v1 to v2 (`^2.4`) and drops the unmaintained `m1x0n/opis-json-schema-error-presenter` add-on in favor of opis v2's built-in `ErrorFormatter`. Coordinates with [GetDKAN/RootedJsonData#24](https://github.com/GetDKAN/RootedJsonData/pull/24).

**Validation error response shape is preserved** — `JsonResponseTrait::getExceptionData()` returns the same `{keyword, pointer, message}` shape downstream consumers expect, sourced from opis v2's API. Only the message wording differs slightly (opis v2 templates vs m1x0n templates).

### Dependency changes

- **Add**: `opis/json-schema: ^2.4` (direct require)
- **Drop**: `m1x0n/opis-json-schema-error-presenter` (opis v1-only; no v2 equivalent needed)
- **Bump**: `getdkan/rooted-json-data` to `^0.2.4` (currently pinned via inline alias `dev-opis-v2-migration as 0.2.4` until upstream tags; revert that alias + `minimum-stability: dev` once tagged)

### Schema changes

| File | Change |
|---|---|
| `modules/dkan_harvest/schema/schema.json` | Set absolute `$id` (`https://schemas.getdkan.com/harvest-plan.json`) — v2 rejects relative URIs. |
| `schema/collections/dataset.json`, `schema/collections/publisher.json` | Bumped nested sub-schemas (publisher, contactPoint, publisher.data) from JSON Schema **draft-04 → draft-07**; renamed bare `id` → `$id` per draft-06+ conventions. |

### Test changes

| File | Change |
|---|---|
| `modules/dkan_harvest/tests/src/Unit/HarvesterTest.php` | Update `testPlanValidation` expectation from v1's string-form `args["missing"]: "x"` to v2's array-form `args["missing"]: ["x"]`. |
| `modules/dkan_harvest/tests/src/Kernel/Commands/HarvestCommandsTest.php` | Same array-form update across 5 data-provider entries. |
| **New: `modules/dkan_common/tests/src/Unit/JsonResponseTraitTest.php`** | Locks down the v1-compatible response shape (`{keyword, pointer, message}` with string values) so the bridge can't silently regress. Also covers the non-`ValidationException` pass-through and the no-error edge case. |
| `modules/dkan_metastore/tests/src/Unit/Plugin/Validation/Constraint/ProperJsonValidatorTest.php` | Tighten `testValidationFailure` to assert `addViolation` is called with a non-empty string referencing the failing field (was previously asserting only on call count). |

### Breaking changes for downstream consumers

| Change | Impact | Mitigation |
|---|---|---|
| Validation error message wording | Slight wording differences from m1x0n templates (e.g. `"minLength: must be at least 1 characters long"` → `"Minimum string length is 1, found 0"`) — same field, same constraint name, just different prose. | None required for typical consumers; update test fixtures asserting on exact strings |
| Custom schemas declaring draft-04 | Will throw `ParseException` at validation time | See JSON Schema migration notes below |
| Custom code importing `m1x0n\\OpisJsonSchemaErrorPresenter\\*` | Class-not-found at runtime | Port to `Opis\JsonSchema\Errors\ErrorFormatter` |
| Custom code calling v1-only opis API (`Schema::fromJsonString`, `getFirstError()`, `keywordArgs()`, `dataPointer()`) | Method/class-not-found | Port to v2 API (`Validator::validate()`, `error()`, `args()`, `data()->fullPath()`) |

### JSON Schema draft-04 → draft-07 notes

opis v2 ships parsers for **draft-06, draft-07, 2019-09, 2020-12**. **Draft-04 is not supported.** Three nested sub-schemas in DKAN's metadata schemas were on draft-04 and have been migrated.

The migration steps for any custom downstream schemas:

1. Bump `"$schema"` declaration to `"http://json-schema.org/draft-07/schema#"`
2. Rename `"id"` → `"$id"` (renamed at the draft-04 → draft-06 boundary)
3. Use absolute URIs for `$id` values (v2 rejects relative URIs at the schema root)
4. If using `exclusiveMinimum` / `exclusiveMaximum` as boolean modifiers, switch to numeric values (draft-06+ semantics)

DKAN has long had a bit of an issue with being based on DCAT-US 1.1 (despite supporting a lot of schema customization) but using a validator that does not support draft 4, which is what DCAT-US 1.1 is based on. While none of the DCAT-US schema properties conflict with later drafts, they do declare draft 4 and use the old `schema` (rather than `$schema`) keyword format. This has always necessitated slightly modifying the original schemas from their original form; now additional modifications will be needed at the sub-schema level.

Looking ahead, DCAT-US 1.1 is also out of date, as [DCAT-US 3.0](https://resources.data.gov/resources/dcat-us3/) is now the recommended schema, and it is [defined in draft 2022-12](https://github.com/GSA/dcat-us/blob/main/jsonschema/definitions/Catalog.json#L2C45-L2C52). DKAN should move to adopt 3.0 as its default schema as soon as possible.

## QA Steps

The response-shape preservation is locked down by `JsonResponseTraitTest`; the QA below verifies end-to-end behavior in the test site.

- [ ] **Composer install resolves cleanly**
  - `ddev composer install` succeeds without errors
  - `ddev composer why opis/json-schema` shows ^2.4 (and no v1 reference remains)
  - `ddev composer why m1x0n/opis-json-schema-error-presenter` shows it is no longer required
- [ ] **POST a valid dataset succeeds**
  - `curl -X POST -u admin:admin -H "Content-Type: application/json" -d @valid-dataset.json https://<site>/api/1/metastore/schemas/dataset/items` returns 201
- [ ] **POST an invalid dataset returns the v1-compatible error shape**
  - POST a dataset with `"title": ""` (or any spec-violating field)
  - Verify response `data` field has shape `{"keyword": "...", "pointer": "...", "message": "..."}` — keys, types, and pointer/keyword values match v1; only message wording may differ
  - Example expected: `{"keyword": "minLength", "pointer": "title", "message": "Minimum string length is 1, found 0"}`
- [ ] **Submit a malformed harvest plan**
  - `ddev drush dkan-harvest:register --identifier=test-bad --plan='{"identifier":"test-bad"}'` (missing required fields)
  - Verify exception message includes a path (e.g. `extract`) and JSON-encoded args (e.g. `{"missing":["type"]}`)
- [ ] **Frontend (`data-catalog-app`) continues to display field-level validation errors** correctly when a dataset edit fails validation
- [ ] **Watchdog clean**
  - `ddev drush watchdog:show --severity=Error --count=20` after performing the above QA steps shows no opis-v2-related fatals

  - `HarvesterTest::testPlanValidation` and 5 entries in `HarvestCommandsTest::providePlanOpts` updated for the v2 error-arg shape
  - New `JsonResponseTraitTest` locks down the v1-compatible response shape
  - `ProperJsonValidatorTest::testValidationFailure` tightened to assert message content, not just call count
  - Release notes for downstream consumers are pending (Phase 3b — not blocking this PR but should land before tagging a release)

## Related

- Upstream library PR: [GetDKAN/RootedJsonData#24](https://github.com/GetDKAN/RootedJsonData/pull/24)
- opis/json-schema v2 docs: https://docs.opis.io/json-schema/2.x/

Co-authored-by: Copilot <copilot@github.com>

### DIFF
--- a/modules/dkan_common/tests/src/Functional/Api1TestBase.php
+++ b/modules/dkan_common/tests/src/Functional/Api1TestBase.php
@@ -6,7 +6,6 @@ use Drupal\Tests\BrowserTestBase;
 use Drupal\Tests\user\Traits\UserCreationTrait;
 use GuzzleHttp\Client;
 use GuzzleHttp\RequestOptions;
-use Opis\JsonSchema\Schema;
 use Opis\JsonSchema\Validator;
 use Osteel\OpenApi\Testing\ValidatorBuilder;
 
@@ -99,10 +98,12 @@ abstract class Api1TestBase extends BrowserTestBase {
   }
 
   protected function assertJsonIsValid($schema, $json) {
-    $opiSchema = is_string($schema) ? Schema::fromJsonString($schema) : new Schema($schema);
-    $validator = new Validator();
+    // Opis v2: Schema::fromJsonString is gone (Schema is an interface).
+    // Validator::validate() accepts a decoded schema (object/bool) directly.
+    $decodedSchema = is_string($schema) ? json_decode($schema, FALSE) : $schema;
     $data = is_string($json) ? json_decode($json) : $json;
-    $result = $validator->schemaValidation($data, $opiSchema);
+    $validator = new Validator();
+    $result = $validator->validate($data, $decodedSchema);
     $this->assertTrue($result->isValid());
   }
 
--- a/modules/dkan_common/tests/src/Functional/Api1TestBase.php
+++ b/modules/dkan_common/tests/src/Functional/Api1TestBase.php
@@ -98,7 +98,6 @@ abstract class Api1TestBase extends BrowserTestBase {
   }
 
   protected function assertJsonIsValid($schema, $json) {
-    // Opis v2: Schema::fromJsonString is gone (Schema is an interface).
     // Validator::validate() accepts a decoded schema (object/bool) directly.
     $decodedSchema = is_string($schema) ? json_decode($schema, FALSE) : $schema;
     $data = is_string($json) ? json_decode($json) : $json;
--- a/modules/dkan_common/tests/src/Unit/JsonResponseTraitTest.php
+++ b/modules/dkan_common/tests/src/Unit/JsonResponseTraitTest.php
@@ -19,13 +19,8 @@ class JsonResponseTraitTest extends TestCase {
   /**
    * @covers ::getExceptionData
    *
-   * Locks down the historical v1/m1x0n response shape:
-   *   { "keyword": <string>, "pointer": <string>, "message": <string> }
-   *
-   * The v2 migration preserves this shape (the values are sourced from
-   * opis v2's API instead of m1x0n, and message wording may differ
-   * slightly, but the keys and types match) so downstream consumers
-   * parsing validation errors keep working unchanged.
+   * Makes sure that getExceptionData returns an array with the expected shape
+   * when a ValidationException is passed.
    */
   public function testGetExceptionDataReturnsV1CompatibleShape(): void {
     $schema = (object) [
--- a/modules/dkan_common/tests/src/Unit/JsonResponseTraitTest.php
+++ b/modules/dkan_common/tests/src/Unit/JsonResponseTraitTest.php
@@ -19,12 +19,15 @@ class JsonResponseTraitTest extends TestCase {
   /**
    * @covers ::getExceptionData
    *
-   * Verifies the v2 ErrorFormatter::format() shape — array keyed by JSON
-   * pointer, values are arrays of message strings. This is the BC-impacting
-   * shape change downstream consumers parsing validation error responses
-   * need to be aware of.
+   * Locks down the historical v1/m1x0n response shape:
+   *   { "keyword": <string>, "pointer": <string>, "message": <string> }
+   *
+   * The v2 migration preserves this shape (the values are sourced from
+   * opis v2's API instead of m1x0n, and message wording may differ
+   * slightly, but the keys and types match) so downstream consumers
+   * parsing validation errors keep working unchanged.
    */
-  public function testGetExceptionDataReturnsPointerKeyedShape(): void {
+  public function testGetExceptionDataReturnsV1CompatibleShape(): void {
     $schema = (object) [
       'type' => 'object',
       'properties' => (object) [
@@ -40,16 +43,16 @@ class JsonResponseTraitTest extends TestCase {
     $data = (new ClassUsingJsonResponseTrait())->callGetExceptionData($exception);
 
     $this->assertIsArray($data);
-    $this->assertNotEmpty($data);
-    foreach ($data as $pointer => $messages) {
-      $this->assertIsString($pointer, 'response keys are JSON pointers');
-      $this->assertIsArray($messages, 'response values are arrays of strings');
-      $this->assertNotEmpty($messages);
-      foreach ($messages as $message) {
-        $this->assertIsString($message);
-        $this->assertNotEmpty($message);
-      }
-    }
+    $this->assertSame(['keyword', 'pointer', 'message'], array_keys($data));
+    $this->assertIsString($data['keyword']);
+    $this->assertNotEmpty($data['keyword']);
+    $this->assertIsString($data['pointer']);
+    $this->assertIsString($data['message']);
+    $this->assertNotEmpty($data['message']);
+    // Pointer should reference the failing field.
+    $this->assertStringContainsString('title', $data['pointer']);
+    // Keyword should name the failing constraint.
+    $this->assertSame('minLength', $data['keyword']);
   }
 
   /**
--- a/modules/dkan_common/tests/src/Unit/JsonResponseTraitTest.php
+++ b/modules/dkan_common/tests/src/Unit/JsonResponseTraitTest.php
@@ -2,7 +2,7 @@
 
 namespace Drupal\Tests\dkan_common\Unit;
 
-use Drupal\dkan_common\JsonResponseTrait;
+use Drupal\Tests\dkan_common\Unit\Mocks\ClassUsingJsonResponseTrait;
 use Opis\JsonSchema\Validator;
 use PHPUnit\Framework\TestCase;
 use RootedData\Exception\ValidationException;
@@ -81,20 +81,6 @@ class JsonResponseTraitTest extends TestCase {
     $this->assertFalse(
       (new ClassUsingJsonResponseTrait())->callGetExceptionData($exception)
     );
-  }
-
-}
-
-/**
- * Stub exposing the protected trait method for tests.
- *
- * Mirrors the pattern in CacheableResponseTraitTest.
- */
-class ClassUsingJsonResponseTrait {
-  use JsonResponseTrait;
-
-  public function callGetExceptionData(\Exception $e) {
-    return $this->getExceptionData($e);
   }
 
 }
--- a/modules/dkan_common/tests/src/Unit/JsonResponseTraitTest.php
+++ b/modules/dkan_common/tests/src/Unit/JsonResponseTraitTest.php
@@ -1,0 +1,97 @@
+<?php
+
+namespace Drupal\Tests\dkan_common\Unit;
+
+use Drupal\dkan_common\JsonResponseTrait;
+use Opis\JsonSchema\Validator;
+use PHPUnit\Framework\TestCase;
+use RootedData\Exception\ValidationException;
+
+/**
+ * @coversDefaultClass \Drupal\dkan_common\JsonResponseTrait
+ *
+ * @group dkan
+ * @group dkan_common
+ * @group unit
+ */
+class JsonResponseTraitTest extends TestCase {
+
+  /**
+   * @covers ::getExceptionData
+   *
+   * Verifies the v2 ErrorFormatter::format() shape — array keyed by JSON
+   * pointer, values are arrays of message strings. This is the BC-impacting
+   * shape change downstream consumers parsing validation error responses
+   * need to be aware of.
+   */
+  public function testGetExceptionDataReturnsPointerKeyedShape(): void {
+    $schema = (object) [
+      'type' => 'object',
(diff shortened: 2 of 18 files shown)


modules/dkan_common/tests/src/Functional/Api1TestBase.php
```diff
@@ -99,10 +98,12 @@ public function setUp(): void {
   }
 
   protected function assertJsonIsValid($schema, $json) {
-    $opiSchema = is_string($schema) ? Schema::fromJsonString($schema) : new Schema($schema);
-    $validator = new Validator();
+    // Opis v2: Schema::fromJsonString is gone (Schema is an interface).
```
dafeder: I think we can leave this comment out; it won't be relevant as soon as we make the change

schema/collections/dataset.json
```diff
@@ -141,8 +141,8 @@
       "minLength": 1
     },
     "publisher": {
-      "$schema": "http://json-schema.org/draft-04/schema#",
-      "id": "https://project-open-data.cio.gov/v1.1/schema/organization.json#",
+      "$schema": "http://json-schema.org/draft-07/schema#",
+      "$id": "https://project-open-data.cio.gov/v1.1/schema/organization.json#",
       "title": "Organization",
```
dafeder, line 145: Do these break the validation as is? We should probably just removed them, but customized schemas likely still have them. But we can't really lie and say a particular schema is draft-07 when it's really 04.

dafeder (changes requested):

modules/dkan_harvest/schema/schema.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "harvest-plan",
```
dcgoodwin2112, line 3: This was failing in OPIs V2, because it expects an absolute URI

dafeder:
@dcgoodwin2112 I think if you rebase the tests should pass

modules/dkan_common/tests/src/Functional/Api1TestBase.php
dcgoodwin2112: Removed - marking as resolved

schema/collections/dataset.json
dcgoodwin2112: Removed Draft 7 $schema

modules/dkan_harvest/schema/schema.json
dafeder: Not a problem, since this is internal to DKAN

modules/dkan_common/tests/src/Unit/JsonResponseTraitTest.php
```diff
@@ -0,0 +1,100 @@
...
+ * Mirrors the pattern in CacheableResponseTraitTest.
+ */
+class ClassUsingJsonResponseTrait {
+  use JsonResponseTrait;
+
+  public function callGetExceptionData(\Exception $e) {
+    return $this->getExceptionData($e);
+  }
+
+}
```
dafeder: We don't generally declare a second class in a file. Either make a second file for the class and give it the class name (you probably need to require it explicitly in at the top of the file as it will not be autoloaded) or build test against an existing class that uses the trait.

modules/dkan_harvest/src/ETL/Factory.php
```diff
@@ -144,19 +143,22 @@ public static function validateHarvestPlan(object|string|NULL $harvest_plan): bo
...
 
+    // Opis v2: Validator::validate() accepts a JSON-string schema; pre-decode
+    // to bypass v2's URI-first parse path.
     $data = $harvest_plan;
-    $schema = Schema::fromJsonString($json_schema);
     $validator = new Validator();
-
-    /** @var ValidationResult $result */
-    $result = $validator->schemaValidation($data, $schema);
+    $result = $validator->validate($data, json_decode($json_schema, FALSE));
```
dafeder: Are we properly set up to catch bad JSON here? I believe json_decode  just returns NULL if the JSON string is bad, which could lead to silent failures we didn't use to have?

modules/dkan_metastore/src/Plugin/Validation/Constraint/ProperJsonValidator.php
```diff
@@ -107,7 +94,10 @@ private function doValidate(string $schema_id, $item): array {
       $this->validMetadataFactory->get($item->value, $schema_id);
     }
     catch (ValidationException $e) {
-      $errors = $this->getValidationErrorsMessages($e->getResult()->getErrors());
+      $rootError = $e->getResult()->error();
+      if ($rootError) {
```
dafeder: If no error $rootError will be `null`. Better practice would be `if ($rootError ?? FALSE)` or else use `ValidationResult::hasError()`

modules/dkan_metastore/src/Plugin/Validation/Constraint/ProperJsonValidator.php
```diff
@@ -116,22 +106,22 @@ private function doValidate(string $schema_id, $item): array {
...
    *
-   * @param array $errors
-   *   Validation errors array.
-   *
-   * @return array
-   *   Presented errors array.
+   * Uses ErrorFormatter::formatKeyed() to walk leaves only (avoiding the
+   * redundancy of formatFlat() which also includes container errors), then
+   * collapses the pointer-keyed groups into the array<string> shape that
+   * addViolation() expects.
```
dafeder: Can you restore the @param @return annotations here?

dafeder (changes requested):
I'm additionally thinking, this could potentially render peoples' dataset schemas invalid immediately. It would be nice to use either a hook_update or hook_requirements to check the existing schemas and report on whether the site is now in an invalid state, and provide useful information on what's wrong.

modules/dkan_metastore/src/Plugin/Validation/Constraint/ProperJsonValidator.php
dcgoodwin2112: Switched to ValidationResult::hasError() — also applied the same fix to JsonResponseTrait::getExceptionData() since it had the identical truthy-check pattern. bb0b9ca

modules/dkan_metastore/src/Plugin/Validation/Constraint/ProperJsonValidator.php
dcgoodwin2112: Restored. Updated the parameter type to \Opis\JsonSchema\Errors\ValidationError $error (singular) since the v2 signature changed from v1's array $errors. bb0b9ca

modules/dkan_harvest/src/ETL/Factory.php
dcgoodwin2112: Good catch — v1's Schema::fromJsonString() threw on bad JSON, v2's silent NULL path masked the failure. Added explicit checks for both file_get_contents returning FALSE and json_decode returning NULL. bb0b9ca

Heads-up: there are ~5 other places doing file_get_contents + json_decode without null guards (DkanApiDocsBase, AbstractQueryController, DatastoreQuery, plus two tests). Out of scope for this PR but worth a follow-up.

Drafted with Copilot <copilot@github.com>

modules/dkan_common/tests/src/Unit/JsonResponseTraitTest.php
dcgoodwin2112: Moved to modules/dkan_common/tests/src/Unit/Mocks/ClassUsingJsonResponseTrait.php — picking up the Mocks/ namespace DKAN already uses for other test doubles (e.g. DatabaseConnectionFactoryMock). Drupal's test bootstrap PSR-4-autoloads the Drupal\Tests\* namespace so no explicit require_once was needed. bb0b9ca

The same inline-stub pattern still exists in CacheableResponseTraitTest.php — happy to clean that up in a follow-up PR if you'd like.

Drafted with: Copilot <copilot@github.com>

dcgoodwin2112:
> I'm additionally thinking, this could potentially render peoples' dataset schemas invalid immediately. It would be nice to use either a hook_update or hook_requirements to check the existing schemas and report on whether the site is now in an invalid state, and provide useful information on what's wrong.

**Added in bb0b9ca.**

- _dkan_metastore_validate_schema_json() — pure helper that tries to parse a schema with opis v2. Forces both the loader's eager parse and a trivial validate() against null data, because loadObjectSchema() returns a LazySchema that defers sub-schema parsing — the trivial validate makes deep structural errors surface too.
- _dkan_metastore_check_schemas() — runs the helper across every schema ID SchemaRetriever::getAllIds() knows about.
- hook_requirements() — surfaces broken schemas on /admin/reports/status as REQUIREMENT_WARNING (site keeps running; affected schema will fail at runtime when accessed).
- dkan_metastore_update_11302() — one-shot report at drush updb time. Doesn't block the upgrade so sites can fix offending schemas after seeing the output; the requirements hook keeps the warning visible until resolved.

**Tests:**

- SchemaCheckHelperTest (unit, 9 tests) covers draft-04 declarations, malformed JSON, non-object/non-bool inputs, root-level structural errors, and LazySchema-deferred sub-schema errors.
- SchemaCheckTest (kernel) confirms DKAN's bundled schemas parse cleanly under opis v2 (regression guard).

Drafted with: Copilot <copilot@github.com>

dafeder:
I think this is about ready. It will necessitate a 4.1 release, detailed release notes, and I'd like to update the upgrade steps in the docs to mention this since anyone who hasn't already upgraded from 2.x is likely to go straight to 4.1.

swirtSJW (commented):
@dafeder Should this include documentation updates that describe what needs to be changed in existing schemas or is that a separate thing?

This seems like it could benefit from an ADR and maybe a change notice?

swirtSJW:
OOps sorry, I just saw your previous comments about documentation.

dafeder:
I don't think this rises to the level of an ADR (which we are yet to actually adopt) but it does seem like a good job for a Drupal change record.

dafeder (approved):

PR #4711, opened 2026-05-08
#4710: Metastore Search - check for empty facet from content

### LINKED ISSUE #4710: Metastore Search - check for empty facet from content
### User Story

In a non-standard DKAN implementation the metastore search with facets might end in a 500 error, complaining in the log

`InvalidArgumentException: Facets much name and total properties. `

This might happen if the publisher__name facet does not exist in the current implementation. An illegal option is added to the "facets from content" in this case, which raises the given exception.

I would propose to check for validity of the facet in FacetsFromContentTrait and skipping empty/illegal facet values.

### Problem and Motivation

_No response_

### Target Audience

- [x] End Users
- [ ] Developers
- [ ] Administrators
- [ ] Other

### Possible Implementation

_No response_

### Acceptance Criteria

_No response_

stefan-korn:
Fixes #4710 

## QA Steps

### DIFF
--- a/modules/metastore/modules/metastore_search/src/FacetsFromContentTrait.php
+++ b/modules/metastore/modules/metastore_search/src/FacetsFromContentTrait.php
@@ -130,11 +130,13 @@ trait FacetsFromContentTrait {
 
         $facet = isset($field) ? $facet->{"$.data." . $field} : $facet->{"$.data"};
 
-        $facets["{$type}:{$facet}"] = (object) [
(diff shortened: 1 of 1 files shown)


dafeder:
Hey @stefan-korn as soon as I get #4726 merged you can update this branch against 2.x and I'll merge. We're really trying to avoid new merges and releases to 2.x now but we can do this one if you need it before you can realistically upgrade.

PR #4720, opened 2026-05-14
Return a 503 for specific PDO exceptions, alternate mix

paul-m:
Fixes [issue#]

## Describe your changes

This is another way to do https://github.com/GetDKAN/dkan/pull/4704 much more generalized.

- Marks DKAN API routes as DKAN API routes in the yml files.
- Adds an exception event subscriber, so we can handle specific exceptions.
- We're handling the case of a `\PDOException` with a code of 1203 (too many connections).
- Always return a `JsonResponse` with a 503 response code for this exception, and a `Retry-After` header of our choosing.

To see it in action, make a local and modify an API controller method to throw the exception. I suggest `Drupal\dkan_datastore\Controller\AbstractQueryController::queryResource`. Throw an exception like this:
```
    throw new \PDOException(
      'AI bots taking over.', 1203
    );
```
Now request: https://dkan.ddev.site/api/1/datastore/query/8701d18a-5792-5451-8700-a621bccc5f6e

See a message like this:
```
{
"message": "Try again later after 120 seconds.",
"code": 503
}
```

Watchdog should show you a message like: "PDOException (code: 1203) AI bots taking over."

## QA Steps

### DIFF
--- a/modules/dkan_common/dkan_common.routing.yml
+++ b/modules/dkan_common/dkan_common.routing.yml
@@ -10,7 +10,8 @@ dkan.common.api:
   path: '/api'
   methods: [GET]
   defaults:
-    { _controller: '\Drupal\dkan_common\Controller\OpenApiController::getVersions'}
+    _controller: '\Drupal\dkan_common\Controller\OpenApiController::getVersions'
+    is_dkan_api_route: true
   requirements:
     _permission: 'access content'
   options:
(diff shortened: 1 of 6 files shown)


paul-m:
Closed in favor of handling this in the controller: https://github.com/GetDKAN/dkan/pull/4704

PR #4721, opened 2026-05-15
Empty datastore error

dafeder:
Throws a specific exception when a table has been created but not populated.

After significant back-and-forth and debate I settled on 404 as the HTTP error code, but this is still negotiable.

## QA Steps

1. Build and load sample content
2. Use drush dkan:dataset-info to get the table name of an imported resource
3. Run `drush sqlc`, and then run `TRUNCATE [table-name]`
4. Do a datastore query against that resource (using any endpoint)
5. Confirm 404 error with with appropriate message.

### DIFF
--- a/modules/dkan_common/src/Storage/AbstractDatabaseTable.php
+++ b/modules/dkan_common/src/Storage/AbstractDatabaseTable.php
@@ -421,6 +421,9 @@ abstract class AbstractDatabaseTable implements DatabaseTableInterface {
    * {@inheritdoc}
    */
   public function tableIsEmpty(): bool {
+    if (!$this->tableExist($this->getTableName())) {
+      return TRUE;
+    }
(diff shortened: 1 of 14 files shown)


paul-m:
After truncating the sample dataset (Florida bike lanes), a query does this:
`https://dkan.ddev.site/api/1/datastore/query/1ff9739f-1a71-5281-8f8f-219d1be58a4d?limit=20&offset=0`
```
{
    "message": "The resource 3a187a87dc6cd47c48b6b4c4785224b7__1779135275__local_file does not contain any data.",
    "status": 404,
    "timestamp": "2026-05-18T13:17:24-07:00"
}
```
Reimporting the resource using `dkan:datastore:drop` and `dkan:datastore:import` then gives back the 200 result.

This looks like the desired behavior.

paul-m (approved):

PR #4727, opened 2026-05-27
Use pcov for test coverage

dafeder:
Hopefully this will significantly speed up our "target" build, which is slowed down dramatically by xdebug.

### DIFF
--- a/.ddev/php/pcov.ini
+++ b/.ddev/php/pcov.ini
@@ -1,0 +1,7 @@
+[pcov]
+pcov.enabled=1
+pcov.directory=/var/www/html
+
+[opcache]
+opcache.jit=off
+opcache.jit_buffer_size=0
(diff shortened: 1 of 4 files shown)


.ddev/php/pcov.ini
```diff
@@ -0,0 +1,7 @@
+[pcov]
+pcov.enabled=1
+pcov.directory=/var/www/html
+
+[opcache]
+opcache.jit=off
+opcache.jit_buffer_size=0
```
dafeder, line 7: I don't totally understand it but apparently the JIT extension conflicts with pcov. Disabling it may impact performance, so we'll have to see if this negates the gains from switching to pcov.

dafeder: Looks like this is not a problem; performance seems unaffected

dafeder:
Appears this cuts the phpunit stage time almost in half!

janette (approved):
50% faster sounds good to me

PR #4729, opened 2026-06-03
Validate buried sub-schemas in metastore schema check

dcgoodwin2112:
## Describe your changes

`_dkan_metastore_validate_schema_json()` only reliably validated the schema root: opis defers sub-schema parsing (`LazySchema`) and `validate(NULL, …)` never reaches `properties`/`$defs`/`items`/etc. or short-circuited `allOf`/`anyOf`/`then`/`else` branches, so buried draft-04 or structural errors passed silently.

- Replaced the root-only check with a recursive walk over genuine schema positions (applicator keywords + opis `$slots`), resolving each node against a single cached load (preserves draft/base/`$ref` context).
- Recover precise messages via a `SchemaParser` subclass exposing the throwing parse path (`parseSchema()` swallows errors into `ExceptionSchema`); reports the JSON pointer of the offending node.
- Non-schema positions (the `properties` map, a property named `format`, `dependencies` property-name arrays) are skipped — no false positives.

## QA Steps

- [ ] `drush cr`, then visit `/admin/reports/status` — no DKAN metastore schema warning on a clean site.
- [ ] Override a metastore schema with a buried draft-04 sub-schema (e.g. under `properties.*`); confirm the status report (and `drush updatedb` / `dkan_metastore_update_11302`) reports it with its pointer.
- [ ] Restore the schema; confirm the warning clears.
- [ ] `phpunit modules/dkan_metastore/tests/src/{Unit,Kernel}/Install/` passes.

## Checklist before requesting review

- [x] I have updated or added tests to cover my code — added buried/short-circuit/false-positive cases to `SchemaCheckHelperTest`; existing `SchemaCheckTest` guards real schemas.
- [ ] I have updated or added documentation — n/a; no behavior change beyond broader detection (upgrade docs already cover the draft-04 → draft-07 migration).

Co-authored by GitHub Copilot

PR #4730, opened 2026-06-07
Fix Opis v2 warnings, add schema validator service

dafeder:
Re-working of #4729 by @dcgoodwin2112: Fixes warnings for broken schemas from Opis v2 upgrade, and significantly refactors the schema validator logic into its own service.

Also took the opportunity to clean up SchemaRetriever docs.

## QA Steps

- Replace `schema/collections/dataset.json` with the contents of this file: https://github.com/GetDKAN/dkan/blob/cea733db3c589636a1d6702710b3dc3f44d2d4d9/schema/collections/dataset.json
- Clear cache.
- Check the Drupal status page. You should see a message like this: "1 schema(s) cannot be parsed
- The following metastore schemas could not be parsed under opis/json-schema v2: dataset. They will fail at runtime when used. See [DKAN change record](https://www.drupal.org/node/3593097) for the draft-04 -> draft-07 migration steps."
- Revert the file and this message should not appear.

### DIFF
--- a/modules/dkan_metastore/tests/src/Kernel/SchemaValidatorTest.php
+++ b/modules/dkan_metastore/tests/src/Kernel/SchemaValidatorTest.php
@@ -29,8 +29,8 @@ class SchemaCheckTest extends KernelTestBase {
   public function testBundledSchemasParseCleanly(): void {
     $module_path = \Drupal::service('extension.list.module')->getPath('dkan_metastore');
     require_once DRUPAL_ROOT . '/' . $module_path . '/dkan_metastore.install';
-    $problems = _dkan_metastore_check_schemas();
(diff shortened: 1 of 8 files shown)


modules/dkan_metastore/tests/src/Kernel/SchemaValidatorTest.php
```diff
@@ -11,7 +11,7 @@
  * @group dkan_metastore
  * @group kernel
  */
-class SchemaCheckTest extends KernelTestBase {
+class SchemaValidatorTest extends KernelTestBase {
```
paul-m, line 14: This test is in a new namespace.

paul-m (changes requested):

paul-m (approved):

PR #4731, opened 2026-06-11
Remove drush.services.yml files, use autowire and attributes

### LINKED ISSUE #4572: Use create() method instead of drush.services.inc
Many of our Drush commands use service yml files for dependency injection. [This is deprecated](https://www.drush.org/11.x/dependency-injection/#create-method); remove these and add `:create()` methods to all command classes.

dafeder:
Fixes #4572 

Updates our Drush commands to properly use autowiring, attributes and stop using drush.services.yml files.

Also some cleanup of class names, docs, etc.

Removed JobstoreCommands, as this turned out to be broken anyway. No one should still be using this; it was to mitigate breaking changes many Drupal versions ago.

Not sure why a few of these show up as new files rather than renames, making it a little harder to understand the diff. In general there should be no logic changes within any of the functions.

## QA Steps 

1. On a clean 4.x site, run `ddev drush list | rg 'dkan:' | wc -l`.
2. Rebuild on this branch. Clear all caches.
3. Run `ddev drush list | rg 'dkan:' | wc -l` again. Confirm there is only one less (because `dkan:jobstore-fixer` is gone)


modules/dkan_datastore/tests/src/Unit/DatastoreLookupTest.php
```diff
@@ -61,22 +53,6 @@ protected function setUp(): void {
...
-    //   $this->createMock(\Drupal\dkan_datastore\DatastoreService::class),
-    //   $this->createMock(\Drupal\dkan_datastore\Service\ResourceLocalizer::class),
-    //   $this->createMock(\Drupal\dkan_metastore\ResourceMapper::class),
-    //   $this->createMock(\Drupal\dkan_datastore\Service\Info\ImportInfoList::class),
-    //   $this->createMock(\Drupal\dkan_datastore\PostImportResultFactory::class),
-    //   $this->datastoreLookupInterface
-    // );
-
-    // // Set the output property.
-    // $this->drush->setOutput($this->output);
```
dafeder, line 79: Not sure why this was here but it's never been not-commented out, so removing

janette (approved):

PR #4733, opened 2026-06-12
#4732 - Limit the amount of items for metastore schema API request

### LINKED ISSUE #4732: API Metastore Schema items - provide limit parameter
### User Story

The API call to `/api/1/metastore/schemas/{schema_id}/items` (Get all items for a specific schema) may provide a lot of items and especially in the "Try it out" mode in the web interface lead to very slow response or even freezing in the browser.

To help with this, I would propose to add a parameter to the query to limit the number of items. And for the "Try it out" part I would recommend to set it to some number, like 10, 25, 50.

By direct call (like Curl) it seems generally safe to call with no limit, even on a large number of items. But still it seems reasonable to limit the number of items in requests via the Web UI.

### Problem and Motivation

_No response_

### Target Audience

- [ ] End Users
- [ ] Developers
- [ ] Administrators
- [ ] Other

### Possible Implementation

_No response_

### Acceptance Criteria

_No response_

stefan-korn:
Fixes #4732 

## QA Steps

### DIFF
--- a/modules/dkan_metastore/docs/openapi_spec.json
+++ b/modules/dkan_metastore/docs/openapi_spec.json
@@ -22,6 +22,17 @@
                 "required": true,
                 "allowEmptyValue": false,
                 "examples": []
+            },
+            "limit": {
+                "name": "limit",
+                "in": "query",
+                "description": "Limit the output of items, mainly for testing purposes.",
+                "schema": {
+                    "type": "integer",
+                    "default": 10
+                },
+                "style": "form",
+                "allowEmptyValue": true
             }
         },
         "responses": {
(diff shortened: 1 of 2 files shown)


dafeder (changes requested):
I talked about this on the issue comments but:

* I think if we introduce limit we should also have offset, so that people can paginate through results. However, I'm open to breaking that off for a second issue if you feel it's scope creep.
* I think the name of the method is a little confusing. I'd suggest `limitItems()` - it's more concise, and makes it clear that we're applying a limit not just determining whether a limit should be there (which I'd expect to return a `bool`)
* Can't merge without test coverage for limits

PR #4742, opened 2026-06-18
Distribution predelete check resource

### LINKED ISSUE #4698: Consider: slowing down the orphan processor
### User Story

There are times when there may be a data hiccup (a datadictionary gets modified...) and within a few cron runs the orphan processor has already removed the resources from the files.  By the time the problem is noticed, there is no easy way to go back.  Reverting a revision is not really an option.
It would be nice if the orphan removal did not take place so rapidly and put us in a position of data loss.  

This could be achieved in a number of ways:

- **Restrict the queue**: The orphan removal queue only runs the first day of the month or every 10 days ...
- **Restirct queue items**: The processor does not process it and puts an item back in the queue if it is less than X days since it was added to the queue.
- **Orphanage**:  Move orphans to a different location in the files directory where they can sit until the orphanage gets burned down every month or year. (this method would not make it as clean to revert a revision)
- ?

### Problem and Motivation

Loss of data due to hiccups of one kind or another is bad.  By the time we are alerted to a problem, it is too late and we only get lucky if there are ways for us to recover the file(s) that were removed.

### Target Audience

- [x] End Users
- [ ] Developers
- [x] Administrators
- [ ] Other

### Possible Implementation

See feature description.

### Acceptance Criteria

Files don't get removed so quickly

janette:
Fixes #4698

## Describe your changes
If you've configured your site to delete orphaned nodes, the resource of any orphaned distribution node should be checked to confirm it is not still in use before being removed from the resource mapper table.

## QA Steps

- [x] Visit /admin/dkan/properties
- [x] Scroll to the bottom and check the "Delete referenced content after a dataset is deleted" box and save
- [x] Edit the SECONDS_PER_DAY in [OrphanNodeProcessor](https://github.com/GetDKAN/dkan/blob/4.x/modules/dkan_metastore/src/Service/OrphanNodeProcessor.php#L16) to make testing easier, 30 works well.
- [x] Create a dataset with a CSV resource, run cron twice
- [x] Edit the dataset, add a title to the distribution, leaving the downloadURL as is and save
- [x] Run cron, the previous distribution will be orphaned. Wait for 30 seconds and run cron again to initiate the deletion.
- [x] Confirm you see a logged message "1 orphaned node has been deleted"
- [x] Go back to /admin/dkan/datasets and click edit for your dataset and confirm the downloadURL is intact.

### DIFF
--- a/modules/dkan_metastore/dkan_metastore.module
+++ b/modules/dkan_metastore/dkan_metastore.module
@@ -157,21 +157,25 @@ function dkan_metastore_cron() {
         $message = 'No orphaned nodes deleted';
       }
       else {
-        $max_displayed = 20;
-        $count_msg = '';
-        if (count($nids) > $max_displayed) {
-          $count_msg = ' (and ' . count($nids) - $max_displayed . ' more)';
-          $nids = array_slice($nids, 0, 20);
+        $max_displayed = 10;
+        $count_msg = '1 orphaned node was deleted.';
+        $extra = '';
+        if (count($nids) > 1) {
+          $count_msg = count($nids) . ' orphaned nodes were deleted.';
         }
-        $message = 'Orphaned nodes ' . implode(',', $nids) . $count_msg . ' were deleted';
+        if (count($nids) > $max_displayed) {
+          $extra = ' (and ' . count($nids) - $max_displayed . ' more)';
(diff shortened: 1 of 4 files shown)


dafeder (commented):
lgtm

dafeder (approved):

PR #4748, opened 2026-07-02
Non-referenced distribution lifecycle changes

### LINKED ISSUE #4735: Dataset Discovery in LifeCycle Flow
- [ ] Add dataset discovery fixture variants and lifecycle coverage for referenced/non-referenced distributions in `modules/dkan_metastore/tests/src/Functional/OnPreReferenceTest.php`
- [ ] Add mixed-validity distribution fixture data (valid, missing, invalid, repeated downloadURL) in `modules/dkan_metastore/tests/src/Functional/Api1/DistributionHandlingTest.php`
- [ ] Add datastore subscriber fixture and contract coverage for multi-URL processing in `modules/dkan_datastore/tests/src/Unit/EventSubscriber/DatastoreSubscriberTest.php`
- [ ] Refactor pre-reference datastore trigger path to use metastore discovery service in `modules/dkan_datastore/src/EventSubscriber/DatastoreSubscriber.php`
- [ ] Wire `LifeCycle::EVENT_PRE_REFERENCE` dataset metadata payload to existing datastore initiation flow in `modules/dkan_metastore/src/LifeCycle/LifeCycle.php`

Created with assistance from [speckit](https://github.com/github/spec-kit).

dafeder:
Fixes #4735

This introduces fairly sweeping changes to the metastore (and less so to the datastore) to alow the metastore "lifecycle" to function correctly, particularly with regards to triggering resource creation (and therefore datastore imports), whether or not distributions are referenced.

Major changes to accomplish this:

1. `ReferenceLookup::getReferencers()` matches references recursively so it works when distributions are stored inline within datasets (and still works when they are referenced)
2. `ResourceMapper` calls have been moved from distribution-specific code in `LifeCycle` to `Referencer` and `Dereferencer`. `LifeCycle` is now significantly simpler.
3. Finding dependencies for cache tag behavior is a bit more complicated now, as it has to accommodate possibility of distribution nodes existing, or not.

Related changes in the datastore are mainly at the controller level. The resource ID has replaced distribution ID in some places as the canonical identifier for datastore resources.

Secondarily, it changes the ResourcePurger logic to fix an intermittent but that could surface when ResourceMapper records are removed before the purger checks to see if there is a local resource to delete. This may not have caused many production issues but was causing a lot of test failures. Specifically, `ResourceLocalizer::remove()` no longer depends on the source perspective still existing. It looks the local_url/local_file perspectives up directly in the resource mapper and derives the localized directory from identifier_version, so files are cleaned up even after orphan processing has removed the source perspective.

Signatures changed in the following Metastore service classes:

1. LifeCycle: removed ResourceMapper and DateFormatter
2. Dereferencer: Aded MetastoreUrlGenerator, ResourceMapper
3. Referencer: Added ResourceMapper

...also, DateFormatter is now a soft dependency for DatasetInfo, injected via setter.

And in the following Datastore controllers:

1. AbstractQueryController: added ReferenceLookup
2. QueryDownloadController: added ReferenceLookup

## Known remaining issues

- I haven't introduced the "both modes" testing trait to the data dictionaries test, and generally have punted on data dictionary logic. That all probably need a significant amount of work still. At least it's all still working in referenced mode.
- I haven't really looked at the import dashboard at all as part of this work, or updated any of its tests. This surely requires some more work as well, but probably much less than dictionaries. 

...

## Notes

* There are a few places I had to add some teardown and cleanup code to deal with the fact that I added some tests that needed to explicitly set the drupal_static flag we use (`metastore_resource_mapper_new_revision`). This is really a problematic piece of logic that should be factored out, I have some experiments in that direction in separate PRs. But it turns out that those static vars can leak between PHPUnit tests, since the test runner will run several tests on the same process. So it's important to clean up correctly even in unit tests.
* Provides better test coverage for workflows that depend on the drupal_static flag. This caused a bunch of failing tests because PHPUnit uses the same PHP process for multiple tests, so flags can persist between tests. This was yet another reminder that we should really factor out those static flag calls. I did some experiments on this with copilot and may come back to it but kept it out of scope for this PR.
* Watching tests revealed that the original logic to add `%modified` to metadata items, a pseudo-field to get access to Drupals last-modified node information, was not actually working, as the field was later being stripped out by removeReferences() in almost all workflows. So I stripped out that whole piece of functionality, as I could not see any places where it had been actually available to API users recently. The only place it appeared to be used was in building the dataset-info report, and because that code has direct access to the $node object, inserting it into the JSON metadata was not necessary. Glad to have eliminated this as it was not a good separation of concerns.
* I noticed the issues with %modified partly thanks to the cypress metastore tests. They still give better coverage than our PHPUnit functional tests for the API. Among other things they test all the methods on several schema types not just datastore.
* Work was a reminder of how annoying the UrlHostTokenResolver is - it breaks dependency injection all over the place. I'm pretty sure this is all really unnecessary and can be handled in a different, more Drupal-native way. Would be another good piece of tech debt to squash.
* Changed signature for `MetastoreService::removeReferences`, which was causing confusion. Added const for `"%Ref:` prefix and updated several classes to use it.
* The /imports endpoint supports both distribution IDs and resource IDs. However, the docs have never mentioned that distribution IDs are accepted. I removed the tests for distribution IDs because obviously they won't work if references are off. I'm on the fence about whether to remove the logic from the controller yet.
* The dataset /docs endpoint is broken for non-referenced distributions, I think.
* A future PR (in this same epic) should add a resource-id-based datastore query endpoint. Distribution IDs should only be supported as a legacy thing.


modules/dkan_metastore/src/LifeCycle/LifeCycle.php
dafeder, line 482: Note - I'm very nervous about removing this part. Need to do testing to make sure we're not losing anything important here in edge cases

modules/dkan_common/tests/src/Kernel/StreamWrapper/DkanStreamWrapperTest.php
```diff
@@ -19,24 +19,14 @@ class DkanStreamWrapperTest extends KernelTestBase {
...
-      $api = json_decode(file_get_contents('dkan://metastore'));
-    }
-    catch (\Throwable $e) {
-      $this->fail('Failed to read from DKAN stream wrapper. This is usually ' .
-      'a side effect of another issue, not a problem with the stream wrapper ' .
-      'itself. Please check for other test failures, especially in the ' .
-      'functional tests, or test the /api/1 endpoint directly. ' .
-      "Full error message:\n\n" . $e->getMessage());
-    }
-    $this->assertEquals('API Documentation', $api->info->title);
```
dafeder, line 32: So, I finally realized why this fails so regularly and mysteriously. This was really not a propper assertion for a kernel test, as it relies on the web server. Whatever it was getting for dkan://metastore was coming from the actual web containers in ddev, not the kernel test environment. So the test was dependent on the dev environment state...

.ddev/commands/host/dkan-module-test-cypress
```diff
@@ -33,8 +33,12 @@ fi
...
   {"name":"testadmin","mail":"testadmin@test.com","role":"administrator"}
 ]'
-readarray -t TU_TEST_USERS < <(jq -c '.[]' <<<"$TEST_USERS_JSON")
-for TU_TEST_USER in "${TU_TEST_USERS[@]}"; do
+TEST_USER_NAMES=""
+jq -r '.[] | .name' <<<"$TEST_USERS_JSON" | while read -r name; do
+  TEST_USER_NAMES="$TEST_USER_NAMES $name"
+done
+
+jq -c '.[]' <<<"$TEST_USERS_JSON" | while read -r TU_TEST_USER; do
```
dafeder, line 41: This wasn't working in zsh form me till i did this. Drawback is it expects jq but that's probably fair?

modules/dkan_common/src/DatasetInfo.php
```diff
@@ -133,6 +150,38 @@ public function getDistributionUuid(string $dataset_uuid, string $index = '0'):
...
+   *
+   * @param string $dataset_uuid
+   *   The uuid of a dataset.
+   * @param string $index
+   *   The index of the resource in the dataset array. Defaults to first.
+   *
+   * @return string|null
+   *   The resource identifier as "id__version", or NULL if none.
+   */
+  public function getResourceIdentifier(string $dataset_uuid, string $index = '0'): ?string {
```
dafeder, line 170: This is primarily used in the datastore query controllers, which now need to know the resource ID rather than using a distribution ID as a shortcut

modules/dkan_common/src/DatasetInfo.php
```diff
@@ -153,7 +204,7 @@ protected function getRevisionInfo(Node $node) : array {
       'moderation_state' => $node->get('moderation_state')->getString(),
       'title' => $metadata->title ?? 'Not found',
       'modified_date_metadata' => $metadata->modified ?? 'Not found',
-      'modified_date_dkan' => $metadata->{'%modified'} ?? 'Not found',
+      'modified_date_dkan' => $modified,
```
dafeder, line 207: We're no longer injecting `%modified` into the metadata, as the only place it was actually being used was here.

modules/dkan_common/tests/src/Kernel/StreamWrapper/DkanStreamWrapperTest.php
dafeder, line 32: Discovered that this only ever works if the site is up; it's not the kernel test environment that supplies the response for dkan://metastore, it's the ddev env itself! We don't want a kernel test dependent on the dev environment state.

modules/dkan_common/tests/src/Traits/DistributionReferenceModeTrait.php
```diff
@@ -0,0 +1,63 @@
...
+
+namespace Drupal\Tests\dkan_common\Traits;
+
+/**
+ * Helpers for tests that run with distribution references on/off.
+ *
+ * Must be used on BrowserTestBase, KernelTestBase or similar classes that
+ * implement a config() method.
+ */
+trait DistributionReferenceModeTrait {
```
dafeder, line 11: Reusable code to help kernel and functional tests run the same test against both reference configs.

modules/dkan_datastore/src/Controller/AbstractQueryController.php
```diff
@@ -170,14 +221,16 @@ public function queryResource(string $identifier, Request $request) {
    *   The json response.
    */
   public function queryDatasetResource(string $dataset, string $index, Request $request) {
-    $distribution_uuid = $this->datasetInfo->getDistributionUuid($dataset, $index);
-    if (empty($distribution_uuid)) {
+    $resource_identifier = $this->datasetInfo->getResourceIdentifier($dataset, $index);
+    if ($resource_identifier === NULL) {
       return $this->getResponse((object) ['message' => "No resource found for dataset $dataset at index $index"], 404);
     }
-    $dependencies = ['distribution' => [$distribution_uuid], 'dataset' => [$dataset]];
```
dafeder, line 177: Resolve the resource ID rather than the distribution ID; this works for both reference configs.

modules/dkan_datastore/src/Service/ResourceLocalizer.php
```diff
@@ -214,21 +214,28 @@ private function registerNewPerspectives(DataResource $resource, string $localFi
    * Also remove local perspectives from mapping DB.
    */
   public function remove($identifier, $version = NULL): void {
```
dafeder, line 216: The result of a ton of debugging that included burning through quite a few tokens with copilot. After we made changes that were supposed to allow all this to work in both modes, the purger tests started to fail randomly. Here is copilot's explanation for posterity:

> Root cause found and fixed. This is a real defect, not a race in the usual sense.

## What was actually happening

The diagnostics proved the purge logic was correct all along:
1. `resourceNotShared(…__1783714335__source)` → `distributions:0` → correctly purgeable.
2. The old version `…335` was in the final `delete` set — the purger did try to delete it.
3. Yet `public://resources/…_335` remained.

The failure was in ResourceLocalizer.php. It called the heavy `self::get()`, which **bails out if the source perspective is missing** ([get() returns null when `getResourceSource()` is null](modules/dkan_datastore/src/Service/ResourceLocalizer.php#L167)). In the test’s queue order, `orphan_resource_remover` runs before `resource_purger` and removes that source perspective, so `remove()` silently skipped the directory deletion.

## Why it looked flaky vs 4.x

This is the coupling your non-ref work introduced around orphan/source lifecycle: whether the source perspective for the old version was already gone when the purger ran determined whether the directory got deleted. That’s why it was intermittent and why 4.x (different orphan/source timing) didn’t show it.

## The fix

In `remove()`:
1. Look up the local-file and local-url perspectives **directly from the resource mapper** (no dependency on the source perspective or the file fetcher).
2. Delete the localized directory derived from `identifier_version`, with a `realpath()` existence guard, so cleanup is guaranteed even if the mapping is already gone.

modules/dkan_datastore/src/Service/ResourcePurger.php
```diff
@@ -267,12 +264,28 @@ private function getResourcesToPurge(int $initialVid, NodeInterface $node, bool
...
     // Extract the identifier and version from the supplied resource details.
     $identifier = DataResource::buildUniqueIdentifier(...json_decode($resource_details));
-    // Determine the number of distributions making use of the current
-    // resource.
-    $distributions = $this->referenceLookup->getReferencers('distribution', $identifier, 'downloadURL');
-    // If more than one distribution is using this resource, remove it from
-    // the purge list.
-    return count($distributions) <= 1;
+
+    if ($this->distributionsAreReferenced()) {
```
dafeder, line 268: Another piece that took a while to get right: the number of references we expect to consider the resource "shared" needs to be different depending on the mode.

modules/dkan_datastore/src/Service/ResourcePurger.php
```diff
@@ -351,24 +364,79 @@ private function getRevisionData(string $vid) : array {
...
+      return [
+        $resource['identifier'],
+        $resource['version'],
+        $resource['perspective'],
+      ];
+    }
+    catch (\Exception) {
+      return NULL;
+    }
+  }
```
dafeder, line 424: Revisit this - maybe can be compacted

modules/dkan_metastore/src/Storage/Data.php
```diff
@@ -168,7 +168,9 @@ public function count($unpublished = FALSE): int {
   public function retrieveAll(?int $start = NULL, ?int $length = NULL, bool $unpublished = FALSE): array {
     $entityIds = $this->listQueryBase($start, $length, $unpublished)->execute();
     return array_map(function ($entity) {
-      return $entity->get($this->metadataField)->getString();
+      return ($entity instanceof ContentEntityInterface)
+        ? $entity->get($this->metadataField)->getString()
+        : NULL;
```
dafeder, line 173: Just to make linter happy

modules/dkan_metastore/src/Storage/MetastoreStorageInterface.php
```diff
@@ -103,6 +103,17 @@ public function isPublished(string $uuid) : bool;
...
+  /**
+   * Archive.
+   *
+   * @param string $uuid
+   *   The identifier for the data.
+   *
+   * @return bool
+   *   True if success.
+   */
+  public function archive(string $uuid): bool;
```
dafeder, line 115: Mostly to make linter happy; was missing from interface

modules/dkan_metastore/src/MetastoreService.php
```diff
@@ -542,32 +553,43 @@ private function objectIsEquivalent(string $schema_id, string $identifier, Roote
...
    *   Metadata JSON object.
-   * @param string $prefix
-   *   Property prefix.
    *
    * @return \RootedData\RootedJsonData
    *   The metadata without any reference artifacts.
-   *
-   * @todo Probably remove the prefix param and just always use "%Ref".
    */
-  public static function removeReferences(RootedJsonData $object, $prefix = "%"): RootedJsonData {
```
dafeder, line 553: I know I'm changing a public signature, but feel pretty safe assuming noone has ever used an alternative reference prefix

modules/dkan_datastore/src/Controller/AbstractQueryController.php
```diff
@@ -145,17 +155,58 @@ public function queryResource(string $identifier, Request $request) {
...
+   * Resolve cache dependencies for a single-resource query identifier.
+   *
+   * @param string $identifier
+   *   Distribution UUID or resource identifier.
+   *
+   * @return array
+   *   Dependency array for \Drupal\dkan_metastore\MetastoreApiResponse.
+   */
+  protected function resolveResourceDependencies(string $identifier): array {
+    $distributionIds = Uuid::isValid($identifier)
```
dafeder, line 180: Note - possibly should add a mode check for performance, to avoid unnecessary call to getReferencers

modules/dkan_datastore/src/Controller/ImportController.php
dafeder, line 110: Again, may want to add direct mode check to avoid unnecessary getReferencers calls, which can be expensive.

modules/dkan_metastore/src/Reference/Referencer.php
```diff
@@ -203,22 +204,11 @@ protected function referenceSingle(string $property_id, $value) {
...
-    // to register it with the resource mapper.
-    if (isset($distribution->downloadURL)) {
-      // Register this distribution's resource with the resource mapper and
-      // replace the download URL with a unique ID registered in the resource
-      // mapper.
-      $distribution->downloadURL = $this->registerWithResourceMapper(
-        UrlHostTokenResolver::hostify($distribution->downloadURL), $this->getMimeType($distribution));
-    }
-
+  public function referenceDataDictionary($distribution): object {
```
dafeder: Collapse this with normalizeDictionaryValue()

modules/dkan_metastore/src/Reference/Referencer.php
```diff
@@ -98,22 +67,58 @@ public function __construct(
...
+  /**
+   * Recurses through distributions to find resources to reference in dataset.
+   *
+   * @param object $data
+   *   Dataset JSON object.
+   */
+  public function referenceDistributions(object $data): void {
+    foreach (($data->distribution ?? []) as &$distribution) {
+      $this->referenceResource($distribution);
+      $this->referenceDataDictionary($distribution);
```
dafeder: This gets renamed, see next comment

modules/dkan_metastore/src/Reference/Referencer.php
```diff
@@ -98,22 +67,58 @@ public function __construct(
...
+
+  /**
+   * Process a single distribution and replace URLs with resource IDs.
+   *
+   * @param object $distribution
+   *   The distribution object from the metadata.
+   */
+  public function referenceResource(object $distribution): void {
+    // Clean up any reference metadata if it exists.
+    unset($distribution->{'%Ref:downloadURL'});
```
dafeder: Replace with constant

modules/dkan_metastore/src/MetastoreService.php
```diff
@@ -481,21 +482,31 @@ public function getCatalog() {
   public function swapReferences(RootedJsonData $object): RootedJsonData {
     $no_schema_object = $this->getValidMetadataFactory()->get("$object", NULL);
     foreach ($no_schema_object->get('$') as $property => $value) {
-      if (substr_count((string) $property, "%Ref:") > 0) {
+      if (substr_count((string) $property, Dereferencer::REF_PREFIX) > 0) {
```
dafeder: replace with str_begins_with

modules/dkan_metastore/src/MetastoreService.php
```diff
@@ -481,21 +482,31 @@ public function getCatalog() {
...
+   *   The property name, including the %Ref prefix.
+   * @param mixed $value
+   *   The value of the %Ref property.
+   * @param \RootedData\RootedJsonData $object
+   *   The metadata object to swap the reference in.
+   *
+   * @return \RootedData\RootedJsonData
+   *   The metadata object with the reference swapped.
    */
   private function swapReference($property, $value, RootedJsonData $object): RootedJsonData {
```
dafeder: Rename to replaceReference

modules/dkan_metastore/src/MetastoreService.php
```diff
@@ -542,32 +553,43 @@ private function objectIsEquivalent(string $schema_id, string $identifier, Roote
...
+   * Meant to be used on output of RootedJsonData::get(), so expects an array
+   * rather than an object or RootedJsonData.
+   *
+   * @param array $array
+   *   Metadata JSON array.
+   *
+   * @return array
+   *   The metadata without any reference artifacts.
+   */
+  protected static function removeReferencesRecursive($array) {
```
dafeder, line 570: Collapse these two

modules/dkan_common/tests/src/Unit/UrlHostTokenResolverTest.php
```diff
@@ -38,17 +38,28 @@ class UrlHostTokenResolverTest extends TestCase {
    */
   public const FILE_PATH = 'tmp/mycsv.csv';
 
+  /**
```
dafeder, line 41: Necessary to deal with the drupal_static flag; add comment

modules/dkan_datastore/src/Controller/AbstractQueryController.php
dafeder: Maybe there's a way to use the IDs we already know about? but probably not

modules/dkan_metastore/src/MetastoreService.php
dafeder: OK well I collapsed this into its parent swapReferences() but since that's public and called elsewhere I decided to keep the name. Open to discussion.

dafeder (commented):
Initial convos w/ @paul-m , not exactly a review

modules/dkan_metastore/src/MetastoreService.php
dafeder: Hmm actually can't do this, the point is the recursive one calls itself, that's why it's recursive 🙃

modules/dkan_common/tests/src/Unit/UrlHostTokenResolverTest.php
```diff
@@ -38,17 +38,28 @@ class UrlHostTokenResolverTest extends TestCase {
...
+   * {@inheritdoc}
```
paul-m, line 42: Comment that this is to prevent re-use of container with drupal_static flag that we use.

.circleci/config.yml
```diff
@@ -225,7 +225,7 @@ jobs:
   cypress:
     machine:
       image: ubuntu-2204:current
-    parallelism: 4
+    parallelism: 5
```
paul-m: Revert this... Only want 5 on phpunit

modules/dkan_common/src/DatasetInfo.php
paul-m: This is a reasonable place for this to live.

modules/dkan_metastore/tests/src/Functional/MetastoreApiPageCacheTest.php
```diff
@@ -136,25 +147,23 @@ public function testDatasetApiPageCache() {
...
     $response = $this->apiRequest('GET', 'api/1/metastore/schemas/dataset/items/' . $identifier, ['show-reference-ids' => TRUE]);
     $dataset = json_decode($response->getBody()->getContents());
-    $distributionId = $dataset->distribution[0]->identifier ?? '';
-    $resourceId = $dataset->distribution[0]->data->{'%Ref:downloadURL'}[0]->identifier ?? '';
-    $response = $this->apiRequest('GET', 'api/1/datastore/imports/' . $distributionId);
-    $this->assertEquals(200, $response->getStatusCode(), $response->getBody());
-    $this->assertEquals('MISS', $response->getHeaders()['X-Drupal-Cache'][0]);
+    $resourceId = $dataset->distribution[0]->data->{'%Ref:downloadURL'}[0]->identifier
+      // If distributions not referenced, distribution has no "data" structure.
+      ?? $dataset->distribution[0]->{'%Ref:downloadURL'}[0]->identifier
```
paul-m: Should be the constant.

modules/dkan_metastore/tests/src/Kernel/LifeCycle/LifeCycleTest.php
```diff
@@ -74,6 +80,86 @@ protected function setUp(): void {
...
+   *   The download URL to use for the distribution.
+   * @param string|null $media_type
+   *   The media type to use for the distribution.
+   * @param string $dist_reference
+   *   The distribution reference setting to use.
+   *
+   * @dataProvider datasetSaveProvider
+   * @covers ::datasetSave
+   */
+  public function testDatasetSave(?string $download_url, ?string $media_type, string $dist_reference) {
```
paul-m, line 96: Add descriptive comments, please.

modules/dkan_metastore/tests/src/Unit/Reference/DereferencerTest.php
```diff
@@ -150,4 +217,250 @@ public function testDereferencePropertyUuidLogging() {
...
+
+  /**
+   * Test that a non-default display perspective adds a second reference entry.
+   *
+   * @covers ::dereferenceDistributions
+   * @covers ::dereferenceResource
+   * @covers ::retrieveDownloadUrlFromResourceMapper
+   * @covers ::createResourceReference
+   */
+  public function testDereferenceResourcesWithNonDefaultPerspective() {
```
paul-m, line 362: Revisit this test... Maybe add inline comments explaining what's going on.

modules/dkan_metastore/tests/src/Unit/Reference/DereferencerTest.php
```diff
@@ -150,4 +217,250 @@ public function testDereferencePropertyUuidLogging() {
...
+    );
+  }
+
+  /**
+   * Set the container to support the UrlHostTokenResolver.
+   *
+   * Note this would not be necessary if we used dependency injection properly
+   * for the UrlHostTokenResolver.
+   */
+  private function setContainerForUrlResolver() {
```
paul-m, line 453: Is this necesary?

modules/dkan_metastore/tests/src/Unit/Reference/ReferenceLookupTest.php
```diff
@@ -0,0 +1,71 @@
...
+        $this->createMock(ModuleHandlerInterface::class),
+      ])
+      ->onlyMethods(['decodeJsonMetadata'])
+      ->getMock();
+
+    $referenceLookup
+      ->expects($this->exactly(count($contains_items)))
+      ->method('decodeJsonMetadata')
+      ->willReturnOnConsecutiveCalls(
+        ['ref-array', ['identifier' => ['other', 'abc-array']]],
```
paul-m, line 58: Add `['ref-string', ['identifier' => 'abc']]`

tests/src/Functional/DatasetBTBTest.php
```diff
@@ -24,6 +26,7 @@
 class DatasetBTBTest extends BrowserTestBase {
```
paul-m, line 26: Split unrelated test changes out into a separate pull request.

paul-m (changes requested):

modules/dkan_metastore/tests/src/Unit/Reference/DereferencerTest.php
dafeder: Well, this is testing logic that from what I can tell has always been here, it just used to be in LifeCycle. I was not aware that a second reference was created. And what's weird is it looks identical to the first reference. This might be something to remove but for now I think this does test what we'd expect given the code - you can see the old version at https://github.com/GetDKAN/dkan/blob/4.x/modules/dkan_metastore/src/LifeCycle/LifeCycle.php#L271

paul-m (approved):

PR #4750, opened 2026-07-03
Remove metastore_resource_mapper_new_revision

dafeder:
Bookmarking an attempt to change revision logic to remove static flag `metastore_resource_mapper_new_revision`. Result of work with copilot and not at all ready for human consumption - will come back to this, full review, test, rewrite for consistency etc. 

Copilot's explanation of what it did:

Implemented the explicit-parameter form end-to-end and validated it with focused tests.

What changed:
1. Explicit new-version intent now flows through lifecycle and referencer calls (no static flag dependency).
- LifeCycle.php
- Referencer.php

2. Pre-reference decisioning was moved to event payload mutation in the datastore subscriber.
- Subscriber now reads triggering properties and sets createNewResourceVersion on event data.
- DatastoreSubscriber.php

3. Legacy static helpers were removed.
- ResourceMapper.php
- dkan_metastore.module

4. Distribution presave uses explicit behavior directly.
- LifeCycle.php

5. Functional regression coverage was updated to a deterministic signal.
- Instead of queue-depth checks, test now compares stored distribution downloadURL reference before/after patch.
- OnPreReferenceTest.php

Validation run:
1. dkan_datastore subscriber unit tests: OK (6 tests, 7 assertions)
2. metastore referencer unit tests: OK (12 tests, 16 assertions)
3. OnPreReference functional test: OK (1 test, 9 assertions)
4. LifeCycle distribution load kernel test: OK (1 test, 20 assertions)

### DIFF
--- a/modules/dkan_datastore/src/EventSubscriber/DatastoreSubscriber.php
+++ b/modules/dkan_datastore/src/EventSubscriber/DatastoreSubscriber.php
@@ -23,11 +23,9 @@ use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 class DatastoreSubscriber implements EventSubscriberInterface {
 
   /**
-   * Drupal Config Factory.
-   *
-   * @var \Drupal\Core\Config\ConfigFactoryInterface
+   * Config factory service.
    */
-  protected $configFactory;
+  private ConfigFactoryInterface $configFactory;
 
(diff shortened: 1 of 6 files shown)


dafeder:
Pretty sure I don't want to pursue this path

PR #4755, opened 2026-07-08
#4621 Intentionally set revision timestamp.

### LINKED ISSUE #4621: Publishing a metastore does not change revision date

swirtSJW:
Fixes #4621

This is a draft because it is completely untested yet 

## QA Steps

### DIFF
--- a/modules/dkan_metastore/src/Controller/MetastoreRevisionController.php
+++ b/modules/dkan_metastore/src/Controller/MetastoreRevisionController.php
@@ -180,6 +180,12 @@ class MetastoreRevisionController implements ContainerInjectionInterface {
       $data = json_decode($request->getContent());
       $entity->set('moderation_state', $data->state);
       $entity->setRevisionLogMessage($data->message);
+      if ($entity->isNewRevision()) {
+        // Set revision creation time to the current time because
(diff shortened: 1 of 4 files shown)


swirtSJW:
the tests are failing with 
> 1) Drupal\Tests\dkan_metastore\Unit\Storage\DataTest::testPublishDraftDataset
Error: Call to a member function getDefinition() on null

I'll have to step through to see how/if this is related to the new code.

PR #4757, opened 2026-07-14
Add Drupal AI Best Practices to composer file

dafeder:
This adds everything needed to

a) Add the AI best practices (optionally) to your dev environment when using the ddev contrib workflow.
b) Not add any of the assets it adds to version control by accident

While adding the AI assets to a project repo might be a good practice (it is what's recommended) I don't think it belongs in a module repo.

## QA Steps

1. After checking out branch, run ``dev poser`
2. Run `ddev composer drupal-ai skills-sync`
3. Confirm that you now have a .agents folder populated with skills, and that it is ignored by version control
4. Run `ddev composer drupal-ai skills-list` and confirm that you have 8 skills installed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,9 @@ docs/.venv
 /.ddev/web-build/*
 !/.ddev/web-build/Dockerfile
 tmp
+# For now, ignore any artifacts from Drupal AI Best Practices.
+/AGENTS.md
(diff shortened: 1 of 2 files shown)


dafeder:
Think we're holding off on this until we're more sure there is demand and that these skills are useful for this kind of  module development

PR #4758, opened 2026-07-15
Fix flaky cypress test

dafeder:
Attempt to fix the flaky Workflow cypress test fails nearly half the time. May be due to search index delay.

Also fixes the cypress command so that it works in both bash and zsh.

Note the update cypress test was run multiple times in CI and had no failure.

### DIFF
--- a/.ddev/commands/host/dkan-module-test-cypress
+++ b/.ddev/commands/host/dkan-module-test-cypress
@@ -33,8 +33,12 @@ fi
 TEST_USERS_JSON='[
   {"name":"testadmin","mail":"testadmin@test.com","role":"administrator"}
 ]'
-readarray -t TU_TEST_USERS < <(jq -c '.[]' <<<"$TEST_USERS_JSON")
-for TU_TEST_USER in "${TU_TEST_USERS[@]}"; do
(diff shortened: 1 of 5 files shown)


.ddev/commands/host/dkan-module-test-cypress
```diff
@@ -33,8 +33,12 @@ fi
 TEST_USERS_JSON='[
   {"name":"testadmin","mail":"testadmin@test.com","role":"administrator"}
 ]'
-readarray -t TU_TEST_USERS < <(jq -c '.[]' <<<"$TEST_USERS_JSON")
-for TU_TEST_USER in "${TU_TEST_USERS[@]}"; do
+TEST_USER_NAMES=""
```
dcgoodwin2112: Is this block used anywhere? I don't see TEST_USER_NAMES referenced anywhere else in the script

dcgoodwin2112 (approved):
lgtm 👍

.ddev/commands/host/dkan-module-test-cypress
dafeder: You're right, not sure what I was doing there. Removed.

PR #4764, opened 2026-07-22
Import controller cache behavior for non-referenced distributions

### LINKED ISSUE #4737: Cache Dependency and Invalidation Updates
**Goal**: Update cache dependency and invalidation behavior to operate from dataset/resource context rather than distribution lookups.

**Independent Test**: Exercise query, import summary, and download cache behavior with referenced and non-referenced distributions and verify invalidation stays correct.

### Tests

- [x] Add unit tests for query controller dependency extraction without mandatory distribution IDs in `modules/dkan_datastore/tests/src/Unit/Controller/AbstractQueryControllerTest.php`
- [x] Add kernel tests for import summary cache dependencies using resource/dataset context in `modules/dkan_datastore/tests/src/Kernel/Controller/ImportControllerTest.php`
- [x] Add functional tests for query download caching/invalidation behavior when distributions are non-referenced in `modules/dkan_datastore/tests/src/Functional/Controller/QueryDownloadControllerTest.php`
- [x] Add kernel test coverage for dataset dispatch summary counter accuracy (processed/skipped/failed) in `modules/dkan_datastore/tests/src/Kernel/DatastoreServiceEventsTest.php`

### Implementation

- [x] Update metastore dependency extraction to not require distribution lookup for cache tags/contexts in `modules/dkan_datastore/src/Controller/AbstractQueryController.php`
- [x] Update query resource and dataset-resource execution paths to prioritize resource/dataset identifiers in `modules/dkan_datastore/src/Controller/AbstractQueryController.php`
- [x] Refactor import summary dependency generation to support dataset+resource mapping invalidation in `modules/dkan_datastore/src/Controller/ImportController.php`
- [x] Update query download response caching to align with resource/dataset invalidation strategy in `modules/dkan_datastore/src/Controller/QueryDownloadController.php`
- [x] Review cache invalidation strategy notes for query/import/streaming endpoints and capture final decisions in `specs/004-minimal-datastore-no-distribution-id/research.md`

Created with assistance from [speckit](https://github.com/github/spec-kit).

dafeder:
* Refactors the cache dependency behavior for ImportController to more simply try both schemaIDs (mode-agnostic)
* Adds kernel test for import controller cache behavior
* Also adds unit test for AbstractQueryController::resolveResourceDependencies


modules/dkan_datastore/tests/src/Kernel/Controller/ImportControllerTest.php
```diff
@@ -1,9 +1,10 @@
 <?php
 
-namespace Drupal\Tests\dkan_datastore\Unit\Kernel;
```
dafeder, line 3: Oops!

modules/dkan_datastore/tests/src/Kernel/Controller/ImportControllerTest.php
```diff
@@ -36,4 +37,68 @@ public function testMultipleImports() {
...
+    $mock_lookup->method('getReferencers')
+      ->willReturnCallback(function ($schemaId, $referenceId) use ($ref_mode) {
+        if ($schemaId === 'distribution' && $ref_mode == 'referenced') {
+          return ['dist1', 'dist2'];
+        }
+        if ($schemaId === 'dataset' && $ref_mode == 'non-referenced') {
+          return ['dataset1', 'dataset2'];
+        }
+        return [];
+      });
```
dafeder, line 61: We just assume these would be the referencer results in ref vs non-referenced mode, not worth setting up actual datasets in both modes that is tested elsewhere.

modules/dkan_datastore/tests/src/Unit/Controller/AbstractQueryControllerTest.php
```diff
@@ -84,6 +88,62 @@ public function testInvalidJson() {
...
+        RootedJsonData $result,
+        array $dependencies = [],
+        ?ParameterBag $params = NULL,
+      ) {
+        // No-op for testing.
+      }
+
+    };
+
+    $ref_method = new \ReflectionMethod($controller, 'resolveResourceDependencies');
```
dafeder, line 128: We're testing a protected method on an abstract class, so need to create anonymous class + use reflection

tests/src/Functional/DatasetBTBTest.php
```diff
@@ -554,6 +555,13 @@ private function getResourceFromDataset(object $dataset) {
     return $dataset->distribution[0]->{'%Ref:downloadURL'}[0]->data;
   }
 
+  /**
+   * Set metastore distribution reference mode for this test run.
+   */
+  private function setDistributionReferenceMode(string $distribution_reference): void {
```
paul-m: I can't find any usages of this method... It only calls another method that's is in use.

paul-m (changes requested):

tests/src/Functional/DatasetBTBTest.php
dafeder: Oops I think this was a mistake from merge conflict resolution

paul-m (approved):

Issue #4453, opened 2025-03-26
More granular Event classes


dafeder:
We have a single generic Event class in DKAN shared everywhere, this doesn't make a lot of sense. Event classes specific to the specific event they're for make a lot more sense, they can have their own validation logic, etc. 

We might not need to do this for every event, especially if it requires no validation. 

Start with:

* AbstractDatabseTable::EVENT_TABLE_CREATE

PR #4767, opened 2026-07-28
Improve orphan handling for non-referenced distributions

### LINKED ISSUE #4741: Cleanup and Orphan Behavior Updates
**Goal**: Remove obsolete mappings and datastore artifacts without depending solely on orphaned distribution references.

**Independent Test**: Run cleanup flows for referenced and non-referenced resources and verify obsolete mappings and artifacts are removed.

### Tests

- [ ] Add functional cleanup tests for referenced and non-referenced orphan resource paths in `modules/dkan_metastore/tests/src/Functional/OrphanCheckerTest.php`

### Implementation

- [ ] Update orphan cleanup path to remove obsolete mappings/artifacts without requiring orphaned distribution references in `modules/dkan_metastore/src/Plugin/QueueWorker/OrphanResourceRemover.php`

dafeder:
Fixes #4741 

Updates various places in code we handle orphan distributions and resources to work in both referenced and non-referenced distribution modes.

* Create `LifeCycle::EVENT_DELETING_DATASET` event and add to `MetastoreSubscriber`. This performs orphan _resource_ cleanup when a dataset is deleted in non-referenced mode. Previously, this cleanup would have happened as a byproduct of the dataset's distributions being orphaned.
* Also refactors the methods in that subscriber to accommodate this
* Change resource purger to always delete file/table, but only delete resource mapper when in referenced mode. This is because in referenced mode, the mapping is removed elsewhere: "safe" is determined by `MetastoreSubscriber::clearItemResources()`, which only runs later, when the old distribution is actually orphaned via the queue. See note below
* Also create a new event class, LifeCycleEvent, which allows better validation downstream and has a more purposeful signature for, well, LifeCycle events. Hopefully the first step in moving away from having just one Event class for everything in DKAN and having no idea what you'll get from `::getData()` (see #4453). Switched to using this event class in some places in and around the LifeCycle (though not all; follow-up required here.)
* Clean up some references to `orphan_resource_remover`, which is no longer actually used as of #4742 (class still not removed)

## Notes

1. I'm not happy with this split between `ResourcePurger`, `MetastoreSubscriber` and `DatastoreSubscriber` (and maybe more?). There are too many places the same actions are happening -- removing datastore tables, resource localizer files, and resource mapper entries. Now the addition of logic around non-referenced distrbutions only makes this more confusing. There should be a unified "resource cleanup" function and a clear accounting of where it gets called from. When it happens directly on a Metastore event, a Datastore event, or deferred in a queue should be intuitive (and documented). But I think fixing that at the same time as adding referenced distributions is too much delta for one feature. We'll address later, separately from this feature work. See #4768
2. I wanted a more direct OrphanChecker test to be able to run but it is admittedly a bit redundant with several of the DatasetBTB tests. We really need to prune `DatasetBTBTest`, it's too overwhelming (not to mention _slow_).

### DIFF
--- a/modules/dkan_metastore/src/EventSubscriber/MetastoreSubscriber.php
+++ b/modules/dkan_metastore/src/EventSubscriber/MetastoreSubscriber.php
@@ -3,8 +3,8 @@
 namespace Drupal\dkan_metastore\EventSubscriber;
 
 use Drupal\dkan_common\DataResource;
-use Drupal\dkan_common\Events\Event;
-use Drupal\dkan_metastore\LifeCycle\LifeCycle as Dkan_metastoreLifeCycle;
+use Drupal\dkan_metastore\LifeCycle\LifeCycle;
+use Drupal\dkan_metastore\LifeCycle\LifeCycleEvent;
 use Drupal\dkan_metastore\MetastoreService;
 use Drupal\dkan_metastore\Plugin\QueueWorker\OrphanReferenceProcessor;
 use Drupal\dkan_metastore\Reference\Dereferencer;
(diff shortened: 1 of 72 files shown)


modules/dkan_metastore/src/EventSubscriber/MetastoreSubscriber.php
```diff
@@ -57,86 +62,114 @@ public static function create(ContainerInterface $container) {
...
       $version = $resourceParams['version'] ?? NULL;
-      $resource_id_wo_perspective = $resource_id . '__' . $version;
-      $resource = $this->resourceMapper->get($resource_id, $perspective, $version);
+      $resources[] = $this->resourceMapper->get($resource_id, $perspective, $version);
+    }
+    $this->cleanResourceMapperTable($resources, $schema_id, $identifier);
+  }
 
+  /**
+   * React to a metastore item being orphaned or deleted.
```
paul-m: Remove this line

tests/src/Functional/DatasetBTBTest.php
```diff
@@ -341,7 +341,6 @@ public function testOrphanDraftDistributionCleanup(string $distribution_referenc
       'localize_import',
       'datastore_import',
       'orphan_reference_processor',
-      'orphan_resource_remover',
```
paul-m, line 344: Add back orphan_resource_remover and let's make another issue to remove everything associated with it.

paul-m (changes requested):

paul-m (approved):

PR #4781, opened 2026-09-02
Remove saveRawMetadata()

paul-m:
It seems like we should not have to run `saveRawMetadata()` all over the place in the Data storage class. Let's try just removing it!


dafeder:
This seems to work with no side effects? But we're worried about merging because we don't really understand what saveRawMetadata was for in the first place. It seems to be connected to revision comparisons? We will come back to this later.

PR #4782, opened 2026-09-02
Add cacheability to metastore response.

### LINKED ISSUE #4783: Metastore API response should be cacheable to get picked up by Akamai
### User Story

It seems as though our api endpoint like /api/1/metastore/schemas/dataset/items/mxtu-43qs?show-reference-ids are ending up with the header of 'no-cache'.  This prevents Akamai from being able to do its job of caching and serving the response.

### Discovery
DKAN API responses ARE getting cached by Drupal. The site is just feeding us mis-information 

**The multitude of mis-information:**

1. When viewing the api endpoint while also logged into the edit domain, Drupal marks it as "non-cacheable" and dates it as expiring in 1978 so that a logged in user never sees cached information. If you want to see the cached variant, you have to logout first or open in ingognito window.
2. When checking on it locally, it will still always report x-drupal-cache = MISS because ddev deliberately configures Drupal’s page cache bin as `cache.backend.null` So the cache is always irretrievable and always results in a MISS.
3. The response "x-drupal-cache: MISS" is what has been cached higher up than Drupal (Varnish or Akamai) So the cache header is itself perpetuating a lie by caching the first time Drupal delivered it.
4. What is hard to tell is if the last updated and expires are also accurate or just a re-telling of a lie.

Image

However this entry 

Image

 does increment every time you refresh the page.  Internet research indicates this value is coming from Varnish, not Akamai. So this suggests that the problem is with Akamai. not caching what Varnish is serving. It also indicates Varnish is counting, so it means it is not reaching Drupal which is good.

I do think that even though drupal is caching it, drupal is still applying a 'no-cache' to the header.  It seems like this is the default reponse if cacheable headers are not specifically called out.

We should specifically declare the response as cacheable.

### Problem and Motivation

We need to make sure Akamai is serving as much as it can, so that too much traffic is not hitting Varnish, because even hits to Varnish, are still counted in Acquia's billing limits.

### Target Audience

- [x] End Users
- [ ] Developers
- [ ] Administrators
- [ ] Other

### Possible Implementation

Lets force the header to be listed as cacheable so that Akamai caches it instead of defering to Varnish.

### Acceptance Criteria

requests made to a dataset are served by Akamai past the first time.

swirtSJW:
Fixes #4783

## Describe your changes
These changes are an initial stab at defining caching.  Discussion is probably needed about the ideal numbers for caching.

## QA Steps

### DIFF
--- a/modules/dkan_metastore/src/MetastoreApiResponse.php
+++ b/modules/dkan_metastore/src/MetastoreApiResponse.php
@@ -58,6 +58,11 @@ class MetastoreApiResponse {
     if ($cacheMetadata = $this->getCacheMetadata($dependencies, $params)) {
       $response->addCacheableDependency($cacheMetadata);
     }
+    // Actively declare cacheable. Otherwise Drupal core will add 'no-cache'.
+    $response->headers->set(
+      'Cache-Control',
+      'public, max-age=3600, s-maxage=86400'
+    );
 
     return $response;
   }


modules/dkan_metastore/src/MetastoreApiResponse.php
```diff
@@ -58,6 +58,11 @@ public function cachedJsonResponse(
     if ($cacheMetadata = $this->getCacheMetadata($dependencies, $params)) {
       $response->addCacheableDependency($cacheMetadata);
     }
+    // Actively declare cacheable. Otherwise Drupal core will add 'no-cache'.
+    $response->headers->set(
+      'Cache-Control',
+      'public, max-age=3600, s-maxage=86400'
```
swirtSJW, line 64: The numbers might need some tweaking.

swirtSJW:
There is limits to what can be tested locally.  Willl have to deploy to an acquia environment to truly test it.

PR #4784, opened 2026-09-02
Add dkan_datastore_preview datastore submodule

dcgoodwin2112:
Fixes [issue#]

## Describe your changes

Adds `dkan_datastore_preview`, a submodule of `dkan_datastore` that renders a paginated, sortable HTML table of datastore data for each CSV/TSV distribution on the server-rendered dataset page (`/node/{nid}`). MVP scope from the ticket: find the datastore table and display it in the `dkan_metastore` twig template; the block and a configurable data-source plugin system are follow-ups.

- `DataSourceInterface` + `DatabaseDataSource` (seam for a future plugin system), `DataPreviewBuilder`, `dkan_datastore_preview` render element, `ImportStatusMessage`
- Server-rendered only: sort links and page-size form use per-table prefixed query params (`dp0_order`, `dp1_page_size`, …) so multiple previews on one page don't share state; no JavaScript
- Shows a status message instead of nothing when a resource isn't imported yet; distribution cache tags flip it to the table after import without a cache rebuild
- `node--data.html.twig` prints `{{ content.data_preview }}` (empty when the module is disabled; `dkan_metastore` gains no dependency on `dkan_datastore`)
- `ddev dkan-site-install` enables the module

## QA Steps

Prerequisite: a dataset with one imported CSV distribution (`drush dkan:datastore:list` shows `done`), enough rows to page (more than 25).

- [ ] `drush en dkan_datastore_preview`, then open the dataset's node page (`/node/{nid}`, not the React `/dataset/{uuid}` page). A captioned preview table renders below Downloads with "Showing 1-25 of N results".
- [ ] Click a column header. URL gains `dp0_order`/`dp0_sort`, the header gets `aria-sort`, rows re-sort. Click again to reverse.
- [ ] Set page size 10 and Apply. "Showing 1-10 of N", a pager appears, and pager links keep `dp0_page_size=10`.
- [ ] Open a dataset whose distributions are not CSV/TSV (or has none). No preview section, no errors in `drush watchdog:show`.
- [ ] `drush dkan:datastore:drop <identifier>` (identifier from `drush dkan:datastore:list`), `drush cr`, reload: "Data preview is not yet available…" message instead of the table.
- [ ] `drush dkan:datastore:import <identifier> && drush queue:run datastore_import`, reload **without** `drush cr`: the table is back.
- [ ] `drush pmu dkan_datastore_preview`, reload: page renders normally with no preview and no errors.

- [ ] I have updated or added documentation — no module README yet; will add with the block/plugin follow-up unless reviewers want it here

Co-authored by: Github Copilot


modules/dkan_datastore/modules/dkan_datastore_preview/dkan_datastore_preview.info.yml
```diff
@@ -0,0 +1,9 @@
+name: DKAN Datastore Preview
+description: Paginated, sortable HTML table previews of datastore data on dataset pages.
+type: module
+core_version_requirement: ^10.4 || ^11
+package: DKAN
+dependencies:
+  - dkan:dkan_common
+  - dkan:dkan_datastore
+  - dkan:dkan_metastore
```
dafeder, line 9: The module is a submodule of datastore, so it should be under modules/dkan_datastore/modules

modules/dkan_datastore/modules/dkan_datastore_preview/src/Service/DataPreviewBuilder.php
```diff
@@ -0,0 +1,368 @@
+<?php
+
+namespace Drupal\dkan_datastore_preview\Service;
```
dafeder, line 3: I know it's a pattern in the other modules but I think we'd like to get away from having a generic "Service" namespace. Let's just put this and ImportStatusMessage in the root namespace (/src).

modules/dkan_datastore/modules/dkan_datastore_preview/src/Service/DataPreviewBuilder.php
```diff
@@ -0,0 +1,368 @@
...
+/**
+ * Builds render arrays for data preview tables.
+ *
+ * Sorting and page size use query parameters prefixed per table instance
+ * (option 'query_prefix', e.g. "dp0_order") so several previews can coexist
+ * on one page without sharing state. Core TableSort is not used because its
+ * "order"/"sort" parameter names cannot be prefixed. Paging uses core's
+ * native comma-separated multi-pager "page" parameter via 'pager_element'.
+ */
+class DataPreviewBuilder {
```
dafeder, line 22: Service classes should have interfaces

modules/dkan_datastore/modules/dkan_datastore_preview/dkan_datastore_preview.services.yml
```diff
@@ -0,0 +1,20 @@
+services:
+  dkan.datastore_preview.builder:
+    class: \Drupal\dkan_datastore_preview\Service\DataPreviewBuilder
+    arguments:
+      - '@request_stack'
+      - '@pager.manager'
+  Drupal\dkan_datastore_preview\Service\DataPreviewBuilder: '@dkan.datastore_preview.builder'
```
dafeder, line 7: These service aliases should be to interfaces, not directly to the class

modules/dkan_datastore/modules/dkan_datastore_preview/dkan_datastore_preview.module
```diff
@@ -0,0 +1,150 @@
...
+
+use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\dkan_common\DataResource;
+use Drupal\node\NodeInterface;
+
+/**
+ * Implements hook_theme().
+ */
+function dkan_datastore_preview_theme() {
```
dafeder, line 16: I realize existing DKAN modules put all hook logic in the module file, but we are overdue to move them into hook classes. Since this module introduces a bunch of new hooks, let's do it right the first time and add them as hook classes off the bat, with D10 stubs in the module file for now.

See https://www.drupal.org/node/3442349

modules/dkan_datastore/modules/dkan_datastore_preview/dkan_datastore_preview.module
```diff
@@ -0,0 +1,150 @@
...
+      continue;
+    }
+
+    $resourceId = $dist['resource_id'] . '__' . $dist['resource_version'];
+
+    $preview = [
+      '#type' => 'container',
+      '#attributes' => ['class' => ['dkan-datastore-preview-distribution']],
+      '#cache' => [
+        'tags' => _dkan_datastore_preview_distribution_cache_tags($dist['distribution_uuid'] ?? ''),
```
dafeder, line 100: We should also use the dataset ID as a cache tag

modules/dkan_datastore/modules/dkan_datastore_preview/dkan_datastore_preview.module
```diff
@@ -0,0 +1,150 @@
...
+ * import completes (via PostImport), so attaching them here flips a "not yet
+ * available" message to the data table without a manual cache rebuild.
+ *
+ * @param string $distribution_uuid
+ *   The distribution's UUID.
+ *
+ * @return array
+ *   Cache tags, or an empty array when the item cannot be loaded.
+ */
+function _dkan_datastore_preview_distribution_cache_tags(string $distribution_uuid): array {
```
dafeder, line 139: I am not seeing expected caching behavior when I test this - if I drop a datastore table, I'd expect to see the the table dissapear from the page immediately without requiring a cache clear. But it does require a cache clear for me. However, doing some debugging I don't think this is the fault of the module; I think the way referencers are resolved from resource Ids is somehow broken in DKAN core. So, noting this but no action requested for now.

dafeder (changes requested):
This is looking really good. See comments bellow; my own tests look good except for the caching behavior, as noted (probably not this PR's fault). I haven't had time to fully read the unit and kernel tests, will do that tomorrow.

I also wonder if this might have made sense as a placeable block rather than something to hard code into the node template. I may be forgetting discussion and decisions already had about this though.

modules/dkan_datastore/modules/dkan_datastore_preview/dkan_datastore_preview.info.yml
dafeder: Wait it is! I must have been tired when I looked at this.

modules/dkan_datastore/modules/dkan_datastore_preview/src/DataSource/DatabaseDataSource.php
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * Data source that queries datastore tables directly via the database.
+ */
+class DatabaseDataSource implements DataSourceInterface {
+
+  /**
+   * Memoized storage objects, keyed by resource id.
+   *
+   * @var array
```
dafeder, line 16: Make it clearer what this is
```suggestion
   * @var \Drupal\dkan_datastore\Storage\DatabaseTable[]
```

modules/dkan_datastore/modules/dkan_datastore_preview/src/DataSource/DatabaseDataSource.php
```diff
@@ -0,0 +1,157 @@
...
+use Drupal\dkan_common\Storage\Query;
+use Drupal\dkan_datastore\DatastoreService;
+
+/**
+ * Data source that queries datastore tables directly via the database.
+ */
+class DatabaseDataSource implements DataSourceInterface {
+
+  /**
+   * Memoized storage objects, keyed by resource id.
```
dafeder, line 14: Trying to make this a little clearer/more explicit

```suggestion
   * Database table storage objects, keyed by resource id.
   * 
   * The $storages array is populated as ::getStorage is called; the array memoizes it for later use.
```

modules/dkan_datastore/modules/dkan_datastore_preview/src/DataSource/DatabaseDataSource.php
```diff
@@ -0,0 +1,157 @@
...
+  protected array $storages = [];
+
+  /**
+   * Constructor.
+   *
+   * @param \Drupal\dkan_datastore\DatastoreService $datastoreService
+   *   The DKAN datastore service.
+   */
+  public function __construct(
+    protected DatastoreService $datastoreService,
```
dafeder, line 27: Promoted constructor arguments should be `readonly` unless explicitly mutable

modules/dkan_datastore/modules/dkan_datastore_preview/src/DataSource/DatabaseDataSource.php
```diff
@@ -0,0 +1,157 @@
...
+   *   [identifier, version|null].
+   */
+  protected function parseResourceId(string $resource_id): array {
+    if (str_contains($resource_id, '__')) {
+      $parts = explode('__', $resource_id, 2);
+      return [$parts[0], $parts[1]];
+    }
+    return [$resource_id, NULL];
+  }
+
```
dafeder, line 112: Pretty sure this could be accomplished just by calling DataResource::getIdentifierAndVersion()

dafeder (changes requested):

dafeder (changes requested):
Let's also add a README or else edit the datastore pages in the docs or something.

dcgoodwin2112:
> Let's also add a README or else edit the datastore pages in the docs or something.

README.md added and included in docs/source/components/dkan_datastore.rst, same as the mysql_import README

modules/dkan_datastore/modules/dkan_datastore_preview/src/Service/DataPreviewBuilder.php
dcgoodwin2112: Moved DataPreviewBuilder and ImportStatusMessage (and their tests) to the module root namespace.

modules/dkan_datastore/modules/dkan_datastore_preview/src/Service/DataPreviewBuilder.php
dcgoodwin2112: Added DataPreviewBuilderInterface and ImportStatusMessageInterface. The element and tests now type against the interfaces; DataSourceInterface already existed for the data source.

modules/dkan_datastore/modules/dkan_datastore_preview/dkan_datastore_preview.services.yml
dcgoodwin2112: All three aliases now point at interfaces (DataPreviewBuilderInterface, DataSourceInterface, ImportStatusMessageInterface).

modules/dkan_datastore/modules/dkan_datastore_preview/dkan_datastore_preview.module
dcgoodwin2112: Done. Hooks live in src/Hook/DataPreviewHooks.php with #[Hook] attributes and injected dependencies; the .module file is down to three #[LegacyHook] stubs that delegate via \Drupal::service(). The class is also registered in services.yml so the stubs resolve on 10.4

modules/dkan_datastore/modules/dkan_datastore_preview/dkan_datastore_preview.module
dcgoodwin2112: Added. Each preview container now carries Cache::mergeTags($entity->getCacheTags(), $distributionTags)

modules/dkan_datastore/modules/dkan_datastore_preview/dkan_datastore_preview.module
dcgoodwin2112: Confirmed. I am seeing the same behavior locally on this one.

modules/dkan_datastore/modules/dkan_datastore_preview/src/DataSource/DatabaseDataSource.php
dcgoodwin2112: Applied suggestion

modules/dkan_datastore/modules/dkan_datastore_preview/src/DataSource/DatabaseDataSource.php
dcgoodwin2112: Applied suggestion

modules/dkan_datastore/modules/dkan_datastore_preview/src/DataSource/DatabaseDataSource.php
dcgoodwin2112: Done across all classes (DatabaseDataSource, DataPreviewBuilder, ImportStatusMessage, DataPreviewHooks)

modules/dkan_datastore/modules/dkan_datastore_preview/src/DataSource/DatabaseDataSource.php
dcgoodwin2112: Done. This has been updated to use DataResource::getIdentifierAndVersion()

